PR #1, opened 2015-07-14
changed url format for better prometheus integration

kbudde:
Hi,

I had to no idea how to integrate your nice little script into prometheus. I've changed the endpoint and added an example config for prometheus. This works for me.
Perhaps you have a better idea?

BR

Kris


brian-brazil:
https://github.com/prometheus/prometheus/issues/861 is the other part of this, as that'll handle escaping too

brian-brazil:
That's upstream now, and the readme explains how to use it so this change isn't needed.

Issue #4, opened 2015-09-22
Community is hardcoded as public in the code


ShalomCohen:
Can you please change the code so that the community will be passed together with the address as parameter? hardcoding it isn't the smartest choice.

brian-brazil:
I'm happy to take a PR to make it a config option. As it could be a secret, we shouldn't be making it visible in prometheus.

brian-brazil:
Fixed by #8

PR #6, opened 2015-10-03
Add support for 'snmpcommunity'

ledeuns:
``snmpcommunity`` is configuration directive to specify the read community.
https://github.com/prometheus/snmp_exporter/issues/4


snmp_exporter/collector.py
```diff
@@ -56,11 +56,16 @@ def collect_snmp(config, host, port=161):
   """Scrape a host and return prometheus text format for it"""
 
   start = time.time()
+  
+  community = 'public'
+  if (config.has_key('snmpcommunity')):
```
brian-brazil: config.get('snmpcommunity', 'public')

I think just 'community' would be sufficient, as it's implicit that it's SNMP.

brian-brazil:
Superseded by #7

PR #30, opened 2016-03-07
Add Cisco WLC config used at FOSDEM

brian-brazil:
@RichiH


RichiH:
IIRC, the WLC supports uptime as well.

Else, it looks good.

brian-brazil:
Do we still have access to kit to test that?

brian-brazil:
I'm going to merge this for now, and we can add more later.

PR #32, opened 2016-03-12
Feature to add labels (based on static oid), description and type to metrics

totallyunknown:
We need the ability to add metric tag based on a oid. For example we want to tag all metrics with the sysName and sysLocation. I've updated the sysUpTime  example in `snmp.yml`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build/
 dist/
 *.pyc
 *.egg-info
+.idea
--- a/README.md
+++ b/README.md
@@ -67,3 +67,19 @@ This is to allow for customisation of what's done during the scrape as many
 special cases are expected.  The varying levels of SNMP MIB-parsing support
 across different langauges also means that a single language may not be
 practical.
+
+## Metric Labels
+
+It's possible to add extra labels to the metrics. For example `sysName or
+`sysLocation` are pretty useful.
+
+```YAML
+  metrics:
+    - name: sysUpTime
+      oid: 1.3.6.1.2.1.1.3
+      labels:
+       - labelname: sysName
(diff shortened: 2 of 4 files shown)


RichiH:
Nitpick, but adding a short section to the docs would be good.

totallyunknown:
Better?

totallyunknown:
I've added some more useful stuff for adding a help text (description) and the metric type (counter or gauge).

brian-brazil:
> For example we want to tag all metrics with the sysName and sysLocation. 

This is an anti-pattern, you should be using target labels for that.

The intent is that this file will be autogenerated from a MIB, I've not yet come across any MIB that'd require a structure like this. Do you have an example?

totallyunknown:
> This is an anti-pattern, you should be using target labels for that.

Is there a difference in the prometheus storage level regarding target labels and labels coming from metrics? I understand that labels with a high cardinality are a problem. In our case the snmp_exporter is running on a bastion host, acting as a gateway to the management network. The prometheus server itself hasn't direct access. 

It feels a bit weird to setup a second process to gather labels via SNMP to use it as target labels in Prometheus. 

There must be a solution how an exporter can provide target labels to Prometheus directly. Should we discuss how it is possible to extend the exposition format with this?

brian-brazil:
> Is there a difference in the prometheus storage level regarding target labels and labels coming from metrics?

There's no difference at that level.

> It feels a bit weird to setup a second process to gather labels via SNMP to use it as target labels in Prometheus.

You'll need to do that anyway to give Prometheus the targets, this is a service discovery problem. This sounds like a specialist use case, so http://www.robustperception.io/using-json-file-service-discovery-with-prometheus/ should point you down the right path.

PR #33, opened 2016-04-01
Save the index of the OID as a label

SuperQ:
In order to avoid label matching collisions, save the OID index value as
a label called 'index'.


SuperQ:
@brian-brazil This solves the uniqueness problems for `ifDescr`

brian-brazil:
This is incorrect in the case of multiple indexes in a table, and adds a label that will make working with these metrics more difficult. If you want this it should be possible entirely from the config.

PR #34, opened 2016-04-05
Add SNMPv1 support

Theci:
Original software only supports SNMPv2, while some older hardware need SNMPv1. Main problem lays within the use of getBulk, which was only introduced in SNMPv2.
By adding an additional parameter we can check the version and act accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,42 @@ Visit http://localhost:9116/metrics?address=1.2.3.4 where 1.2.3.4 is the IP of t
 SNMP device to get metrics from. You can also specify a `module` parameter, to
 choose which module to use from the config file.
 
+
+## Auththentication
+
+The default configuration is to use SNMPv2 with the community `public`.  This can be changed in the yaml config.  NOTE: Version 2 implies SNMP version 2c.
+
+Example:
+```YAML
+default:
+  version: 2
+  auth:
+    community: SomeCommunityString
+  walk:
+    - ...
+  metrics:
+    - ...
+```
+
+For SNMPv3, the authentication requires different parameters.  The `auth_proto` defaults to `MD5` and the `priv_proto` defaults to `DES`.  The `security_level` defaults to `noAuthNoPriv`.
+
+Example:
+```YAML
+default:
+  version: 3
+  auth:
+    username: SomeUser
+    password: TotallySecret
+    auth_proto: SHA
+    priv_proto: AES
+    security_level: SomethingReadOnly
+    priv_password: SomeOtherSecret
+  walk:
+    - ...
+  metrics:
+    - ...
+```
+
 ## Prometheus Configuration
 
 The snmp exporter needs to be passed the address as a parameter, this can be
--- a/README.md
+++ b/README.md
@@ -29,7 +29,15 @@ choose which module to use from the config file.
 
 ## Auththentication
 
+###SNMPv2
+
 The default configuration is to use SNMPv2 with the community `public`.  This can be changed in the yaml config.  NOTE: Version 2 implies SNMP version 2c.
+
+####Authentication parameters
+
+Name | Description
+--------|------------
+community | Community string defined on the device
 
 Example:
 ```YAML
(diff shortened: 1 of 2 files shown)


snmp_exporter/collector.py
```diff
@@ -5,21 +5,42 @@
...
 from prometheus_client import Metric, CollectorRegistry, generate_latest, Gauge
 
-def walk_oids(host, port, oids, community):
-  session = netsnmp.Session(Version=2, DestHost=host, RemotePort=port,
-      Community=community, UseNumeric=True, Retries=3)
+def walk_oids(host, port, oids, version, auth):
+  if version == 3:
+    auth_user = auth['username']
+    auth_pass = auth['password']
+    auth_proto = auth.get('auth_proto', 'MD5')
```
brian-brazil: let's spell this as protocol

snmp_exporter/collector.py
```diff
@@ -5,21 +5,42 @@
...
+    sec_level = auth.get('security_level', 'noAuthNoPriv')
+    session = netsnmp.Session(Version=3, DestHost=host, RemotePort=port,
+        SecName=auth_user, AuthPass=auth_pass, AuthProto=auth_proto,
+        PrivProto=priv_proto, PrivPass=priv_pass, SecLevel=sec_level,
+        UseNumeric=True, Retries=3)
+  elif version == 2:
+    community = auth.get('community', 'public')
+    session = netsnmp.Session(Version=2, DestHost=host, RemotePort=port,
+        Community=community, UseNumeric=True, Retries=3)
+  elif version==1:
```
brian-brazil: spaces around operators

README.md
```diff
@@ -26,6 +26,56 @@ Visit http://localhost:9116/metrics?address=1.2.3.4 where 1.2.3.4 is the IP of t
...
+  version: 1
+  auth:
+    community: SomeCommunityString
+  walk:
+    - ...
+  metrics:
+    - ...
+````
+
+For SNMPv3, the authentication requires different parameters.  The `auth_proto` defaults to `MD5` and the `priv_proto` defaults to `DES`.  The `security_level` defaults to `noAuthNoPriv`.
```
brian-brazil: Can you make it clear which settings are required?

snmp_exporter/collector.py
```diff
@@ -5,21 +5,42 @@
...
     while True:
       # getbulk starts from the last oid we saw.
       vl = netsnmp.VarList(netsnmp.Varbind('.' + last_oid))
-      if not session.getbulk(0, 25, vl):
-        return
-
+      if version==2 or version==3:
+        if not session.getbulk(0, 25, vl):
+          return
+      if version==1:
```
brian-brazil: An if/else would probably be clearer here

README.md
```diff
@@ -26,6 +26,79 @@ Visit http://localhost:9116/metrics?address=1.2.3.4 where 1.2.3.4 is the IP of t
...
+
+###Authentication parameters
+
+Name | Description | required
+--------|--------------|--------------
+username | A string representing the name of the user | yes
+password |  If messages sent on behalf of this user can be authenticated, the (private) authentication key for use with the authentication protocol. Defined as authKey in RFC3414 | yes
+auth_protocol | An indication of whether messages sent on behalf of this user can be authenticated, and if so, the type of authentication protocol which is used. 2 protocols are defined in RFC3414: MD5 (HMAC-MD5-96) and SHA (HMAC-SHA-96) | no
+priv_protocol | An indication of whether messages sent on behalf of this user can be protected from disclosure, and if so, the type of privacy protocol which is used. Only one protocol is defined in RFC3414: DES (CBC-DES Symmetric Encryption Protocol) | no
+security_level | The Level of Security from which the User-based Security module determines if the message needs to be protected from disclosure and if the message needs to be authenticated. | no (if this is changed, required depends on this)
```
Theci: security_level determines if priv_password is required or not. At it's default value NoAuthNoPriv no authentication or privacy protocol is used so priv_password is not required

snmp_exporter/collector.py
```diff
@@ -5,21 +5,42 @@
...
-      if not session.getbulk(0, 25, vl):
-        return
-
+      if version == 1:
+        if not session.getnext(vl):
+          return
+      else:
+        if not session.getbulk(0, 25, vl):
+          return
+
```
Theci: Implemented if/else for clarity and switched it around to only evaluate if version 1

README.md
brian-brazil: That should be captured in the required field for priv_password then.

brian-brazil:
What's the status of this PR? It'd be great to get v1 and v3 support in.

Theci:
I've took note of your comments and modified the lines accordingly. Also expanded SNMPv3 documentation. The SNMPv1 support has been tested over 3 week period and showed no problems. SNMPv3 has only been tested briefly but the authentication and data collection worked for that period.

brian-brazil:
That looks good, can you squash your commits please?

snmp_exporter/collector.py
```diff
@@ -5,21 +5,42 @@
...
+def walk_oid(session, version, oid):
     last_oid = oid
     while True:
       # getbulk starts from the last oid we saw.
       vl = netsnmp.VarList(netsnmp.Varbind('.' + last_oid))
-      if not session.getbulk(0, 25, vl):
-        return
-
+      if version == 1:
+        if not session.getnext(vl):
```
thomseddon: It seems that if the index of table being iterated isn't the penultimate item in the oid, this function will just keep iterating in an infinite loop.

My preferred fix for this is just to use session.walk - all the back and forth is then handled in net-snmp c and I believe it mimics the getbulk most closely. For example:

```python
if not session.walk(vl):
  return
```

Alternatively you could check you're not in an infinite loop like so

```python
next_oid = v.tag[1:] + '.' + v.iid
if not (last_oid + '.').startswith(oid + '.') or next_oid == last_oid:
  return
last_oid = next_oid
```

thomseddon: My proposal would be: https://github.com/thomseddon/snmp_exporter/commit/e6b8fe9b44613eb5053198afa7c4e0bc68885ee8 (also removes trailing white spaces)

brian-brazil:
This has been sitting here a while. Would someone like to condense it into a final PR that I can merge?

rossigee:
To avoid an error parsing an empty string to an int when v1 host fails to respond...

```
--- a/snmp_exporter/collector.py
+++ b/snmp_exporter/collector.py
@@ -97,7 +97,8 @@ def collect_snmp(config, host, port=161):
   values = walk_oids(host, port, config['walk'], config.get('version', 2), config.get('auth', {}))
   oids = {}
   for oid, value in values:
-    oids[oid_to_tuple(oid)] = value
+    if oid != '':
+      oids[oid_to_tuple(oid)] = value
 
   # Netsnmp doesn't tell us if an error has occured, so
   # try to spot it by no results.

```

Theci:
I've squashed my commits and took note of @rossigee and @thomseddon their comments and adjusted code accordingly.

brian-brazil:
Thanks!

snmp_exporter/collector.py
```diff
@@ -5,20 +5,41 @@
 
 from prometheus_client import Metric, CollectorRegistry, generate_latest, Gauge
 
-def walk_oids(host, port, oids, community):
-  session = netsnmp.Session(Version=2, DestHost=host, RemotePort=port,
-      Community=community, UseNumeric=True, Retries=3)
+def walk_oids(host, port, oids, version, auth):
+  if version == 3:
+    auth_user = auth['username']
```
brian-brazil: This doesn't agree with the docs in the README, this should be auth.get too.

PR #38, opened 2016-06-06
Refactored snmp.yml into a directory of snippets and added modules for Disk, CPU and Temperature

Svedrin:
Hi,

since the default configuration only queries network interfaces, I extended it to also query:

- Disk Utilization (Used/Free space)
- Disk Performance (IOPS, Load)
- CPU System/User/Idle
- Linux Load Average over 1/5/15 minutes

To be able to do this more easily, I also split up the snmp.yml config file into a directory of smaller snippets that can be maintained more easily (e.g., you can mix auto-generated files with files you wrote yourself).

This pull requests contains commits for these two changes along with an updated snmp.yml.

Cheers,
Svedrin.


RichiH:
The OIDs in `walk` are not sorted correctly.

I wonder how this split works with multiple models other than `default`; this might it simpler, or more complex, I honestly don't know.

brian-brazil:
All these metrics would more appropriately come from the node exporter. We'd also need a new config structure, as walk remains centralised here.

Svedrin:
@RichiH I didn't know they need to be sorted - what would be the correct order, simply ascending?

I don't quite know what structure is best, but this way allows the default module to be extended as well as other modules to be added, so I think it's a start :)

Svedrin:
@brian-brazil while that is true, it would require me to roll out the node exporter everywhere, while SNMP is already deployed in our network. So I'd like to just use it, if possible.

brian-brazil:
You're free to use it yourself. 

We generally want to encourage users to use the best general tools, so we're unlikely to add stats that duplicate other components as defaults as that'd lead to suboptimal usage. This is similar to how our exporter guidelines say not to include machine stats.

Svedrin:
@brian-brazil okay, how about adding it as a "we don't encourage this, if you absolutely MUST, do it this way" kind of documentation?

brian-brazil:
We've generally avoided such documentation across Prometheus, as the experience is that users ignore the warning.

Svedrin:
...most recent example being me, I guess. :D well then, just asking.

PR #40, opened 2016-06-07
Fix infinite loop for endOfMibView SNMP response

Svedrin:
Hi,

when querying an OID that the target device does not support, netsnmp's getbulk seems to return a tuple that only contains an empty string -- i.e., `('',)`. This caused SNMP Exporter to go into an infinite loop.

I extended the result check to cope with this situation.

Cheers,
Svedrin.

### DIFF
--- a/snmp_exporter/collector.py
+++ b/snmp_exporter/collector.py
@@ -14,16 +14,13 @@ def walk_oids(host, port, oids, community):
       yield v
 
 def walk_oid(session, oid):
+    prev_oid = None
     last_oid = oid
-    while True:
+    while prev_oid != last_oid:
+      prev_oid = last_oid
       # getbulk starts from the last oid we saw.
       vl = netsnmp.VarList(netsnmp.Varbind('.' + last_oid))
-      result = session.getbulk(0, 25, vl)
-      # result check: when querying an OID that the target
-      # device does not support, some HP switches prematurely
-      # send endOfMibView instead of noSuchObject. This causes
-      # getbulk to return ('',).
-      if not result or (len(result) == 1 and not result[0]):
+      if not session.getbulk(0, 25, vl):
         return
 
       for v in vl:
--- a/snmp_exporter/collector.py
+++ b/snmp_exporter/collector.py
@@ -18,7 +18,8 @@ def walk_oid(session, oid):
     while True:
       # getbulk starts from the last oid we saw.
       vl = netsnmp.VarList(netsnmp.Varbind('.' + last_oid))
-      if not session.getbulk(0, 25, vl):
+      result = session.getbulk(0, 25, vl)
+      if not result or (len(result) == 1 and not result[0]):
         return
 
       for v in vl:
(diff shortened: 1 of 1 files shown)


snmp_exporter/collector.py
```diff
@@ -18,7 +18,8 @@ def walk_oid(session, oid):
     while True:
       # getbulk starts from the last oid we saw.
       vl = netsnmp.VarList(netsnmp.Varbind('.' + last_oid))
-      if not session.getbulk(0, 25, vl):
+      result = session.getbulk(0, 25, vl)
+      if not result or (len(result) == 1 and not result[0]):
```
brian-brazil: Can you add a comment?

Svedrin:
done.

snmp_exporter/collector.py
```diff
@@ -18,7 +18,10 @@ def walk_oid(session, oid):
     while True:
       # getbulk starts from the last oid we saw.
       vl = netsnmp.VarList(netsnmp.Varbind('.' + last_oid))
-      if not session.getbulk(0, 25, vl):
+      result = session.getbulk(0, 25, vl)
+      # result check: when querying an OID that the target
+      # device does not support, getbulk returns ('',)
```
brian-brazil: Your PR description says endOfMibView, this says a MIB that's not supported. Which is it?

Svedrin:
I reproduced the issue using `snmpbulkwalk`. What basically happens is the equivalent of:

```
# snmpwalk -v2c -c $COMM 10.30.11.246 1.3.6.1.4.1.2021.13.15.1.1
UCD-DISKIO-MIB::diskIOEntry = No more variables left in this MIB View (It is past the end of the MIB tree)
```

The switch directly sends back an endOfMibView message:

```
09:24:31.926674 IP peha042-n1.proemion.com.60928 > 10.30.11.246.snmp:  C=community GetNextRequest(31)  E:2021.13.15.1.1
09:24:31.928307 IP 10.30.11.246.snmp > peha042-n1.proemion.com.60928:  C=community GetResponse(31)  E:2021.13.15.1.1=[endOfMibView]
```

I'm not sure if it _should_ do that (a different switch just says "No Such Object available on this agent at this OID" which sounds way more reasonable), but that's what it does, and this causes the bahavior I observed.

brian-brazil:
Then the comment should be about endOfMib.

Svedrin:
I clarified the comment a bit.

brian-brazil:
This still doesn't sound right. With GETBULK we're always looking for multiple results, and if we request something beyond the end of the MIB then endOfMib is the correct response.

Svedrin:
ok, so... I'm lost. How would you write it?

Svedrin:
I ran a couple more tests. I tried to add a check against the type of the varbinds in `vl` to correctly recognize and handle ENDOFMIBVIEW results, something along the lines of:

```python
      for varbind in vl:
        if varbind.type != "ENDOFMIBVIEW":
          break
      else:
        return
```

Unfortunately I couldn't get it to work, because I always get a row of empty `OCTETSTRING`s before getting a row with `ENDOFMIBVIEW` type. So I still have to check the actual values instead of just the type, or that check will never match anything.

Frankly, I don't know enough about SNMP or these libs to be able to fix this properly. Right now, making sure the OID has actually changed before calling `getbulk()` again seems to be the easiest and most robust solution in preventing `walk_oid` from looping indefinitely.

brian-brazil:
This is no longer relevant with the Go rewrite.

PR #44, opened 2016-06-14
Add support for SNMP v1

thomseddon:
Hey,

We have some devices that support v1 only, this patch adds support!

There doesn't seem to be a specific max line length in use however if the config retrieval had been added in-line, as done with community, it would have made line 83 the longest in the file, hence why it's broken out. Let me know if you'd like an alternative style.

Thanks,


brian-brazil:
I'm actually planning on merging #34 as that adds both v1 and v3.

thomseddon:
Ah yes, will keep an eye on it

PR #45, opened 2016-06-15
add InetAddress capabilities

AlexRRR:
Hello! Recently I need this feature (a table containing both ipv4 and ipv6 addresses indexes) so I thought I would share it. This pull request would allow for OIDs which require `InetAddressType` and `InetAddress` decoding as table indexes.

as RFC4001 states:

>  When a generic Internet address is used as an index, both the
   InetAddressType and InetAddress objects MUST be used.  The
   InetAddressType object MUST be listed before the InetAddress object
   in the INDEX clause.

using `InetAddress` as type in the yaml file would read both indexes and decode the address according to its type. Currently only ipv4 and ipv6 supported.

### DIFF
--- a/snmp_exporter/collector.py
+++ b/snmp_exporter/collector.py
@@ -62,6 +62,20 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
+    # InetAddress is always formed by [InetAddressType][InetAddressXX] 
+    elif index['type'] == 'InetAddress':
+      address_type = suboid[0]
+      octets = suboid[1:2][0]
+      address = suboid[2: 2 + octets]
+      label_oids[index['labelname']] = suboid[0: 2 + octets] 
+      # ipv4
+      if address_type == 1:
+        labels[index['labelname']] = '.'.join(str(s) for s in address)
+      # ipv6
+      elif address_type == 2:
+        labels[index['labelname']] = ':'.join(("{0:02X}".format(s) for s in address))   
+      suboid = suboid[2 + octets :]
+
   for lookup in lookup_config:
     index_oid = itertools.chain(*[label_oids[l] for l in lookup['labels']])
     full_oid = oid_to_tuple(lookup['oid']) + tuple(index_oid)
--- a/tests/collector.py
+++ b/tests/collector.py
@@ -6,15 +6,18 @@ class TestCollector(unittest.TestCase):
 
 
   def test_parse_indexes(self):
-    oids = {(1, 2, 3, 4): 'eth0', (1, 3, 65, 32, 255): "octet"}
+    oids = {(1, 2, 3, 4): 'eth0', (1, 3, 65, 32, 255): 'octet',
+            (1, 1, 4, 10, 181, 1, 30): 'ipv4',
+            (1, 2, 16, 42, 6, 29, 128, 0, 1, 0, 3, 0, 0, 0, 0, 0, 1, 1, 52):
+            'ipv6'}
     self.assertEqual({}, parse_indexes((), [], [], oids))
-    self.assertEqual({'l': '4'}, 
+    self.assertEqual({'l': '4'},
         parse_indexes((4,), [{'labelname': 'l', 'type': 'Integer32'}], [], oids))
-    self.assertEqual({'l': 'eth0'}, 
-        parse_indexes((4,), [{'labelname': 'l', 'type': 'Integer32'}], 
+    self.assertEqual({'l': 'eth0'},
+        parse_indexes((4,), [{'labelname': 'l', 'type': 'Integer32'}],
                       [{'labels': ['l'], 'labelname': 'l', 'oid': '1.2.3'}], oids))
-    self.assertEqual({'a': '3', 'b': '4', 'l': 'eth0'}, 
-        parse_indexes((3, 4,), [{'labelname': 'a', 'type': 'Integer32'}, {'labelname': 'b', 'type': 'Integer32'}], 
+    self.assertEqual({'a': '3', 'b': '4', 'l': 'eth0'},
+        parse_indexes((3, 4,), [{'labelname': 'a', 'type': 'Integer32'}, {'labelname': 'b', 'type': 'Integer32'}],
                       [{'labels': ['a', 'b'], 'labelname': 'l', 'oid': '1.2'}], oids))
     self.assertEqual({'l': '0'},
         parse_indexes((), [{'labelname': 'l', 'type': 'Integer32'}], [], {}))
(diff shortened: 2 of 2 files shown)


RichiH:
> Currently only ipv4 and ipv6 supported.

IPv1, IPv2, and IPv3 being deprecated and IPv5 being burnt by a stray research project, what else is there to support?

AlexRRR:
Well, you would be surprised, again from the RFC:

> InetAddressType ::= TEXTUAL-CONVENTION
    STATUS      current
    DESCRIPTION
        "A value that represents a type of Internet address.

         unknown(0)  An unknown address type.  This value MUST
                     be used if the value of the corresponding
                     InetAddress object is a zero-length string.
                     It may also be used to indicate an IP address
                     that is not in one of the formats defined
                     below.

         ipv4(1)     An IPv4 address as defined by the
                     InetAddressIPv4 textual convention.

         ipv6(2)     An IPv6 address as defined by the
                     InetAddressIPv6 textual convention.

         ipv4z(3)    A non-global IPv4 address including a zone
                     index as defined by the InetAddressIPv4z
                     textual convention.

         ipv6z(4)    A non-global IPv6 address including a zone
                     index as defined by the InetAddressIPv6z
                     textual convention.

         dns(16)     A DNS domain name as defined by the
                     InetAddressDNS textual convention.

snmp_exporter/collector.py
```diff
@@ -62,6 +62,20 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
+    # InetAddress is always formed by [InetAddressType][InetAddressXX] 
+    elif index['type'] == 'InetAddress':
+      address_type = suboid[0:1][0]
+      octets = suboid[1:2][0]
+      address = suboid[2: 2 + octets]
+      label_oids[index['labelname']] = suboid[0: 2 + octets] 
+      #ipv4
```
brian-brazil: space after #

snmp_exporter/collector.py
```diff
@@ -62,6 +62,20 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
+    # InetAddress is always formed by [InetAddressType][InetAddressXX] 
+    elif index['type'] == 'InetAddress':
+      address_type = suboid[0:1][0]
```
brian-brazil: `suboid[0]` would do the same

RichiH:
@AlexRRR You are correct, I am surprised :)

dns is kinda obvious in hindsight; ipvNz is, well...

````
The zone index (bytes 17-20) is used to disambiguate
identical address values on nodes that have interfaces
attached to different zones of the same scope.  The zone index
may contain the special value 0, which refers to the default
zone for each scope.
````

File: snmp_exporter/collector.py [hunk 3]
```diff
@@ -62,6 +62,20 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
...
+      octets = suboid[1:2][0]
+      address = suboid[2: 2 + octets]
+      label_oids[index['labelname']] = suboid[0: 2 + octets] 
+      #ipv4
+      if address_type == 1:
+        labels[index['labelname']] = '.'.join(str(s) for s in address)
+      #ipv6
+      elif address_type == 2:
+        labels[index['labelname']] = ':'.join(("{0:02X}".format(s) for s in address))
```
brian-brazil: We should probably fall back to a hex string here for the other types

brian-brazil:
Can you add unittests please?

snmp_exporter/collector.py
```diff
@@ -62,9 +62,9 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
-    # InetAddress is always formed by [InetAddressType][InetAddressXX] 
+    #InetAddress is always formed by [InetAddressType][InetAddressXX] 
     elif index['type'] == 'InetAddress':
-      address_type = suboid[0:1][0]
+      address_type = suboid[0]
       octets = suboid[1:2][0]
```
brian-brazil: same here

snmp_exporter/collector.py
```diff
@@ -62,9 +62,9 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
-    # InetAddress is always formed by [InetAddressType][InetAddressXX] 
+    #InetAddress is always formed by [InetAddressType][InetAddressXX]
```
brian-brazil: There should be a space after the #

tests/collector.py
```diff
@@ -24,3 +27,9 @@ def test_parse_indexes(self):
     self.assertEqual({'l': 'octet'},
         parse_indexes((3, 65, 32, 255), [{'labelname': 'l', 'type': 'OctetString'}],
                       [{'labels': ['l'], 'labelname': 'l', 'oid': '1'}], oids))
+    self.assertEqual({'l': '10.181.1.30'},
```
brian-brazil: You should also test this without the lookup, so we know the rendering into the label value is correct

tests/collector.py
```diff
@@ -24,3 +27,9 @@ def test_parse_indexes(self):
...
+        parse_indexes((4, 10, 181, 1, 30), [{'labelname': 'l', 'type': 'InetAddress'}],
```
brian-brazil: Are you missing the InetAddressType here?

AlexRRR: You are absolutely right ⚡

tests/collector.py
```diff
@@ -9,7 +9,7 @@ def test_parse_indexes(self):
     oids = {(1, 2, 3, 4): 'eth0', (1, 3, 65, 32, 255): 'octet',
             (1, 4, 10, 181, 1, 30): '10.181.1.30',
```
brian-brazil: This is missing the length, use a different value to make sure you're testing the right thing.

AlexRRR: not sure what you meant by the missing length (4 is the length there), but I do see your point about making sure I am testing the right thing.

brian-brazil: With the lookup you've specified, 4 is the type and 10 is the length.

AlexRRR: got it, could you please take a look at my new attempt?

snmp_exporter/collector.py
```diff
@@ -62,6 +62,20 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
+    #InetAddress is always formed by [InetAddressType][InetAddressXX] 
+    elif index['type'] == 'InetAddress':
+      address_type = suboid[0]
+      octets = suboid[1:2][0]
```
brian-brazil: This can be simplified to `suboid[1]`

snmp_exporter/collector.py
```diff
@@ -62,6 +62,20 @@ def parse_indexes(suboid, index_config, lookup_config, oids):
       label_oids[index['labelname']] = [length] + content
       labels[index['labelname']] = ''.join((chr(s) for s in content))
       suboid = suboid[length+1:]
+    #InetAddress is always formed by [InetAddressType][InetAddressXX]
```
brian-brazil: Always a space after #

AlexRRR:
I've fixed the comments and squashed commits.

brian-brazil:
Thanks!

PR #48, opened 2016-08-03
Ensure example config is included, add systemd unit

jacekn:
MANIFEST.in specifies example_configs directory which does not exist. Create it, move snmp.yml there and also add basic systemd unit file

### DIFF
--- a/prometheus-snmp-exporter.service
+++ b/prometheus-snmp-exporter.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=prometheus SNMP exporter
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/snmp_exporter /etc/prometheus/snmp.yaml
+
+[Install]
+WantedBy=multi-user.target


brian-brazil:
A directory like that would be for SNMP exporter configs, not all types of configs.

This file is hardcoding paths, which I don't think we can presume.

jacekn:
Where would you like to store the unit file?

There are hardcoded paths but I think it's OK for an _example_ unit file. Vendors can write their own and users can either modify it or put files where the unit file expects them.

brian-brazil:
Chuck it in the root.

jacekn:
Done

RichiH:
This might be too late, but shouldn't it be `/etc/prometheus.d/`?

jacekn:
Ubuntu 16.04 prometheus package uses /etc/prometheus/ for config, I think it makes sense to go with that but since it's an example I don't mind changing it if people have strong opinions.

brian-brazil:
A .d would imply we're combining multiple files into one, which we aren't.

brian-brazil:
Can you put the primary config back where it was, and squash your commits?

RichiH:
@brian-brazil Just for completeness' sake, you were right about `.d`.

jacekn:
Commit squashed

brian-brazil:
Thanks!

PR #49, opened 2016-08-12
Fix python3 compatibility problems

jacekn:
Fixed a few python3 compatibility problems

### DIFF
--- a/snmp_exporter/http.py
+++ b/snmp_exporter/http.py
@@ -1,14 +1,19 @@
 import traceback
-import urlparse
-from BaseHTTPServer import BaseHTTPRequestHandler
-from BaseHTTPServer import HTTPServer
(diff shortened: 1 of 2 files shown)


snmp_exporter/http.py
```diff
@@ -1,14 +1,18 @@
...
+from .collector import collect_snmp
 
-
-from collector import collect_snmp
+try:
+    from urlparse import urlparse
+    from BaseHTTPServer import BaseHTTPRequestHandler
+    from BaseHTTPServer import HTTPServer
+    from SocketServer import ForkingMixIn
+except ImportError:
```
brian-brazil: A comment here that this is for Python 3 would be useful

jacekn:
Comment added

brian-brazil:
Thanks!

PR #61, opened 2016-09-06
Add IpAddress index type.

jcollie:
Needed for BGP4-MIB (and likely others).

PR #62, opened 2016-09-06
Add generic BGP MIBs and Cisco-specific BGP MIBs.

jcollie:
Requires the IpAddress index type from #61.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -1221,3 +1221,285 @@ apcups:
         - labels: [iemStatusProbeName]
           labelname: iemStatusProbeName
           oid: 1.3.6.1.4.1.318.1.1.10.2.3.2.1.2
+
+# The BGP MIB is from RFC 1657/4273 and should be supported by many
+# router vendors. Of course, you'll need to actually be running BGP
+# for this to be useful.
+#
+# ftp://ftp.cisco.com/pub/mibs/v2/BGP4-MIB.my
+# https://tools.ietf.org/html/rfc1657
+# https://tools.ietf.org/html/rfc4273
+
+bgp4:
+  version: 2
+  walk:
+    - 1.3.6.1.2.1.15.2
+    - 1.3.6.1.2.1.15.3
+  metrics:
+    - name: bgpLocalAs
+      oid: 1.3.6.1.2.1.15.2
+    - name: bgpPeerState
+      oid: 1.3.6.1.2.1.15.3.1.2
+      indexes:
+        - labelname: bgpPeerIdentifier
+          type: IpAddress
+      lookups:
+        - labels: [bgpPeerIdentifier]
+          labelname: bgpPeerRemoteAddr
+          oid: 1.3.6.1.2.1.15.3.1.7
+    - name: bgpPeerAdminStatus
+      oid: 1.3.6.1.2.1.15.3.1.3
+      indexes:
+        - labelname: bgpPeerIdentifier
+          type: IpAddress
+      lookups:
+        - labels: [bgpPeerIdentifier]
+          labelname: bgpPeerRemoteAddr
+          oid: 1.3.6.1.2.1.15.3.1.7
+    - name: bgpPeerNegotiatedVersion
+      oid: 1.3.6.1.2.1.15.3.1.4
+      indexes:
+        - labelname: bgpPeerIdentifier
+          type: IpAddress
+      lookups:
+        - labels: [bgpPeerIdentifier]
+          labelname: bgpPeerRemoteAddr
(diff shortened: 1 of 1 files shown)


snmp.yml
```diff
@@ -1221,3 +1221,377 @@ apcups:
...
+    - name: bgpLocalAs
+      oid: 1.3.6.1.2.1.15.2
+    - name: bgpPeerState
+      oid: 1.3.6.1.2.1.15.3.1.2
+      indexes:
+        - labelname: bgpPeerIdentifier
+          type: IpAddress
+      lookups:
+        - labels: [bgpPeerIdentifier]
+          labelname: bgpPeerLocalAddr
```
brian-brazil: The index of this is only bgpPeerRemoteAddr

snmp.yml
```diff
@@ -1221,3 +1221,377 @@ apcups:
...
+          oid: 1.3.6.1.2.1.15.3.1.7
+
+# The Cisco BGP4 MIB is proprietary to Cisco routers. Not all routers
+# support all parts of the MIB, even if they are running BGP so your
+# mileage may vary. If possible, upgrade to the latest version of the
+# firmware possible.
+#
+# ftp://ftp.cisco.com/pub/mibs/v2/CISCO-BGP4-MIB.my
+
+cbgp4:
```
brian-brazil: cisco_bgp4 to be consistent

snmp.yml
```diff
@@ -1221,3 +1221,377 @@ apcups:
...
+  walk:
+    - 1.3.6.1.4.1.9.9.187.1.2
+  metrics:
+    - name: cbgpPeerAcceptedPrefixes
+      oid: 1.3.6.1.4.1.9.9.187.1.2.4.1.1
+      indexes:
+        - labelname: bgpPeerRemoteAddr
+          type: IpAddress
+        - labelname: cbgpPeerAddrFamilyAfi
+          type: Integer32
```
brian-brazil: It'd probably be best to add an InetAddressType for this

snmp.yml
```diff
@@ -1221,3 +1221,377 @@ apcups:
...
+        - labelname: cbgpPeerAddrFamilySafi
+          type: Integer32
+      lookups:
+        - labels: [bgpPeerRemoteAddr, cbgpPeerAddrFamilyAfi, cbgpPeerAddrFamilySafi]
+          labelname: cbgpPeerAddrFamilyName
```
brian-brazil: I don't see this in the index

snmp.yml
jcollie: That seems like overkill, as InetAddressType is just a textual convention for an integer.

snmp.yml
jcollie: I think that cbgpPeerAddrFamilyName is strictly in a separate SNMP table, but it's very useful in that it maps (cbgpPeerAddrFamilyAfi, cbgpPeerAddrFamilySafi) to a textual description of the address family, e.g. (1, 1) -> "IPv4 Unicast" or (2, 1) -> "IPv6 Unicast".

snmp.yml
brian-brazil: `ipv4` is far more meaningful than `1`

snmp.yml
brian-brazil: These configs will ultimately be generated directly from the MIBs, so if it can't be determined by the MIBs it shouldn't be included.

snmp.yml
jcollie: I can see your point about "ipv4" being more meaningful. There are potentially a lot of textual conventions that simply map an integer to a label though. Adding code for each one I think would be a maintenance issue. I think that it'd be better to have a generic mechanism in the configuration for defining these, especially if they could eventually be auto-generated. Also, since the indexes are used in looking up other OIDs, converting them from integers to string labels would mean a lot more complexity in making matches.

brian-brazil: I wouldn't add code for all of them, however InetAddressType is a core SNMP type.

The lookup logic handles this already, it always works by oid.

snmp.yml
```diff
@@ -1221,3 +1221,285 @@ apcups:
...
+    - 1.3.6.1.2.1.15.3
+  metrics:
+    - name: bgpLocalAs
+      oid: 1.3.6.1.2.1.15.2
+    - name: bgpPeerState
+      oid: 1.3.6.1.2.1.15.3.1.2
+      indexes:
+        - labelname: bgpPeerIdentifier
+          type: IpAddress
+      lookups:
```
brian-brazil: You don't need this lookup, it's already the index

brian-brazil:

PR #66, opened 2016-09-21
collector.go: Add snmp_collection_errors_total

RichiH:


collector.go
```diff
@@ -13,6 +13,12 @@ import (
 )
 
 var (
+	snmpCollectionErrors = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "snmp_collection_errors_total",
+			Help: "Errors encountered while collection data.",
```
brian-brazil: Collecting

collector.go
```diff
@@ -48,6 +55,7 @@ func ScrapeTarget(target string, config *Module) ([]gosnmp.SnmpPDU, error) {
 		p, err := strconv.Atoi(port)
 		if err != nil {
 			return nil, fmt.Errorf("Error converting port number to int for target %s: %s", target, err)
+			snmpCollectionErrors.Inc()
```
brian-brazil: This doesn't fit with the other two errors, as it's a user error rather than an snmp error.

collector.go
```diff
@@ -118,6 +128,7 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
 	if err != nil {
 		log.Errorf("Error scraping target %s: %s", c.target, err)
 		ch <- prometheus.NewInvalidMetric(prometheus.NewDesc("snmp_error", "Error scraping target", nil, nil), err)
+		snmpCollectionErrors.Inc()
```
brian-brazil: Isn't this double-counting with ScrapeTarget? Rarely should one metric like this be incremented in more than one function.

brian-brazil:
Stale review.

PR #76, opened 2016-10-19
Added MikroTik RouterOS to snmp.yml

ryanturner:
I've tested this on various devices and versions of RouterOS. Their SNMP implementation is a bit flaky (it varies by version, they don't expose everything to SNMP, and they don't document everything that they do expose), and I don't have an environment to test everything that they do implement. As a result, I'm hesitant to implement those. I imagine you have recommended changes to what's a lookup vs index etc; I'm happy to work and make those changes if you'll help me understand what can be improved on.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
           type: Integer32
         - labelname: fgWcWtpSessionWtpId
           type: OctetString
+          
+# MikroTik RouterOS Devices
+# There are a few undocumented MIBs in use in this, and a few MIBs that I was unable to test so they were not implemented
+# May require tweaking to meet your exact needs
+mikrotik_routeros:
+  walk:
+    - 1.3.6.1.4.1.14988.1.1
+  metrics:
+    - name: mtxrWlRtabAddr
+      oid: 	1.3.6.1.4.1.14988.1.1.1.2.1.1
+    - name: mtxrWlRtabRouterOSVersion
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.10
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabUptime
+      oid: 	1.3.6.1.4.1.14988.1.1.1.2.1.11
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabSignalToNoise
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.12
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabIface
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.2
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabTxBytes
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.4
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabRxBytes
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.5
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabTxPackets
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.6
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabRxPackets
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.7
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabTxRate
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.8
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+    - name: mtxrWlRtabRxRate
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.9
+      indexes:
+        - labelname: mtxrWlRtabAddr
+          type: PhysAddress48
+
+    - name: mtxrWlApIndex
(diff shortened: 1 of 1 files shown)


snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+# May require tweaking to meet your exact needs
+mikrotik_routeros:
+  walk:
+    - 1.3.6.1.4.1.14988.1.1
+  metrics:
+    - name: mtxrWlRtabAddr
+      oid: 	1.3.6.1.4.1.14988.1.1.1.2.1.1
+    - name: mtxrWlRtabRouterOSVersion
+      oid: 1.3.6.1.4.1.14988.1.1.1.2.1.10
+      indexes:
```
brian-brazil: There are two indexes in this table:

```
mtxrWlRtabEntry OBJECT-TYPE
    SYNTAX MtxrWlRtabEntry
    MAX-ACCESS not-accessible
    STATUS current
    DESCRIPTION "Wireless registration table. It is indexed by remote
        mac-address and local interface index"
    INDEX { mtxrWlRtabAddr, mtxrWlRtabIface }
    ::= { mtxrWlRtabTable 1 }
```

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+          type: PhysAddress48
+
+    - name: mtxrWlApIndex
+      oid: 1.3.6.1.4.1.14988.1.1.1.3.1.1
+    - name: mtxrWlApOverallTxCCQ
+      oid: 1.3.6.1.4.1.14988.1.1.1.3.1.10
+      indexes:
+        - labelname: mtxrWlApIndex
+          type: Integer32
+      lookups:
```
brian-brazil: I don't think there's a way this lookup could be automatically calculated from the MIB, so this should be removed as once this is all auto-generated this won't be here.

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+        - labels: [mtxrWlApIndex]
+          labelname: mtxrWlApSsid
+          oid: 1.3.6.1.4.1.14988.1.1.1.3.1.4
+        - labels: [mtxrWlApBand]
+          labelname: mtxrWlApBand
+          oid: 1.3.6.1.4.1.14988.1.1.1.3.1.8
+    - name: mtxrWlApNoiseFloor
```
brian-brazil: Not going to expose mtxrWlApOverallTxCCQ and mtxrWlApAuthClientCount from this table?

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+          type: Integer32
+      lookups:
+        - labels: [mtxrWlApIndex]
+          labelname: mtxrWlApSsid
+          oid: 1.3.6.1.4.1.14988.1.1.1.3.1.4
+    - name: mtxrHlCoreVoltage
+      oid: 1.3.6.1.4.1.14988.1.1.3.1
+    - name: mtxrHlTemperature
+      oid: 1.3.6.1.4.1.14988.1.1.3.10
+      indexes:
```
brian-brazil: This is not part of a table, it has no indexes

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+      lookups:
+        - labels: [mtxrSerialNumber]
+          labelname: mtxrSerialNumber
+          oid: 1.3.6.1.4.1.14988.1.1.7.3
+        - labels: [mtxrSerialNumber]
+          labelname: mtxrFirmwareVersion
+          oid: 1.3.6.1.4.1.14988.1.1.7.4
+    - name: mtxrLicUpgrUntil
+      oid: 1.3.6.1.4.1.14988.1.1.4.2
+      indexes:
```
brian-brazil: This is not in a table.

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+      lookups:
+        - labels: [mtxrLicSoftwareId]
+          labelname: mtxrLicVersion
+          oid: 1.3.6.1.4.1.14988.1.1.4.4
+        - labels: [mtxrLicSoftwareId]
+          labelname: mtxrLicSoftwareId
+          oid: 1.3.6.1.4.1.14988.1.1.4.1
+
+    - name: mtxrNeighborIfIndex
+      oid: 1.3.6.1.4.1.14988.1.1.11.1.1.8
```
brian-brazil: I can't find this particular entry in the MIB, however the index is mtxrNeighborTable and there should be no lookups.

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+          labelname: mtxrNeighborVersion
+          oid: 1.3.6.1.4.1.14988.1.1.11.1.1.4
+
+
+    - name: mtxrInterfaceStatsIndex
+      oid: 1.3.6.1.4.1.14988.1.1.14.1.1.1
+      indexes:
+        - labelname: mtxrInterfaceStatsIndex
+          type: Integer32
+      lookups:
```
brian-brazil: No lookups on these

snmp.yml
```diff
@@ -1565,3 +1565,946 @@ fortinet_fortigate:
...
+          labelname: mtxrInterfaceStatsName
+          oid: 1.3.6.1.4.1.14988.1.1.14.1.1.2
+    - name: mtxrDHCPLeaseCount
+      oid: 1.3.6.1.4.1.14988.1.1.6.1
+    - name: mtxrScriptRunCmd
+      oid: 1.3.6.1.4.1.14988.1.1.8.1.1.3
+      indexes:
+        - labelname: mtxrScriptIndex
+          type: Integer32
+      lookups:
```
brian-brazil: No lookups here

snmp.yml
ryanturner: Can you elaborate on why this means it should be removed? It is being reported appropriately in my testing.

snmp.yml
ryanturner: Can you elaborate on why this means it should be removed? It is being reported appropriately in my testing.

snmp.yml
brian-brazil: The indexes need to be removed, not the metric.

snmp.yml
brian-brazil: As above.

brian-brazil:
Please resubmit using the generator. (We support more types now too).

PR #86, opened 2016-11-20
Add support for ServerTech PDUs

SuperQ:
Add support for ServerTech PDUs based on MIBs
* Add `PDUOutlet` and `PDUInfeed` tree types.
* Add basic metrics for PDU infeeds and outlets.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -1565,3 +1565,82 @@ fortinet_fortigate:
           type: Integer32
         - labelname: fgWcWtpSessionWtpId
           type: OctetString
+
+# ServerTech Sentry 3 MIB
+#
+# Used by ServerTech PDUs
+#
+# ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3OIDTree.txt
+# ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3.mib
+servertech_sentry3:
+  walk:
+    - 1.3.6.1.2.1.1
+    - 1.3.6.1.4.1.1718.3.2.2.1
+    - 1.3.6.1.4.1.1718.3.2.3.1
+  metrics:
+    - name: sysUpTime
+      oid: 1.3.6.1.2.1.1.3
+    - name: infeedLoadValue
+      oid: 1.3.6.1.4.1.1718.3.2.2.1.7
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+    - name: infeedVoltage
+      oid: 1.3.6.1.4.1.1718.3.2.2.1.11
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+    - name: infeedPower
+      oid: 1.3.6.1.4.1.1718.3.2.2.1.12
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+    - name: infeedEnergy
+      oid: 1.3.6.1.4.1.1718.3.2.2.1.16
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+    - name: outletLoadValue
+      oid: 1.3.6.1.4.1.1718.3.2.3.1.7
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+        - labelname: outletIndex
+          type: Integer32
+    - name: outletVoltage
+      oid: 1.3.6.1.4.1.1718.3.2.3.1.13
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+        - labelname: outletIndex
+          type: Integer32
+    - name: outletPower
+      oid: 1.3.6.1.4.1.1718.3.2.3.1.14
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+        - labelname: outletIndex
+          type: Integer32
+    - name: outletEnergy
+      oid: 1.3.6.1.4.1.1718.3.2.3.1.18
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+        - labelname: outletIndex
+          type: Integer32


collector.go
```diff
@@ -241,6 +241,22 @@ func indexesToLabels(indexOids []int, metric *Metric, oidToPdu map[string]gosnmp
 				parts[i] = fmt.Sprintf("%02X", o)
 			}
 			labels[index.Labelname] = strings.Join(parts, ":")
+		case "PDUOutlet":
```
brian-brazil: Can you point me to the RFC where this is defined?

SuperQ: There is no RFC, this is enterprise custom defined in the MIB:

```
    outletEntry OBJECT-TYPE
        SYNTAX      OutletEntry
        MAX-ACCESS  not-accessible
        STATUS      current
        DESCRIPTION
            "Row definition for the outlet table."
        INDEX       { towerIndex, infeedIndex, outletIndex }
        ::= { outletTable 1 }
```

```
    infeedEntry OBJECT-TYPE
        SYNTAX      InfeedEntry
        MAX-ACCESS  not-accessible
        STATUS      current
        DESCRIPTION
            "Row definition for the input feed table."
        INDEX       { towerIndex, infeedIndex }
        ::= { infeedTable 1 }
```

SuperQ: Yup.  I mentioned tree indexing style like this way of doing things previously. You rejected supporting it generically.

brian-brazil: Looking at the full MIB that's a standard multiple-index table. It's supported.

SuperQ: Ok, derp, I think I understand this now.  Will update PR.

SuperQ:
Ok, now we have a good example of an `Integer32` index tree. :)

snmp.yml
```diff
@@ -1565,3 +1565,132 @@ fortinet_fortigate:
...
+    community: public
+  walk:
+    - 1.3.6.1.2.1.1
+    - 1.3.6.1.4.1.1718.3.2.2.1
+    - 1.3.6.1.4.1.1718.3.2.3.1
+  metrics:
+    - name: sysUpTime
+      oid: 1.3.6.1.2.1.1.3
+    # Infeed metrics
+    - name: infeed_load_hundredths_amper
```
brian-brazil: I'm fairly sure that's not the name in the MIB

snmp.yml
```diff
@@ -1565,3 +1565,132 @@ fortinet_fortigate:
...
           type: OctetString
+
+# ServerTech Sentry 3 MIB
+#
+# Used by ServerTech PDUs
+#
+# ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3OIDTree.txt
+# ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3.mib
+servertech_sentry3:
+  version: 2
```
brian-brazil: These are defaults, no need to specify

snmp.yml
```diff
@@ -1565,3 +1565,132 @@ fortinet_fortigate:
...
+    - name: infeed_votage_tenths_volt
+      oid: 1.3.6.1.4.1.1718.3.2.2.1.11
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+      lookups:
+        - labels: [towerIndex, infeedIndex]
+          labelname: infeedName
```
brian-brazil: This label isn't adding specificity.

snmp.yml
```diff
@@ -1565,3 +1565,132 @@ fortinet_fortigate:
...
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+        - labelname: outletIndex
+          type: Integer32
+      lookups:
+        - labels: [towerIndex, infeedIndex, outletIndex]
+          labelname: outletName
```
brian-brazil: If you were to do this it should replace outletIndex.

For now I'm tending to not having any lookups, until we have the generator.

snmp.yml
SuperQ: No, but it's more useful than what's in the MIB.

snmp.yml
SuperQ: Names are user generated, not always unique.

snmp.yml
SuperQ: Well, technically it's in the MIB.  This is a more useful descriptive version of the MIB information:

```
    infeedLoadValue OBJECT-TYPE
        SYNTAX      INTEGER(-1..60000)
        UNITS       "hundredth Amps"
        MAX-ACCESS  read-only
        STATUS      current
        DESCRIPTION
            "The load measured on the input feed line.  A non-negative
             value indicates the measured load in hundredths of Amps.  A
             negative value indicates that a load value was not able to
             be measured."
        ::= { infeedEntry 7 }
```

SuperQ: If you want to stick to something that could be machine generated, and fit with Prometheus naming convention, how about something like this:

`infeed_load_value_hundredth_amps` or `infeedLoadValue_hundredth_amps`

brian-brazil: You should use exactly what's in the MIB, as that's what the generator will produce.

snmp.yml
brian-brazil: This can't be automatically determined from the MIB, so it doesn't belong.

snmp.yml
brian-brazil: This isn't adding any specificity then, so doesn't belong as a label.

snmp.yml
SuperQ: Well, I propose that the generator should generate using both the name and the units available.

snmp.yml
SuperQ: Sure it can, we could automatically map all `DisplayString` items from the MIB.

```
    OutletEntry ::= SEQUENCE {
        outletIndex          INTEGER,
        outletID             DisplayString,
        outletName           DisplayString,
        outletCapabilities   BITS,
        outletStatus         INTEGER,
        outletLoadStatus     INTEGER,
        outletLoadValue      INTEGER,
        outletLoadLowThresh  INTEGER,
        outletLoadHighThresh INTEGER,
        outletControlState   INTEGER,
        outletControlAction  INTEGER,
        outletCapacity       INTEGER,
        outletVoltage        INTEGER,
        outletPower          INTEGER,
        outletApparentPower  INTEGER,
        outletPowerFactor    INTEGER,
        outletCrestFactor    INTEGER,
        outletEnergy         INTEGER,
        outletWakeupState    INTEGER,
        outletPostOnDelay    INTEGER
    }
```

snmp.yml
brian-brazil: That's not what we've been doing thus far.

snmp.yml
brian-brazil: That doesn't work in general, and I don't want to tie us down before the generator is in place.

snmp.yml
SuperQ: So, by that logic, we should drop support for lookups, as lookups in SNMP are never guaranteed to be unique.  And there's never a case where we should allow mapping them into labels.

brian-brazil: Yip, they're not turning out to be as useful and clean as I had thought.

snmp.yml
```diff
@@ -1565,3 +1565,116 @@ fortinet_fortigate:
...
+      oid: 1.3.6.1.2.1.1.3
+    # Infeed metrics
+    - name: infeedLoadValue
+      oid: 1.3.6.1.4.1.1718.3.2.2.1.7
+      indexes:
+        - labelname: towerIndex
+          type: Integer32
+        - labelname: infeedIndex
+          type: Integer32
+      #lookups:
```
brian-brazil: Can you remove these comments please?

brian-brazil:
Thanks!

PR #93, opened 2016-12-18
Add support for eaton connectups

alfh:
Eaton UPSes with ConnectUPS cards expose statistics via SNMP.

This change adds support for Eaton UPS to snmp_exporter.

I think the OIDs used are so general that it might work with some other manufacturers' UPSes as well, but I am only able to test it with my Eaton UPS.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -1222,6 +1222,317 @@ apcups:
           labelname: iemStatusProbeName
           oid: 1.3.6.1.4.1.318.1.1.10.2.3.2.1.2
 
+# Eaton ConnectUPS Web/SNMP cards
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you link to the actual MIB?

RichiH:
http://powerquality.eaton.de/Support/Software-Drivers/Downloads/connectivity-firmware.asp seems to have those files.

brian-brazil:
The generator is at a PoC stage at https://github.com/prometheus/snmp_exporter/tree/generator, all future additions will be via it.

brian-brazil:

PR #94, opened 2016-12-19
snmp.yml: Add Aruba wireless controllers

RichiH:
MIBs can be found at http://support.arubanetworks.com/default.aspx behind paywall/registration. There is no legal way to have world-readable access to the files.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -1644,3 +1644,1071 @@ servertech_sentry3:
           type: Integer32
         - labelname: outletIndex
           type: Integer32
+
+aruba_controller:
+  walk:
+    - 1.3.6.1.2.1.1.3
+    - 1.3.6.1.2.1.2
+    - 1.3.6.1.2.1.31.1.1
+    - 1.3.6.1.4.1.14823.2.2.1.1.3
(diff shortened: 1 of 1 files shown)


RichiH:
@juliusv @SuperQ: FYI.

This is without string support, and without matching up AP names via MAC address.

snmp.yml
```diff
@@ -1644,3 +1644,1071 @@ servertech_sentry3:
           type: Integer32
         - labelname: outletIndex
           type: Integer32
+
+aruba_controller
```
SuperQ: `aruba_controller:`

RichiH:
Thanks; I pushed a fixed version.

SuperQ:
It might be nice to simplify the `walk:` section.  But if it's too much of a performance hit to walk larger trees, this is fine.

RichiH:
I don't have data on that and the test servers are down. I suggest we play with that during 33c3 and merge a cleanup after testing it.

brian-brazil:
The generator is at a PoC stage at https://github.com/prometheus/snmp_exporter/tree/generator, all future additions will be via it.

brian-brazil:

PR #96, opened 2016-12-21
Update snmp.yml to generated version.

brian-brazil:
apcups version 1 manually added.

@RichiH @SuperQ @elisiano @jcollie 

I've checked all these, but may have missed something. Some have had an increase in the number of metrics returned.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -1,1646 +1,2431 @@
----
-# Default module: interface stats and uptime.
-default:
-  walk:
-    - 1.3.6.1.2.1.1.3
-    - 1.3.6.1.2.1.2
-    - 1.3.6.1.2.1.31.1.1
-  metrics:
-    - name: sysUpTime
-      oid: 1.3.6.1.2.1.1.3
-    - name: ifNumber
-      oid: 1.3.6.1.2.1.2.1
-    - name: ifMtu
-      oid: 1.3.6.1.2.1.2.2.1.4
-      indexes:
-        - labelname: ifDescr
-          type: Integer32
-      lookups:
-        - labels: [ifDescr]
-          labelname: ifDescr
-          oid: 1.3.6.1.2.1.2.2.1.2
-    - name: ifSpeed
-      oid: 1.3.6.1.2.1.2.2.1.5
-      indexes:
-        - labelname: ifDescr
-          type: Integer32
-      lookups:
-        - labels: [ifDescr]
-          labelname: ifDescr
-          oid: 1.3.6.1.2.1.2.2.1.2
-    - name: ifAdminStatus
-      oid: 1.3.6.1.2.1.2.2.1.7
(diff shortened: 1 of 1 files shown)


SuperQ:
Lookups that replace index labels are invalid.  They are not guaranteed unique.  Please use the original index OID names (ex. `ifEntry`).

These string lookup labels are valid for aggregations, for example in wifi networking we have traffic by SSID, but there are sometimes multiple radios per target handling the same SSID.  Therefore we need the lookup strings in able to aggregate by SSID.  The SNMP index only includes the radio ID.

SuperQ:
Here's a concrete example:

MIB Entry:
```
ruckusWLANChannel OBJECT-TYPE
    SYNTAX  INTEGER (1..14)
    MAX-ACCESS read-only
    STATUS current
    DESCRIPTION
        "Specifies the current operating channel."
    ::= { ruckusWLANEntry 10 }
```

Yaml:
```
    - name: ruckusWLANChannel # Specifies the current operating channel.
      oid: 1.3.6.1.4.1.25053.1.1.6.1.1.1.1.1.10
      indexes:
        - labelname: ruckusWLANEntry
          type: Integer32
      lookups:
        - labels: [ruckusWLANEntry]
          labelname: ruckusWLANSSID
          oid: 1.3.6.1.4.1.25053.1.1.6.1.1.1.1.1.1
```

Metrics:
```
ruckusWLANChannel{ruckusWLANEntry="23",ruckusWLANSSID="Test"} 6
ruckusWLANChannel{ruckusWLANEntry="24",ruckusWLANSSID="Guests Test"} 6
ruckusWLANChannel{ruckusWLANEntry="25",ruckusWLANSSID="Test 2"} 6
ruckusWLANChannel{ruckusWLANEntry="26",ruckusWLANSSID="Test"} 48
ruckusWLANChannel{ruckusWLANEntry="27",ruckusWLANSSID="Guests Test"} 48
ruckusWLANChannel{ruckusWLANEntry="28",ruckusWLANSSID="Test 2"} 48
ruckusWLANChannel{ruckusWLANEntry="6",ruckusWLANSSID=""} 6
ruckusWLANChannel{ruckusWLANEntry="7",ruckusWLANSSID="foobar"} 6
```

brian-brazil:
This is merely a port of what's already there. I only had to remove one lookup as it was incorrect (and redundant).

SuperQ:
So this is still hand-edited, just more in line with the generator formatting?

brian-brazil:
This is generated, just with the extra version line required for APC (though the generator can handle that now). The config should be equivalent, with some extra metrics.

SuperQ:
The config is still invalid. I guess we will need to fix the generator.

brian-brazil:
Make sure you're using a snmp_exporter build from head.

PR #98, opened 2016-12-26
Add regexp string to metric handling

SuperQ:
Add a regexp-based metric handling option to pull numeric data out of
string metrics.

Example use:
```
wlsx_systemext:
  walk:
    - 1.3.6.1.4.1.14823.2.2.1.2.1
  metrics:
    - name: wlsxSysExtInternalTemparature
      oid: 1.3.6.1.4.1.14823.2.2.1.2.1.10
      regex: ([0-9]+\.[0-9]+) degrees Celsius .*
      indexes:
        - labelname: wlsxSystemExtGroup
          type: Integer32
```

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -154,7 +154,12 @@ PduLoop:
 			}
 			if head.metric != nil {
 				// Found a match.
-				ch <- pduToSample(oidList[i+1:], &pdu, head.metric, oidToPdu)
+				m, err := pduToMetric(oidList[i+1:], &pdu, head.metric, oidToPdu)
+				if err != nil {
+					log.Errorf("Error converting pdu to sample %s", err)
+				} else {
+					ch <- m
+				}
 				break
 			}
 		}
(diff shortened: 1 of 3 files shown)


SuperQ:
Yes, this is ugly.  Vendors do questionable things with their SNMP values.  This will also be needed when we add the generic string metric to label handling.

RichiH:
This would allow us to run an unpatched version a time 33c3 so all for it :)

brian-brazil:
Any config option needs to come from the generator, and that means figuring out how per-oid settings are going to work as I'm expecting a few things there like type overrrides.

> This will also be needed when we add the generic string metric to label handling.

I'm not seeing why we should need that. I think it'd make sense to implement this feature first, as there are aspects of that change that are needed to do this change properly.


This is a feature I've wanted to hold off on considering, as there's more important features like simple string handling and exclusions that needs to be worked out first so we can get the generator in good shape.

fjaeckel:
I'd love to see the possibility of using regexps on return values of OIDs.

brian-brazil:
Per #150 we'll be doing something a bit more sophisticated for this.

PR #100, opened 2016-12-29
Handle missing label lookups

SuperQ:
When performing lookups across different tables there are cases where
data in the lookup table doesn't match the data in the metric table.

This makes `indexesToLabels()` always return the requested label sets.
Currently you end up with scrape errors due to Prometheus client
miss-matched label sets in metric families.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -312,6 +312,8 @@ func indexesToLabels(indexOids []int, metric *config.Metric, oidToPdu map[string
 		}
 		if pdu, ok := oidToPdu[oid]; ok {
 			labels[lookup.Labelname] = pduValueAsString(&pdu)
+		} else {
+			labels[lookup.Labelname] = ""
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


collector.go
```diff
@@ -312,6 +312,8 @@ func indexesToLabels(indexOids []int, metric *config.Metric, oidToPdu map[string
 		}
 		if pdu, ok := oidToPdu[oid]; ok {
 			labels[lookup.Labelname] = pduValueAsString(&pdu)
+		} else {
+			labels[lookup.Labelname] = "Lookup missing"
```
brian-brazil: An empty string would be better I think.

Can you add a unittest please?

SuperQ: Empty string is also fine.  I'll see about adding a test.

SuperQ:
@brian-brazil It looks like there was already a test for a missing metric, but somehow the testing didn't show it.  I've fixed the results of the test to match the new behavior.

brian-brazil:
Ah I see you've a non-generated config which isn't overwriting labels. The intention was that the original label value would be preserved if the lookup failed.

PR #106, opened 2017-01-10
Add debug logging for SNMP walks

SuperQ:
Add debug level logging for individual SNMP walks.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -67,6 +67,8 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
 	result := []gosnmp.SnmpPDU{}
 	for _, subtree := range config.Walk {
 		var pdus []gosnmp.SnmpPDU
+		log.Debugf("Walking target %q subtree %q", snmp.Target, subtree)
+		walkStart := time.Now()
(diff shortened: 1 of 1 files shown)


collector.go
```diff
@@ -67,13 +67,17 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
...
+		walkStart := time.Now()
 		if snmp.Version == gosnmp.Version1 {
 			pdus, err = snmp.WalkAll(subtree)
 		} else {
 			pdus, err = snmp.BulkWalkAll(subtree)
 		}
 		if err != nil {
 			return nil, fmt.Errorf("Error walking target %s: %s", snmp.Target, err)
+		} else {
+			log.Debug("Walk of target %q subtree %q completed in %s", snmp.Target, subtree, time.Since(walkStart).Seconds())
```
juliusv: You're missing the time unit in the output here. I would just omit the `.Seconds()`, then it will automatically print the right unit.

Also, Debug -> Debugf?

SuperQ: Ahh, yup, Debugf.  Also will fix `.Seconds()`

juliusv:
:+1:

PR #109, opened 2017-01-12
Adding installation of gopkg.in/yaml.v2 to Build Instructions

momorientes:
Might be obvious to someone used to go.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -7,6 +7,7 @@ This config generator uses NetSNMP to parse MIBs, and generates configs for the 
 
 ```
 sudo apt-get install build-essential libsnmp-dev snmp-mibs-downloader  # Debian-based distros
+go get gopkg.in/yaml.v2
 go build
 ```
 


brian-brazil:
This should not be required, it should be picked up from the vendor directory.

momorientes:
didn't work for me just now. Fairly new to golang tough, so not sure if I did something wrong.
All I did was: 
  - clone the repo
  - cd generator
  - apt-get install ...
  - go build

brian-brazil:
Ah, so your GOPATH isn't setup. You should do that and make sure the repo is at the right path

RichiH:
Arguably, this PR makes a good case to put a pre-built snmp_exporter on the Prometheus download page. Normal usage should not require a Go environment in the first place.

brian-brazil:
We can't pre-build it, as it depends on the netsnmp you have installed on your machine.

momorientes:
Maybe adding a section about correctly setting up a Go environment on the [getting started](https://prometheus.io/docs/introduction/getting_started/) page would prevent further issues as prometheus continues to grow.

PR #113, opened 2017-01-20
Add support for strings

brian-brazil:
This exposes strings as a label value on a 1-valued gauge.

@SuperQ @RichiH

### DIFF
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -74,6 +74,11 @@ func prepareTree(nodes *Node) map[string]*Node {
 
 	// Set type on MAC addresses.
 	walkNode(nodes, func(n *Node) {
+		// For some odd reason ifPhysAddress's MIB isn't being parsed correctly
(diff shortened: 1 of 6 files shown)


generator/tree.go
```diff
@@ -74,6 +74,11 @@ func prepareTree(nodes *Node) map[string]*Node {
 
 	// Set type on MAC addresses.
 	walkNode(nodes, func(n *Node) {
+		// For some odd reason ifPhysAddress's MIB isn't bein parsed correctly
```
RichiH: Being.

RichiH:
Thanks; I just skimmed over it and will try to find time to test over the weekend.

generator/tree.go
brian-brazil: That's the Dublin accent finally getting to me :)

RichiH:
Belated "my tests looked good".

Issue #80, opened 2016-10-21
Allow configurable repeat value


brian-brazil:
Currently it's hardcoded to 25 which I thought would be safe given that other monitoring tools seem to use this value, however I've had a report of a switch breaking at 23. This should be made configurable per-module.

Issue #17, opened 2016-01-04
Timeout definition for snmp_exporter


RichiH:
While I can set a timeout for the scraping of the exporter, I can't set a timeout for when the snmp_exporter stops trying, internally. I can't seem to find an option within the exporter codebase, either.

Unless I am missing something, an option to have snmp_exporter jobs kill themselves after X time would be appreciated.

brian-brazil:
There's no such option currently, but I'd be happy to accept a PR.

brian-brazil:
There's an option now, however it's not of too much use. The upstream library needs improvements in the timeout department, preferably by using contexts.

brian-brazil:
The other option is to use the raw BulkGet calls that the library exposes, which would probably be simpler.

SuperQ:
We can how handle this like we do in the blackbox_exporter[0] and get the timeout from `X-Prometheus-Scrape-Timeout-Seconds` headers.

[0]: https://github.com/prometheus/blackbox_exporter/blob/master/main.go#L74-L86

brian-brazil:
The underlying snmp library needs better timeout handling first.

brian-brazil:
https://github.com/soniah/gosnmp/pull/203 will permit us to do this.

PR #115, opened 2017-01-26
Allows snmp retries, timeout, and maxrepititions to be specified in the snmp.yml file.

didorothy:
@matthiasr This relates to Issues #17 and #80 as well as adding an option that was not specified by those tickets since it was related and simple (retries).

Basically just adds retries, max_repititions, and timeout options to the snmp.yml file. These are defined per config.Module so that it can be changed for different device types.

An example snmp.yml file would look like the following:
```yaml
default:
  version: 2
  max_repititions: 25
  timeout: 60
  retries: 3
  walk:
    - 1.3.6.1.2.1.1.3
...
```

The default values for when these are omitted are the same as currently in master. The timeout value is in seconds. So in the example above it would timeout after 60 seconds. I ran a brief test on Windows but will likely be using this on a Linux machine shortly (as I understand it should not matter).

My specific needs relate to timeout because I have a device or two that takes a little longer to respond.

If any adjustments are needed let me know. So far I have just dabbled with Go code so I may have made some mistakes. Also, I did not see unit tests for the configuration and don't really have time to add them right now. If you would like some before accepting this let me know. I will need you to provide some guidance on how to make sure it will integrate with your CI server (links are fine). I've not done unit tests for Go before, my experience is mostly with Python & .NET.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -39,9 +39,9 @@ func oidToList(oid string) []int {
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
-	snmp.Retries = 3
-	snmp.MaxRepetitions = 25
-	snmp.Timeout = time.Second * 60
+	snmp.Retries = config.Retries
+	snmp.MaxRepetitions = config.MaxRepititions
+	snmp.Timeout = time.Second * config.Timeout
 
 	snmp.Target = target
 	snmp.Port = 161
--- a/collector.go
+++ b/collector.go
@@ -39,9 +39,9 @@ func oidToList(oid string) []int {
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
-	snmp.Retries = config.Retries
 	snmp.MaxRepetitions = config.MaxRepititions
-	snmp.Timeout = time.Second * config.Timeout
+	snmp.Retries = config.Retries
+	snmp.Timeout = config.Timeout * time.Duration(snmp.Retries)
 
 	snmp.Target = target
 	snmp.Port = 161
--- a/collector.go
+++ b/collector.go
@@ -40,6 +40,7 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
 	snmp.MaxRepetitions = config.MaxRepititions
+	// user specifies timeout of each retry attempt but GoSNMP expects total timeout for all attemtps
 	snmp.Retries = config.Retries
 	snmp.Timeout = config.Timeout * time.Duration(snmp.Retries)
 
--- a/collector.go
+++ b/collector.go
@@ -40,7 +40,7 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
 	snmp.MaxRepetitions = config.MaxRepititions
-	// user specifies timeout of each retry attempt but GoSNMP expects total timeout for all attemtps
+	// User specifies timeout of each retry attempt but GoSNMP expects total timeout for all attemtps.
 	snmp.Retries = config.Retries
 	snmp.Timeout = config.Timeout * time.Duration(snmp.Retries)
 
--- a/config/config.go
+++ b/config/config.go
@@ -26,9 +26,9 @@ func LoadFile(filename string) (*Config, error) {
 var (
 	DefaultModule = Module{
 		Version:        2,
-		Retries:        3,
 		MaxRepititions: 25,
-		Timeout:        60,
+		Retries:        3,
+		Timeout:        time.Second,
 	}
 	DefaultAuth = Auth{
 		Community:     "public",
@@ -47,8 +47,8 @@ type Module struct {
 	Metrics []*Metric `yaml:"metrics"`
 
 	Version        int           `yaml:"version,omitempty"`
-	Retries        int           `yaml:"retries,omitempty"`
 	MaxRepititions uint8         `yaml:"max_repititions,omitempty"`
+	Retries        int           `yaml:"retries,omitempty"`
 	Timeout        time.Duration `yaml:"timeout,omitempty"`
 	Auth           *Auth         `yaml:"auth,omitempty"`
 
--- a/config/config.go
+++ b/config/config.go
@@ -28,7 +28,7 @@ var (
 		Version:        2,
 		MaxRepititions: 25,
 		Retries:        3,
-		Timeout:        time.Second,
+		Timeout:        time.Second * 20,
 	}
 	DefaultAuth = Auth{
 		Community:     "public",
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"strings"
+	"time"
 
 	"github.com/soniah/gosnmp"
 	"gopkg.in/yaml.v2"
@@ -24,7 +25,10 @@ func LoadFile(filename string) (*Config, error) {
 
 var (
 	DefaultModule = Module{
-		Version: 2,
+		Version:        2,
+		Retries:        3,
+		MaxRepititions: 25,
+		Timeout:        60,
 	}
 	DefaultAuth = Auth{
 		Community:     "public",
@@ -42,8 +46,11 @@ type Module struct {
 	Walk    []string  `yaml:"walk"`
 	Metrics []*Metric `yaml:"metrics"`
 
-	Version int   `yaml:"version,omitempty"`
-	Auth    *Auth `yaml:"auth,omitempty"`
+	Version        int           `yaml:"version,omitempty"`
+	Retries        int           `yaml:"retries,omitempty"`
+	MaxRepititions uint8         `yaml:"max_repititions,omitempty"`
+	Timeout        time.Duration `yaml:"timeout,omitempty"`
+	Auth           *Auth         `yaml:"auth,omitempty"`
 
 	XXX map[string]interface{} `yaml:",inline"`
 }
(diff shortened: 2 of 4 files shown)


collector.go
```diff
@@ -39,9 +39,9 @@ func oidToList(oid string) []int {
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
-	snmp.Retries = 3
-	snmp.MaxRepetitions = 25
-	snmp.Timeout = time.Second * 60
+	snmp.Retries = config.Retries
+	snmp.MaxRepetitions = config.MaxRepititions
+	snmp.Timeout = time.Second * config.Timeout
```
brian-brazil: This doesn't work quite the way you'd think, divide it by the Retries.

collector.go
```diff
@@ -39,9 +39,9 @@ func oidToList(oid string) []int {
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
-	snmp.Retries = 3
-	snmp.MaxRepetitions = 25
-	snmp.Timeout = time.Second * 60
+	snmp.Retries = config.Retries
+	snmp.MaxRepetitions = config.MaxRepititions
```
brian-brazil: Put this above the timeout settings to avoid confusion

config/config.go
```diff
@@ -24,7 +25,10 @@ func LoadFile(filename string) (*Config, error) {
 
 var (
 	DefaultModule = Module{
-		Version: 2,
+		Version:        2,
+		Retries:        3,
+		MaxRepititions: 25,
+		Timeout:        60,
```
brian-brazil: We should drop this to a second as it turns out to be a per-query timeout. Shouldn't this be a time.Duration?

collector.go
didorothy: Which line are you referring to? Github is not clear about which line you are referring to. Are you considering the `snmp.Retries` value as part of the timeout settings? The order was kept the same as it is in master.

brian-brazil: Retries is related to timeout. Now that this is exposed to users it should be cleaned up.

collector.go
didorothy: As I understand it, up to `snmp.Retries` are made within the time specified by `snmp.Timeout`. Should I understand this differently? It looks to me like it uses the `snmp.Timeout` to set a final timeout then the GoSNMP library divides the `snmp.Timeout` value by the `snmp.Retries` value for the amount of time to wait for a response to an individual attempt. Technically speaking the GoSNMP library may not fulfill all the specified attempts because the final timeout could occur prior to all of the requests being performed (granted it is unlikely). [See here.](https://github.com/soniah/gosnmp/blob/master/marshal.go#L105)

If my understanding is correct then what you are really saying is that you think that a Timeout specified in a config file should do what the user expects (which you seem to think is time allowed per attempt) and not follow the rules of the underlying code. If that is what you'd prefer it can be done that way. Let me know.

brian-brazil: > As I understand it, up to snmp.Retries are made within the time specified by snmp.Timeout.

That's correct.

> If my understanding is correct then what you are really saying is that you think that a Timeout specified in a config file should do what the user expects (which you seem to think is time allowed per attempt) and not follow the rules of the underlying code.

Yes.

config/config.go
didorothy: > We should drop this to a second as it turns out to be a per-query timeout. 

You should first review my comments on `collector.go` lines 42-43. If you would like the change described there then yes what you say makes more sense though I would suggest 20 seconds as the current hard coded values called for 3 retries in 60 seconds.

> Shouldn't this be a time.Duration?

In the struct I specified it as a `time.Duration` which is a, pardon the non-Go description, renaming of int64. As I said I'm not a Go expert. I had initially attempted to use `int` but found that Go did not want to do automatic type casting since it was not a constant (60 is a constant so I assume that the compiler automatically does the type cast at compilation time). To me the more natural use of a timeout is to specify it in seconds. Granted it could be done in milliseconds or nanoseconds (nanoseconds is the smallest increment of `time.Duration`) but I would expect most people would not expect an increment of less than a second on a request/response connection like this. For your convenience: [see here](https://golang.org/pkg/time/#Duration).

didorothy:
I'll work on the generator code though I have never used it so I'm not sure exactly what it does or how it functions. I've always just looked up the MIBs I need and run queries directly with a python tool or with snmpwalk. In fact I have built some customized Module definitions in my snmp.yml file that are customized to my needs manually. I'll try to add another commit soon.

config/config.go
brian-brazil: > I would suggest 20 seconds as the current hard coded values called for 3 retries in 60 seconds.

The timeout is per query, not overall. It'd be a very unusual network that had more than 1s latency locally.

> I had initially attempted to use int but found that Go did not want to do automatic type casting since it was not a constant (60 is a constant so I assume that the compiler automatically does the type cast at compilation time). 

Try `time.Second` as the default.

brian-brazil:
Plumb through these values the same way auth is done. You don't have to touch the grungy bits of the generator.

didorothy:
I was unable to build on Windows (lack of gcc, and probably net_snmp.h related files as well) but I did not get any linting messages on the generator code indicating a problem.

config/config.go
didorothy: Just to clarify, if you specify the value as a full `time.Duration` and not a multiplier value for `time.Second` then the value in the `snmp.yaml` file will have to be specified as the following to be 1 second:

```yaml
default:
  timeout: 1000000000
...
```

In my opinion, this would be very annoying to the end user. Do you want the configuration to work that way or for the value to be specified in the `snmp.yml` file in terms of a multiplier of a larger increment like `time.Second` or `time.Millisecond`? 

If you are confused by my (mis)use of the `time.Duration` type I can change the type in `config.Module` to be `int64` and then do appropriate run-time casting. In my mind doing a cast from `int64` to `time.Duration` is extra work that the yaml tool is taking care of for me. Though I would agree that it could add clarity to the code to do so.

brian-brazil: Hmm, look at how we do it in Prometheus itself: https://github.com/prometheus/prometheus/blob/master/config/config.go

didorothy: Ah, thanks! The piece I was missing was that yaml.v2 provides a more convenient way of interpreting `time.Duration` using a string value as opposed to just an integer value. Sorry for the misunderstanding. I had not looked very deeply at how yaml.v2 worked. I'll have these adjustments committed shortly.

didorothy:
I believe that I have addressed all of the above discussion in this latest commit. If I missed something let me know. Thanks for the help.

collector.go
```diff
@@ -39,9 +39,9 @@ func oidToList(oid string) []int {
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
-	snmp.Retries = 3
-	snmp.MaxRepetitions = 25
-	snmp.Timeout = time.Second * 60
+	snmp.MaxRepetitions = config.MaxRepititions
+	snmp.Retries = config.Retries
+	snmp.Timeout = config.Timeout * time.Duration(snmp.Retries)
```
brian-brazil: Could you add a comment saying why we do this?

brian-brazil:
I've tested the generator on my Linux box and it all seems fine.

SuperQ:
I'm trying to understand the change to a 1s timeout.  Many SNMP devices are _very_ slow at responding to walk requests.

brian-brazil:
Do you have devices which take 1s to respond to a single request?

SuperQ:
Yes, all the time, for example the aruba wifi controllers can easily take 30-45s to respond to a single walk.  This one we'll add custom timeouts into the generator.

Another example is a basic 48-port switch can take ~1.5s to respond to `1.3.6.1.2.1.2` or `1.3.6.1.2.1.31.1.1`.  I've seen stacked switches take a couple of seconds to walk as well.

I think a 5-10s would be a safer per-walk default timeout.

brian-brazil:
> Another example is a basic 48-port switch can take ~1.5s to respond to 1.3.6.1.2.1.2 or 1.3.6.1.2.1.31.1.1.

Mine's taking about a millisecond (hard to measure exactly as the startup time of snmpbulkget is dominating).

collector.go
didorothy: Would the following comment above line 43 meet what you are looking for?

> timeout settings, user specifies timeout of each retry attempt but GoSNMP expects total timeout for all attemtps

brian-brazil: What's after the comma is enough

didorothy:
@SuperQ for what it is worth `snmpwalk` on Ubuntu 16.04 has a default timeout of 1 second between retries. I have a Cisco smart switch (SG200-50) that takes as long as 2.5 minutes (average seems to be around 40 seconds) to send all networking information but it does not fail to send all the requested data when using the `snmpwalk` command using that default timeout. Also, the timeout is per GetNext or GetBulk request not the timeout for the length of the walk. As long as each GetNext or GetBulk call completes in a timely manner there is no issue with a long running walk.

collector.go
didorothy: I've added the comment.

SuperQ:
@didorothy Thanks for the data. I'm just worried that drastically changing the defaults (60s -> 1s) is going to be problematic for `snmp_exporter` users.  I don't have enough real-world sample data to say for sure.

collector.go
```diff
@@ -39,9 +39,10 @@ func oidToList(oid string) []int {
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
-	snmp.Retries = 3
-	snmp.MaxRepetitions = 25
-	snmp.Timeout = time.Second * 60
+	snmp.MaxRepetitions = config.MaxRepititions
+	// user specifies timeout of each retry attempt but GoSNMP expects total timeout for all attemtps
```
brian-brazil: Start with a capital letter and end with a period

brian-brazil:
It seems the timeout doesn't work the way I think, so we should put it back at 20s which is the same as now.

didorothy:
I have corrected the grammar on the comment and the default timeout to 20 seconds.

brian-brazil:
Thanks!

PR #118, opened 2017-02-04
Represent OctetStrings as hex

### LINKED ISSUE #116: Problem with invalid characters in label when polling SNMP object with syntax BITS
Using a build of snmp_exporter/generate from this morning (commit 888db958efb39ced5a717cc67a213d2805a9706c) I get the following entry when I generate my snmp.yml (using CISCO-STACKWISE-MIB).

```
  - name: cswEnableIndividualStackNotifications
    oid: 1.3.6.1.4.1.9.9.500.1.1.5
    type: string
```

This is what the metric looks like when polled from the command line:

```
# HELP cswEnableIndividualStackNotifications 
# TYPE cswEnableIndividualStackNotifications gauge
cswEnableIndividualStackNotifications{cswEnableIndividualStackNotifications="<FF><FF><FE>^@"} 1
```

However, when Prometheus (1.4.1) tries to ingest this metric, it gives the error:

```
invalid label value "\xff\xff\xfe\x00"
```

I think that either snmp_exporter should have the ability to ignore specific MIB entries (in both the generator and exporter) or the generator should not create a metric by default for strings that have a syntax of "BITS".

Here is the relevant section of the CISCO-STACKWISE-MIB:

```
cswEnableIndividualStackNotifications OBJECT-TYPE
    SYNTAX          BITS {
                        stackPortChange(0),
                        stackNewMaster(1),
                        stackMismatch(2),
                        stackRingRedundant(3),
                        stackNewMember(4),
                        stackMemberRemoved(5),
                        stackPowerLinkStatusChanged(6),
                        stackPowerPortOperStatusChanged(7),
                        stackPowerVersionMismatch(8),
                        stackPowerInvalidTopology(9),
                        stackPowerBudgetWarning(10),
                        stackPowerInvalidInputCurrent(11),
                        stackPowerInvalidOutputCurrent(12),
                        stackPowerUnderBudget(13),
                        stackPowerUnbalancedPowerSupplies(14),
                        stackPowerInsufficientPower(15),
                        stackPowerPriorityConflict(16),
                        stackPowerUnderVoltage(17),
                        stackPowerGLS(18),
                        stackPowerILS(19),
                        stackPowerSRLS(20),
                        stackPowerSSLS(21),
                        stackMemberToBeReloadedForUpgrade(22)
                    }
    MAX-ACCESS      read-write
    STATUS          current
    DESCRIPTION
        "This object is used to enable/disable individual notifications
        defined in this MIB module. Turning on a particular bit would
        enable the corresponding trap and, similarly, turning off a
        particular bit would disable the corresponding trap. The
        following notifications are controlled by this object:

         stackPortChange(0): enables/disables cswStackPortChange
                             notification.

         stackNewMaster(1): enables/disables cswStackNewMember
                            notification.

         stackMismatch(2): enables/disables cswStackMismatch
                           notification.

         stackRingRedundant(3): enables/disables cswStackRingRedundant 
                                notification.

         stackNewMember(4): enables/disables cswStackNewMember 
                            notification.

         stackMemberRemoved(5): enables/disables cswStackMemberRemoved 
                                notification.

         stackPowerLinkStatusChanged(6): enables/disables 
                       cswStackPowerPortLinkStatusChanged notification.

         stackPowerPortOperStatusChanged(7): enables/disables 
                       cswStackPowerPortOperStatusChanged notification.

         stackPowerVersionMismatch(8): enables/disables 
                           cswStackPowerVersionMismatch notification.

         stackPowerInvalidTopology(9): enables/disables 
                           cswStackPowerInvalidTopology notification

         stackPowerBudgetWarning(10): enables/disables 
                        cswStackPowerBudgetWarning notification. 

         stackPowerInvalidInputCurrent(11): enables/disables 
                        cswStackPowerInvalidInputCurrent notification.

         stackPowerInvalidOutputCurrent(12): enables/disables 
                        cswStackPowerInvalidOutputCurrent notification.

         stackPowerUnderBudget(13): enables/disables 
                         cswStackPowerUnderBudget notification.

         stackPowerUnbalancedPowerSupplies(14): enables/disables 
                     cswStackPowerUnbalancedPowerSupplies notification.

         stackPowerInsufficientPower(15): enables/disables 
                         cswStackPowerInsufficientPower notification.

         stackPowerPriorityConflict(16): enables/disables 
                         cswStackPowerPriorityConflict notification.

         stackPowerUnderVoltage(17): enables/disables 
                         cswStackPowerUnderVoltage notification.

         stackPowerGLS(18): enables/disables cswStackPowerGLS
                            notification. 

         stackPowerILS(19): enables/disabled cswStackPowerILS
                            notification.

         stackPowerSRLS(20): enables/disables cswStackPowerSRLS 
                             notification.

         stackPowerSSLS(21): enables/disables cswStackPowerSSLS 
                             notification.
         stackMemberToBeReloadedForUpgrade(22): enables/disables 
                   cswStackMemberToBeReloadedForUpgrade notification." 
    ::= { cswGlobals 5 }
```

brian-brazil:
Fixes #116


mohanprasaths:
Hi @brian-brazil 

I have used this fix and it is working fine for Hex-Strings but it also converts "strings" to "Hex -Strings" too.

_**SNMP walk for Interface Description (ifDescr) (String handling):**_
selection_016

**_SNMP walk for rttMonEchoAdminSourceAddress (IP address as Hex string):_**
selection_012

**_SNMP Exporter output before Fix:_** 
**For IfDescr:**
**It was printing the String perfectly in labels**
selection_013
**If I have them as seperate metric , it just return "0" (_Not getting expected value_):** 
selection_014

**For rttMonEchoAdminSourceAddress :**
**It was printing some garbage value in labels before (_Not expected value_):**
selection_017
**If I have them as seperate metric , it just return "0" (_Not expected value_):** 
selection_018

_**SNMP Exporter output After Fix :**_
**For ifDescr:**
**It was printing the String as Hex String in labels (Not expected value)**
selection_009
**If I have them as seperate metric , it just return "0" (_Not getting expected value_):**
selection_010

**For rttMonEchoAdminSourceAddress:**
**It was printing hex values which can be converted to IP :**
selection_008
**If I have them as seperate metric , it just return "0" (_Not expected value_):**
selection_018

So If I apply the fix , all the string are converted to "hex string" .

When I tried to debug this  , all my string are switching to the case "[]byte" instead of "string" ,So I got all my string converted to hex value. 

```
func pduValueAsString(pdu *gosnmp.SnmpPDU) string {
	switch pdu.Value.(type) {
	case int:
		return strconv.Itoa(pdu.Value.(int))
	case uint:
		return strconv.FormatUint(uint64(pdu.Value.(uint)), 10)
	case int64:
		return strconv.FormatInt(pdu.Value.(int64), 10)
	case string:
		if pdu.Type == gosnmp.ObjectIdentifier {
			// Trim leading period.
			return pdu.Value.(string)[1:]
		}
		return pdu.Value.(string)
	case []byte:
		// OctetString
    if len(pdu.Value.([]byte)) == 0 {
      return ""
    }
		return fmt.Sprintf("0x%X", pdu.Value.([]byte))
	case nil:
		return ""
	default:
		// This shouldn't happen.
		log.Infof("Got PDU with unexpected type: Name: %s Value: '%s', Go Type: %T SNMP Type: %s", pdu.Name, pdu.Value, pdu.Value, pdu.Type)
		snmpUnexpectedPduType.Inc()
		return fmt.Sprintf("%s", pdu.Value)
	}
}

```
Can you please guide me through this?
Thanks

brian-brazil:
We'll need to plumb through the type from the MIBs then to tell which are DisplayString.

mohanprasaths:
@brian-brazil  Can we get the type of PDU from the bulk walk? 
selection_015

brian-brazil:
No, it's only in the MIB. From the walk we can only tell it's a octet string or a bit string.

brian-brazil:
I've done a first pass at that. Still needs the tests to be fixed+added, and a cleanup+refactor.

brian-brazil:
I've cleaned this all up, it should handle everything now.

Still unsure why the hints on ifTable are getting lost.

SuperQ:
CC @RichiH, if you have time to test this, that'd be helpful.

brian-brazil:
The the answer is that RFC1213-MIB is missing type hints.

brian-brazil:
Noone has indicated this is badly broken (though I have my doubts about NETADDR), so let's get this in to unbreak head and see what crops up. Would someone mind reviewing the code?

RichiH:
I can't really review (other than typos) and am totally stuck in different work atm :/

brian-brazil:
I've done a bit more testing against the equipment I have, seems to be all be working so I'm going to merge now rather than delaying further. We can deal with any issues as they come up.

rrix:
I updated my docker container to `prom/snmp-exporter:master` to pull this change in, and since then, it dies  with:

```
Mar 20 18:00:13 hypervisor01 docker[49086]: time="2017-03-21T01:00:13Z" level=fatal msg="Unknown index type string" source="collector.go:355"
```

My snmp.yml was generated by the generator script and is attached: [snmp.yml.txt](https://github.com/prometheus/snmp_exporter/files/856811/snmp.yml.txt)

I can provide any MIB files and the generator configuration if that is a more useful format here, but they are mostly from https://files.pfsense.org/misc/mibs/

AlexRRR:
https://github.com/prometheus/snmp_exporter/pull/118#issuecomment-277936972

@brian-brazil  Does this mean that where we find in the generated file an OID with type `OctetString` we have to manually (according to the MIB file) interpret if  we should change the type to `DisplayString` if it is not a `Bit` string? 

Or am I missing something?

brian-brazil:
Everything comes from the MIB.

PR #124, opened 2017-02-16
Replace libsnmp with libnetsnmp, add /usr/local prefix

aduitsis:
According to the issue I've opened yesterday, these are two simple changes that replace -lsnmp with -lnetsnmp and add -L/usr/local/lib to the cgo ldflags and also -I/usr/local/include to the cgo cflags. These changes do not appear to break builds in Linux, but at the same time make building on FreeBSD possible without patching the code.

### DIFF
--- a/generator/net_snmp.go
+++ b/generator/net_snmp.go
@@ -1,7 +1,8 @@
 package main
 
 /*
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

PR #133, opened 2017-03-16
Fixing a spelling error introduced in snmp_options pull request.

didorothy:
I discovered that I introduced a spelling error when I made my last pull request. This causes consistency issues both in the snmp_exporter and with the underlying go library used for making the connections and transferring data.

I believe that this request fixes all of those misspellings. `MaxRepititions` is now consistently `MaxRepetitions`.


brian-brazil:
Thanks!

PR #139, opened 2017-03-27
vendor: update gosnmp

euank:
Notably to pickup https://github.com/soniah/gosnmp/pull/92

This was necessary to get the exporter to work against a tripplite PDU.

Without it, the scrape requests timeout, and I observed `.1.3.6.1.4.1.850.10.1.2.7.4.0=[inetaddr len!=4]` when I tcpdumped snmp traffic due to the value of the ipaddress being wrong (snmpwalk prints it as 'IpAddress: 2.0.0.0').


brian-brazil:
Thanks!

PR #141, opened 2017-03-29
Update config.go

kariedo:
@brian-brazil 
Fix 'authNoPriv' misspell.
This prevents 'authNoPriv' from working.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,8 +76,8 @@ func (c *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			return fmt.Errorf("Auth username is missing, required for SNMPv3")
 		}
 		if c.Auth.SecurityLevel != "authPriv" &&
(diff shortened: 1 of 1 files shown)


brian-brazil:
This doesn't look quite right, given this string is used elsewhere in the file and follows our docs.

kariedo:
@brian-brazil I agree that this string is spelled differently in parts of the copy, but my hot-fix gets it working at least.
config check doesn't allow 'authNoPriv' to be placed in the snmp.yml:
`c.Auth.SecurityLevel != "noAuthPriv" && c.Auth.SecurityLevel != "noAuthNoPriv" {`
where we expect later:
`case "authNoPriv":`

brian-brazil:
Finally had a chance to dig into this, this is the only place where it's wrong so this is the correct fix. Thanks!

PR #143, opened 2017-04-05
community and version can be passed by get param

Griesbacher:
Hi @brian-brazil, 
this patch allows Prometheus to pass a custom version and community per request, which makes it more flexible, due to overriding the values from the config. I'm not sure if this is breaking the base concept of this tool but maybe it's usefully for others, too.
It's just makes sense for snmp v1/2.
Best Regards,
Philip

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/prometheus/common/version"
 
 	"github.com/prometheus/snmp_exporter/config"
+	"strconv"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


brian-brazil:
If you need different communitys and versions, you should create a module each.

Griesbacher:
Yes, I understood that, but if you got alot of them, your config explodes.

brian-brazil:
Usually you'd have at most a password per type of device, which you'd need a module for anyway.

There's no plans to allow url parameters to override module configuration.

Griesbacher:
OK, I've got no problem with that, we just recently had this problem.

kdessai19:
@Griesbacher  this was much needed. what's the usage command to include community parameter in the execution stage

PR #145, opened 2017-04-12
Make docs easier for those not familiar with Go

brian-brazil:
@SuperQ 

The $HOME/go is a go1.8 default.


RichiH:
LGTM

Richard

Sent by mobile; excuse my brevity and the wall of text Gmail appends by
default.

PR #147, opened 2017-04-22
snmp: Add ddwrt module

daenney:
The list of SNMP OIDs to care about for DD-WRT can be found here: https://www.dd-wrt.com/wiki/index.php/SNMP#Known_OID.C2.B4s_via_SNMP

This works with routers running DD-WRT. It includes the usual things like network interfaces from SNMPv2-SMI and some information about the device itself through HOST-RESOURCES-MIB. I left out support for `hrSWRun`, `hrSWRunPerf`, `hrSWInstalled` and `hrMIBAdminInfo` as they seem bad fits for what Prometheus would care about and in the case of my device don't contain any data.

The `1.3.6.1.4.1.2021` bits come from the UCD-SNMP-MIB, part of Net-SNMP which is what appears to be the SNMPd on these devices:
* `.4`: memory
* `.10`: laTable (load information)
* `.11`: systemStats
* `.255`: unknown (used by DD-WRT to store wireless client info so I had to make up names for it)

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -17,7 +17,7 @@ modules:
 
 # Cicso Wireless LAN Controller
   cisco_wlc:
-    walk: 
+    walk:
       - interfaces
       - ifXTable
       - 1.3.6.1.4.1.14179.2.1.1.1.38   # bsnDot11EssNumberofMobileStations
@@ -103,21 +103,21 @@ modules:
 
 # Palo Alto Firewalls
 #
-# Palo Alto MIBs can be found here: 
+# Palo Alto MIBs can be found here:
 # https://www.paloaltonetworks.com/documentation/misc/snmp-mibs.html
-# 
+#
 # PanOS 7.0 enterprise MIBs:
 # https://www.paloaltonetworks.com/content/dam/pan/en_US/assets/zip/technical-documentation/snmp-mib-modules/PAN-MIB-MODULES-7.0.zip
 #
 # Tested on a Palo Alto Networks PA-3020 series firewall
 #
   paloalto_fw:
-    walk: 
-      - sysUpTime 
-      - interfaces  
-      - hrDevice 
-      - hrSystem 
-      - hrStorage 
+    walk:
+      - sysUpTime
+      - interfaces
+      - hrDevice
+      - hrSystem
+      - hrStorage
       - 1.3.6.1.4.1.25461.2.1.2.1 # panSys
       - 1.3.6.1.4.1.25461.2.1.2.3 # panSession
       - 1.3.6.1.4.1.25461.2.1.2.5 # panGlobalProtect
@@ -132,7 +132,7 @@ modules:
 # Tested on Arista DCS-7010T-48 switch
 #
   arista_sw:
-    walk: 
+    walk:
       - sysUpTime
       - interfaces
       - ifXTable
@@ -175,3 +175,26 @@ modules:
         new_index: diskID
       - old_index: raidIndex
         new_index: raidName
+
+# DD-WRT
+#
+# The list of SNMP OIDs to care about for DD-WRT can be found here: https://www.dd-wrt.com/wiki/index.php/SNMP#Known_OID.C2.B4s_via_SNMP
+#
+# Tested on  DD-WRT v3.0-r31825 (04/06/17) with an ASUS RT-AC66U
+#
+  ddwrt:
+    walk:
+      - sysUpTime
+      - interfaces
+      - ifXTable
+      - 1.3.6.1.2.1.25.2 # hrStorage
+      - 1.3.6.1.4.1.2021.4 # memory
+      - 1.3.6.1.4.1.2021.10.1.1 # laIndex
+      - 1.3.6.1.4.1.2021.10.1.2 # laNames
+      - 1.3.6.1.4.1.2021.10.1.5 # laLoadInt
+      - 1.3.6.1.4.1.2021.11 # systemStats
+    lookups:
+    - old_index: ifIndex
+      new_index: ifDescr
+    - old_index: laIndex
+      new_index: laNames
--- a/generator/README.md
+++ b/generator/README.md
@@ -33,12 +33,14 @@ Put the extracted mibs in a location NetSNMP can read them from. `$HOME/.snmp/mi
 * Fortinet: http://kb.fortinet.com/kb/documentLink.do?externalID=11607
 * Servertech: ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3.mib
 * Palo Alto PanOS 7.0 enterprise MIBs: https://www.paloaltonetworks.com/content/dam/pan/en_US/assets/zip/technical-documentation/snmp-mib-modules/PAN-MIB-MODULES-7.0.zip
-* Arista Networks: https://www.arista.com/assets/data/docs/MIBS/ARISTA-ENTITY-SENSOR-MIB.txt  
+* Arista Networks: https://www.arista.com/assets/data/docs/MIBS/ARISTA-ENTITY-SENSOR-MIB.txt
                    https://www.arista.com/assets/data/docs/MIBS/ARISTA-SW-IP-FORWARDING-MIB.txt
                    https://www.arista.com/assets/data/docs/MIBS/ARISTA-SMI-MIB.txt
 * Synology: http://dedl.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip
 
 https://github.com/librenms/librenms/tree/master/mibs can also be a good source of MIBs.
+
+* UCD-SNMP-MIB (Net-SNMP): http://www.net-snmp.org/docs/mibs/UCD-SNMP-MIB.txt
 
 
 ## Exporter file format
--- a/snmp.yml
+++ b/snmp.yml
@@ -2021,6 +2021,723 @@ cisco_wlc:
       labelname: bsnAPName
       oid: 1.3.6.1.4.1.14179.2.2.1.1.3
       type: OctetString
+ddwrt:
+  walk:
+  - 1.3.6.1.2.1.1.3
+  - 1.3.6.1.2.1.2
+  - 1.3.6.1.2.1.25.2
+  - 1.3.6.1.2.1.31.1.1
+  - 1.3.6.1.4.1.2021.10.1.1
+  - 1.3.6.1.4.1.2021.10.1.2
+  - 1.3.6.1.4.1.2021.10.1.5
+  - 1.3.6.1.4.1.2021.11
+  - 1.3.6.1.4.1.2021.4
+  metrics:
+  - name: sysUpTime
+    oid: 1.3.6.1.2.1.1.3
+    type: gauge
+  - name: ifNumber
+    oid: 1.3.6.1.2.1.2.1
+    type: gauge
+  - name: ifIndex
+    oid: 1.3.6.1.2.1.2.2.1.1
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifDescr
+    oid: 1.3.6.1.2.1.2.2.1.2
+    type: DisplayString
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifType
+    oid: 1.3.6.1.2.1.2.2.1.3
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifMtu
+    oid: 1.3.6.1.2.1.2.2.1.4
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifSpeed
+    oid: 1.3.6.1.2.1.2.2.1.5
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifPhysAddress
+    oid: 1.3.6.1.2.1.2.2.1.6
+    type: PhysAddress48
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifAdminStatus
+    oid: 1.3.6.1.2.1.2.2.1.7
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifOperStatus
+    oid: 1.3.6.1.2.1.2.2.1.8
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifLastChange
+    oid: 1.3.6.1.2.1.2.2.1.9
+    type: gauge
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifInOctets
+    oid: 1.3.6.1.2.1.2.2.1.10
+    type: counter
+    indexes:
+    - labelname: ifDescr
+      type: gauge
+    lookups:
+    - labels:
+      - ifDescr
+      labelname: ifDescr
+      oid: 1.3.6.1.2.1.2.2.1.2
+      type: DisplayString
+  - name: ifInUcastPkts
+    oid: 1.3.6.1.2.1.2.2.1.11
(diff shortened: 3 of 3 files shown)


daenney:
This is the info I get from my DD-WRT from the UCDAVIS MIB, in case someone wants to check I didn't botch the types (I've 00-ified the MACs):

```
UCD-SNMP-MIB::memIndex.0 = INTEGER: 0
UCD-SNMP-MIB::memErrorName.0 = STRING: swap
UCD-SNMP-MIB::memTotalSwap.0 = INTEGER: 0 kB
UCD-SNMP-MIB::memAvailSwap.0 = INTEGER: 0 kB
UCD-SNMP-MIB::memTotalReal.0 = INTEGER: 239368 kB
UCD-SNMP-MIB::memAvailReal.0 = INTEGER: 187012 kB
UCD-SNMP-MIB::memTotalFree.0 = INTEGER: 187044 kB
UCD-SNMP-MIB::memMinimumSwap.0 = INTEGER: 16000 kB
UCD-SNMP-MIB::memShared.0 = INTEGER: 0 kB
UCD-SNMP-MIB::memBuffer.0 = INTEGER: 4544 kB
UCD-SNMP-MIB::memCached.0 = INTEGER: 24956 kB
UCD-SNMP-MIB::memSwapError.0 = INTEGER: error(1)
UCD-SNMP-MIB::memSwapErrorMsg.0 = STRING: Running out of swap space (0)
UCD-SNMP-MIB::laIndex.1 = INTEGER: 1
UCD-SNMP-MIB::laIndex.2 = INTEGER: 2
UCD-SNMP-MIB::laIndex.3 = INTEGER: 3
UCD-SNMP-MIB::laNames.1 = STRING: Load-1
UCD-SNMP-MIB::laNames.2 = STRING: Load-5
UCD-SNMP-MIB::laNames.3 = STRING: Load-15
UCD-SNMP-MIB::laLoad.1 = STRING: 0.00
UCD-SNMP-MIB::laLoad.2 = STRING: 0.01
UCD-SNMP-MIB::laLoad.3 = STRING: 0.05
UCD-SNMP-MIB::laConfig.1 = STRING: 12.00
UCD-SNMP-MIB::laConfig.2 = STRING: 12.00
UCD-SNMP-MIB::laConfig.3 = STRING: 12.00
UCD-SNMP-MIB::laLoadInt.1 = INTEGER: 0
UCD-SNMP-MIB::laLoadInt.2 = INTEGER: 1
UCD-SNMP-MIB::laLoadInt.3 = INTEGER: 4
UCD-SNMP-MIB::laErrorFlag.1 = INTEGER: noError(0)
UCD-SNMP-MIB::laErrorFlag.2 = INTEGER: noError(0)
UCD-SNMP-MIB::laErrorFlag.3 = INTEGER: noError(0)
UCD-SNMP-MIB::laErrMessage.1 = STRING:
UCD-SNMP-MIB::laErrMessage.2 = STRING:
UCD-SNMP-MIB::laErrMessage.3 = STRING:
UCD-SNMP-MIB::ssIndex.0 = INTEGER: 1
UCD-SNMP-MIB::ssErrorName.0 = STRING: systemStats
UCD-SNMP-MIB::ssSwapIn.0 = INTEGER: 0 kB
UCD-SNMP-MIB::ssSwapOut.0 = INTEGER: 0 kB
UCD-SNMP-MIB::ssIOSent.0 = INTEGER: 0 blocks/s
UCD-SNMP-MIB::ssIOReceive.0 = INTEGER: 0 blocks/s
UCD-SNMP-MIB::ssSysInterrupts.0 = INTEGER: 477 interrupts/s
UCD-SNMP-MIB::ssSysContext.0 = INTEGER: 36 switches/s
UCD-SNMP-MIB::ssCpuUser.0 = INTEGER: 1
UCD-SNMP-MIB::ssCpuSystem.0 = INTEGER: 2
UCD-SNMP-MIB::ssCpuIdle.0 = INTEGER: 95
UCD-SNMP-MIB::ssCpuRawUser.0 = Counter32: 32953
UCD-SNMP-MIB::ssCpuRawNice.0 = Counter32: 0
UCD-SNMP-MIB::ssCpuRawSystem.0 = Counter32: 49885
UCD-SNMP-MIB::ssCpuRawIdle.0 = Counter32: 1848870
UCD-SNMP-MIB::ssRawInterrupts.0 = Counter32: 9208326
UCD-SNMP-MIB::ssRawContexts.0 = Counter32: 696812
UCD-DLMOD-MIB::dlmodNextIndex.0 = INTEGER: 1
UCD-SNMP-MIB::versionIndex.0 = INTEGER: 1
UCD-SNMP-MIB::versionTag.0 = STRING: 5.0.9
UCD-SNMP-MIB::versionDate.0 = STRING: $Date: 2004/06/20 21:54:28 $
UCD-SNMP-MIB::versionCDate.0 = STRING: Sat Apr 22 21:06:51 2017
UCD-SNMP-MIB::versionIdent.0 = STRING: $Id: versioninfo.c,v 1.1.2.1 2004/06/20 21:54:28 nikki Exp $
UCD-SNMP-MIB::versionConfigureOptions.0 = STRING: \" --quiet --prefix=/tmp/snmp --target=mipsel-linux --host=mipsel '--with-cc=ccache mipsel-linux-uclibc-gcc' --with-ar=mipsel-linux-uclibc-ar --with-endianness=little '--with-cflags=-Os -pipe -mips32r2 -mtune=74kc -mdspr2 -fno-caller-saves  -msoft-float  -mno-branch-likely -minterlink-mips16 -mips16 -DCAN_USE_SYSCTL=1 -ffunction-sections -fdata-sections -Wl,--gc-sections' --enable-mini-agent --disable-debugging --disable-privacy --without-opaque-special-types --with-persistent-directory=/tmp/snmp-persist --with-default-snmp-version=3 --with-sys-contact=root --with-sys-location=Unknown --with-logfile=/dev/null --with-out-transports=UDPIPv6,TCPIPv6,AAL5PVC,IPX,TCP,Unix --enable-shared=no --enable-static --with-gnu-ld --enable-internal-md5 --with-copy-persistent-files=no --without-openssl -sysconfdir=/tmp --with-mib-modules=mibII,host,mibII/ip,mibII/tcp,mibII/udp,mibII/icmp,mibII/var_route,mibII/kernel_linux,ucd_snmp --with-out-mib-modules=host/hr_swrun,agent_mips,agentx,notification,utilities,target --disable-ip
UCD-SNMP-MIB::versionClearCache.0 = INTEGER: 0
UCD-SNMP-MIB::versionUpdateConfig.0 = INTEGER: 0
UCD-SNMP-MIB::versionRestartAgent.0 = INTEGER: 0
UCD-SNMP-MIB::versionSavePersistentData.0 = INTEGER: 0
UCD-SNMP-MIB::versionDoDebugging.0 = INTEGER: 1
UCD-SNMP-MIB::snmperrIndex.0 = INTEGER: 0
UCD-SNMP-MIB::snmperrNames.0 = STRING: snmp
UCD-SNMP-MIB::snmperrErrorFlag.0 = INTEGER: noError(0)
UCD-SNMP-MIB::snmperrErrMessage.0 = STRING:
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.1.1 = INTEGER: 1
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.1.2 = INTEGER: 2
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.1.3 = INTEGER: 3
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.1.4 = INTEGER: 4
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.1.5 = INTEGER: 5
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.4.1 = Hex-STRING: 00 00 00 00 00 00
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.4.2 = Hex-STRING: 00 00 00 00 00 00
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.4.3 = Hex-STRING: 00 00 00 00 00 00
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.4.4 = Hex-STRING: 00 00 00 00 00 00
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.4.5 = STRING: "hT�Q��"
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.13.1 = INTEGER: -86
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.13.2 = INTEGER: -86
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.13.3 = INTEGER: -86
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.13.4 = INTEGER: -86
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.13.5 = INTEGER: -86
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.26.1 = INTEGER: 66
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.26.2 = INTEGER: 25
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.26.3 = INTEGER: 19
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.26.4 = INTEGER: 27
UCD-SNMP-MIB::ucdavis.255.3.54.1.3.32.1.26.5 = INTEGER: 49
```

daenney:
A general observation... there's an enormous amount of duplicate OID mangling in `snmp.yml` b/c a lot of devices use the same OIDs. Wouldn't it make more sense to split this out in one module per MIB, so have `snmpv2smi`, `hostresources`, ucdsnmp` and only have `ddwrt` walk the extra OIDs it cares about, just as to similarly limit `apcups` to just walking `1.3.6.1.4.1.318`?

Since `modules` appears to be an array it should be possible to have a target be checked by multiple modules?

daenney:
One more question, for the `laTable` ones I would actually prefer to associate the `laNames` with `laLoadInt` but adding a

daenney:
Ah, I've figured it out after looking at how it's done in the `default` module. But damn this breaks my brain.

daenney:
I'm running into a very odd problem. I think it's a bug in DD-WRT honestly but I'm wondering if there's some way to work around it.

It seems that there are `hrStorageIndex` entries that have the exact same `hrStorageDescr` entry:
```
# HELP hrStorageIndex 
# TYPE hrStorageIndex gauge
hrStorageIndex{hrStorageIndex="1"} 1
[..]
hrStorageIndex{hrStorageIndex="2"} 2
[..]

# HELP hrStorageDescr 
# TYPE hrStorageDescr gauge
hrStorageDescr{hrStorageDescr="/",hrStorageIndex="1"} 1
hrStorageDescr{hrStorageDescr="/",hrStorageIndex="2"} 1
[..]
```

This makes it impossible to do the `lookups` trick to then be able to associate `hrStorageDescr` as label with the rest of the `hrStorage*` entries since the exporter will rightfully complain that

```
collected metric hrStorageDescr label:<name:"hrStorageDescr" value:"/" > gauge:<value:1 >  was collected before with the same name and label values
```

Is it possible to work around that? I tried keeping two `indexes` on the entry but that just results in an empty label being added to it?

SuperQ:
These need to be added to the [generator](https://github.com/prometheus/snmp_exporter/blob/master/generator/generator.yml) configuration.

SuperQ:
You're right about not duplicating the `default` walk.  And only implement the specific walks for the non-default network metrics.  Breaking things up by system in more generic modules like you suggest is probably a good idea, since a lot of these are standardized MIBs, and not ddwrt specific.

Of course, it'd be way better to use the `node_exporter` with a ddwrt box, since you can skip SNMP altogether. 😄 

Another ddwrt option would be to use collectd, since it has the option to provide a prometheus target endpoint.

daenney:
I spent two days trying to get node_exporter to run on my router :P. Switched firmware a few times over, Merlin one's kernel was too old but it turns out DD-WRT compiles for MIPS without FPU emulation which makes Go break. It's been a long long day 😛. Hence why I went down this road.

SuperQ:
Yea, embedded systems can be a huge pain.

daenney:
Eh, so how do I add the `1.3.6.1.4.1.2021.255` to generator? The `.255` is unknown so is basically used to stash other things in but as a consequence there's no proper names for those OIDs that it can piggy-back on.

In that same spirit, do I need to do something specific once I've amended generator, like run it and include the result in the PR too?

brian-brazil:
> there's no proper names for those OIDs that it can piggy-back on since no MIB exists for it.

We need the MIB, as that's how SNMP is designed. If one doesn't already exist, your best option in this case is probably to contribute one to dd-wrt and make sure it ends up somewhere easy to download.

> In that same spirit, do I need to do something specific once I've amended generator, like run it and include the result in the PR too?

Yes. You'll also need to include a pointer to the MIB in the README.

daenney:
> We need the MIB, as that's how SNMP is designed. If one doesn't already exist, your best option in this case is probably to contribute one to dd-wrt and make sure it ends up somewhere easy to download.

Hmm, might be a while, if ever, for them to pull that in. But I can create one and host it on a GitHub repository for now. Would that be enough for the scope of this PR?

daenney:
Slightly different question. Lets say for `hrStorageCapacity` it currently looks like `hrDiskStorageCapacity{hrDeviceIndex="1552"} 3.0253056e+07` (basically 32GB). Is there any way to make it get both `hrDeviceIndex` and `hrDeviceDescr` as labels on that thing? Based on the lookup I could get `hrDeviceIndex` replaced by `hrDeviceDescr`, but what if I wanted both (for programatic reason).

SuperQ:
@daenney If you created a valid MIB, that's good enough for me.

For the labels, officially you should only include label lookups that are minimally unique.  So you would have to pick one of the two.

In practice, I do this all the time[0].

daenney:
Alright, in that case I'll amend the PR to include the defined entries from UCDAVIS and deal with the `.255` range separately as it might take me some time to whip up a MIB for that. It's been years... I'll maintain an artisanally handcrafted `snmp.yml` somewhere else with the added bits and extra labels for now.

Final question, I think. There are some entries from `laTable` that are really not worth scraping/including. Is there any way to indicate that? `laLoad` for example is useless in this context since it's an `DisplayString` of a load, like `0.01`. However since that value changes all the time you end up with a humongous amount of time series which all have `{laLoad=string} 1` as a result.

SuperQ:
There's no current option in the `generator.yml` to allow for an ignored OID list.  This would be a useful feature.

For now, you have to be a bit verbose in the walk list and only include the exact list you want included.

daenney:
I'm struggling a bit to get the `snmp-mibs-downloader` to work. I added `cisco.conf` with:
```
HOST=ftp://ftp.cisco.com
ARCHIVE=v2.tar.gz
ARCHTYPE=tgz
DIR=pub/mibs/v2/
ARCHDIR=auto/mibs/v2
CONF=ciscolist
DEST=cisco
```

And a `ciscolist` that basically mentions every MIB in the archive (though it probably shouldn't?):
```
ACCOUNTING-CONTROL-MIB.my
ACTONA-ACTASTOR-MIB.my
ADMIN-AUTH-STATS-MIB.my
ADSL-DMT-LINE-MIB.my
ADSL-LINE-MIB.my
[..]
T11-TC-MIB.my
TCP-MIB.my
TN3270E-MIB.my
TN3270E-RT-MIB.my
TOKENRING-MIB.my
TUNNEL-MIB.my
UDP-MIB.my
VDSL2-LINE-MIB.my
VDSL2-LINE-TC-MIB.my
VPN-TC-STD-MIB.my
VRRP-MIB.my
XGCP-MIB.my
draft-ietf-ipfc-fcmgmt-int-mib-04.my
test-mib.my
```

However, running a `download-mibs cisco` just results in a whole bunch of this:

```
NOTE: CISCO-ATM-ACCESS-LIST-MIB: ignored.
WARNING: Module(s) not found: -
NOTE: CISCO-ATM-ADDR-MIB: ignored.
WARNING: Module(s) not found: -
NOTE: CISCO-ATM-CAPABILITY: ignored.
```

daenney:
Oh, it's a dual column format, file name + MIB name. My bad.

daenney:
Alright. I think that's it. Thank you so much for your help and guidance @SuperQ, @brian-brazil!

brian-brazil:
>   But I can create one and host it on a GitHub repository for now. Would that be enough for the scope of this PR?

We would prefer the MIB from the vendor, otherwise we can't know that you got all the details right.

daenney:
> We would prefer the MIB from the vendor, otherwise we can't know that you got all the details right.

There is no MIB from the vendor. Someone wrote a simple script named `wl_snmp.sh` that's just hooking into that part of the tree. I can try and contribute one to them but it's largely based on goodwill from their side to even keep that up to date / in sync. For now I've left it out of this PR.

daenney:
I've removed `hrSystem`, it's largely useless on this device. The only interesting one was uptime that `sysUpTime` already provides.

generator/generator.yml
```diff
@@ -132,11 +132,34 @@ modules:
...
+#
+# Tested on  DD-WRT v3.0-r31825 (04/06/17) with an ASUS RT-AC66U
+
+  ddwrt:
+    walk:
+      - sysUpTime
+      - interfaces
+      - ifXTable
+      - 1.3.6.1.2.1.25.2 # hrStorage
+      - 1.3.6.1.4.1.2021.4 # ucdavis mem
```
brian-brazil: this object is called `memory`

generator/generator.yml
```diff
@@ -132,11 +132,34 @@ modules:
...
+      # ucdavis laTable (load avg)
+      - 1.3.6.1.4.1.2021.10.1.1 # laIndex
```
brian-brazil: This is called processIndex in the MIB I'm looking at. Are you sure you're using the right MIB?

daenney: screen shot 2017-04-24 at 14 15 53

Are you sure you're looking at the right MIB: http://www.oidview.com/mibs/2021/UCD-SNMP-MIB.html?

generator/generator.yml
daenney: Fixed

generator/generator.yml
brian-brazil: Ah indeed, I was looking at the wrong object.

generator/generator.yml
```diff
@@ -132,11 +132,34 @@ modules:
...
+      - 1.3.6.1.4.1.2021.10.1.2 # laNames
+      - 1.3.6.1.4.1.2021.10.1.5 # laLoadInt
+      - 1.3.6.1.4.1.2021.11 # ucdavis systemStats
```
brian-brazil: just systemStats

daenney: Why? The other one where I'm requesting everything is prefixed with `ucdavis`, like `ucdavis memory`. The `laLoad` one has a header of `ucdavis laLoad` and then the three OIDs that are of interest explicitly listed. Consistency wise, this is what's done before. I'm happy to change it but then I'd like to do this once so if you can tell me what you'd like the whole block to look like I'm happy to amend it.

brian-brazil: The convention in this file is for the comment to be just the object name, without any reference to the module name.

daenney: Alright. Done. I've updated all the comments to reflect that then.

brian-brazil:
Thanks!

daenney:
Thank you too!

PR #154, opened 2017-04-25
Add missing CPU, Load and Storage OIDs

jayme-github:
Sorry @brian-brazil - I missed some. :slightly_frowning_face:

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -153,6 +153,13 @@ modules:
       - interfaces
       - sysUpTime
       - ifXTable
+      - laNames
+      - laLoadInt
+      - ssCpuUser
+      - ssCpuSystem
+      - ssCpuIdle
+      - memory
+      - hrStorage
       - 1.3.6.1.4.1.6574.1       # synoSystem
       - 1.3.6.1.4.1.6574.2       # synoDisk
       - 1.3.6.1.4.1.6574.3       # synoRaid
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks!

Svedrin:
this is a bit odd... I tried to add the very same metrics to SNMP exporter a while back so I could scrape my linux boxes more easily which have snmpd deployed already, without needing to deploy node_exporter everywhere. So now I just need to tell snmp_exporter that everything's a Synology NAS and it'll happily scrape the values for me? that doesn't sound right :/ maybe the CPU and storage parts could be split out into a more generic module?

brian-brazil:
If you can use the node exporter you should. SNMP is inferior in many many ways.

juliusv:
The context on IRC from @Svedrin was:

> when I proposed to add disk/cpu metrics to snmp-exporter that proposal was rejected due to "we want you to use node-exporter for that" reasons, but now the very same metrics have been added to the synology module... so I just have to tell snmp-exporter that my linux box is a synology NAS and it'll happily scrape the values for me?

So I guess the question was why allow CPU/disk metrics for Synology, but not for other things. Is it not possible to run node exporter on Synology? Just wondering about the consistency of the decision.

SuperQ:
I originally suggested that using the node_exporter on the Synology was the best path.  Using SNMP for this is not what I would recommend.

PR #162, opened 2017-05-17
Update generator

SuperQ:
Use lookup node's index label in the generator:
* Instead of using the literal name from the generator's config, use the
  Label from the MIB index.  This allows OIDs to be used in the
  `new_index` field in the generator.yml.

Use an explicit OID for ifName lookups.
* This avoids the ifName conflict caused by the Citrix netscaler NS-ROOT-MIB module.

Update snmp.yml  
* This updates with the latest generator code, removes non-accessible
  metrics from being used.

Misc:
* Ignore the generator binary from the git repo.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -10,7 +10,8 @@ modules:
     walk: [sysUpTime, interfaces, ifXTable]
     lookups:
       - old_index: ifIndex
-        new_index: ifName
+        # Use OID to avoid conflict with Netscaler NS-ROOT-MIB.
+        new_index: 1.3.6.1.2.1.31.1.1.1.1 # ifName
 # If only ifIndex is unique.
   default_ifindex:
     walk: [sysUpTime, interfaces, ifXTable]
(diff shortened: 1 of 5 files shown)


generator/generator.yml
```diff
@@ -10,7 +10,7 @@ modules:
     walk: [sysUpTime, interfaces, ifXTable]
     lookups:
       - old_index: ifIndex
-        new_index: ifName
+        new_index: 1.3.6.1.2.1.31.1.1.1.1 # ifName
```
brian-brazil: This should have a comment as to why we're doing this

generator/tree.go
```diff
@@ -195,14 +195,14 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 					}
 					indexNode := nameToNode[lookup.NewIndex]
 					// Avoid leaving the old labelname around.
-					index.Labelname = lookup.NewIndex
+					index.Labelname = indexNode.Label
```
brian-brazil: This change should have a unittest

SuperQ: Added a test for OID-based lookups.

PR #179, opened 2017-06-23
Update by request

vitaly-m:
Hello @brian-brazil,

I think this change may be useful for complex configuration and for big amount of targets per one exporter.

The on demand update will be enabled only in case -config.update-on-demand specified, if not specified it works as before.

### DIFF
--- a/main.go
+++ b/main.go
@@ -109,13 +109,14 @@ func updateConfiguration(w http.ResponseWriter, r *http.Request) {
 }
 
 func (sc *SafeConfig) reloadConfig(configFile string) (err error) {
-	sc.Lock()
-	sc.C, err = config.LoadFile(configFile)
-	sc.Unlock()
+	conf, err := config.LoadFile(configFile)
 	if err != nil {
 		log.Errorf("Error parsing config file: %s", err)
 		return err
 	}
+	sc.Lock()
+	sc.C = conf
+	sc.Unlock()
 	log.Infoln("Loaded config file")
 	return nil
 }
--- a/main.go
+++ b/main.go
@@ -41,7 +41,7 @@ var (
 			Help: "Errors in requests to the SNMP exporter",
 		},
 	)
-	updateOnDemand = flag.Bool("config.update-on-demand", false, "Configuration will be updated only on demand in case true.")
+	updateOnDemand = flag.Bool("config.update-on-demand", false, "Configuration will be updated only on demand if specified.")
 	cfg            *config.Config
 	err            error
 )
--- a/main.go
+++ b/main.go
@@ -99,8 +99,7 @@ func updateConfiguration(w http.ResponseWriter, r *http.Request) {
 	case "POST":
 		rc := make(chan error)
 		reloadCh <- rc
-		var err error
-		if err = <-rc; err != nil {
+		if err := <-rc; err != nil {
 			http.Error(w, fmt.Sprintf("failed to reload config: %s", err), http.StatusInternalServerError)
 		}
 	default:
--- a/main.go
+++ b/main.go
@@ -52,7 +52,6 @@ var (
 	sc = &SafeConfig{
 		C: &config.Config{},
 	}
-	err      error
 	reloadCh chan chan error
 )
 
@@ -100,6 +99,7 @@ func updateConfiguration(w http.ResponseWriter, r *http.Request) {
 	case "POST":
 		rc := make(chan error)
 		reloadCh <- rc
+		var err error
 		if err = <-rc; err != nil {
 			http.Error(w, fmt.Sprintf("failed to reload config: %s", err), http.StatusInternalServerError)
 		}
@@ -132,6 +132,7 @@ func main() {
 	log.Infoln("Build context", version.BuildContext())
 
 	// Bail early if the config is bad.
+	var err error
 	sc.C, err = config.LoadFile(*configFile)
 	if err != nil {
 		log.Fatalf("Error parsing config file: %s", err)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -41,6 +41,9 @@ var (
 			Help: "Errors in requests to the SNMP exporter",
 		},
 	)
+	updateOnDemand = flag.Bool("config.update-on-demand", false, "Configuration will be updated only on demand if specified.")
```
brian-brazil: No need for this option, the current behaviour should not be kept around.

main.go
```diff
@@ -88,6 +93,22 @@ func handler(w http.ResponseWriter, r *http.Request) {
 	log.Debugf("Scrape of target '%s' with module '%s' took %f seconds", target, moduleName, duration)
 }
 
+func updateConfiguration(w http.ResponseWriter, r *http.Request) {
```
brian-brazil: Look at how it's done in https://github.com/prometheus/blackbox_exporter/blob/master/main.go

main.go
```diff
@@ -100,17 +121,18 @@ func main() {
...
-	for module, _ := range *c {
+	for module, _ := range *cfg {
 		snmpDuration.WithLabelValues(module)
 	}
 
-	http.Handle("/metrics", promhttp.Handler()) // Normal metrics endpoint for SNMP exporter itself.
-	http.HandleFunc("/snmp", handler)           // Endpoint to do SNMP scrapes.
+	http.Handle("/metrics", promhttp.Handler())     // Normal metrics endpoint for SNMP exporter itself.
+	http.HandleFunc("/snmp", handler)               // Endpoint to do SNMP scrapes.
+	http.HandleFunc("/reload", updateConfiguration) // Endpoint to do SNMP scrapes.
```
SuperQ: Looks like comment copy-n-pasta.

SuperQ:
Nice, thanks for starting this work.

vitaly-m:
OK, thank you for your comments will fix.

vitaly-m:
Unfortunately I didn't understand the logic with SIGHUP.

brian-brazil:
Just copy what the blackbox exporter is doing with HUP.

vitaly-m:
Am I understand correctly that I can send a special signal to the process and it will reload the configuration? if yes, how can I check that?

SuperQ:
You mean like `kill -HUP $(pidof snmp_exporter)`?

vitaly-m:
Tested, should work.

RichiH:
Now that this is consistent during runtime, it would make sense to expose the config via `/config` like Prometheus does. Community strings and the like need to be stripped, of course.

SuperQ:
@RichiH Sounds like a feature request issue is in order :smiley_cat:

vitaly-m:
This feature with `/config` is required for our cases too. But I think we should not mix config update and config read in this pull request.

vitaly-m:
Should I change something else?

main.go
```diff
@@ -41,6 +49,11 @@ var (
 			Help: "Errors in requests to the SNMP exporter",
 		},
 	)
+	sc = &SafeConfig{
+		C: &config.Config{},
+	}
+	err      error
```
SuperQ: This should not be declared in the global scope.

vitaly-m: Why not? Configuration is global for the application and it should be available in different functions in main.go.

SuperQ: Not the config, the `err`.  The error var needs to be used to the functions that use it, otherwise there is a global race condition.  Consider `var err error` or `err := foo` to replace this.

main.go
```diff
@@ -132,6 +132,7 @@ func main() {
 	log.Infoln("Build context", version.BuildContext())
 
 	// Bail early if the config is bad.
+	var err error
```
SuperQ: This is one of those cases where you can use it with `:=`

```
sc.C, err := config.LoadFile(*configFile)
```

main.go
```diff
@@ -100,6 +99,7 @@ func updateConfiguration(w http.ResponseWriter, r *http.Request) {
 	case "POST":
 		rc := make(chan error)
 		reloadCh <- rc
+		var err error
```
SuperQ: Same here.
```
if err := <-rc; err != nil {
```

main.go
vitaly-m: It confuse me, because I thought that if I use := operator, then new local variable will be declared and in that case sc.C will not be filled, or compilation error will occur.

SuperQ: In the case of multiple returns, `:=` will only create if necessary.  I've seen the compile error if the declare is used in a `switch` or `if` and there are uses outside of the context.  For those you need to declare with a `var` outside the switch/if blocks.

vitaly-m: The previous error from from fmt, this one is from compiler:
```
./main.go:134: non-name sc.C on left side of :=
```

SuperQ: Ahh, you're right.  I guess this is fine as-is.

SuperQ:
@brian-brazil Care to take one last look at this?

main.go
```diff
@@ -88,9 +94,34 @@ func handler(w http.ResponseWriter, r *http.Request) {
...
+		}
+	default:
+		log.Errorf("POST method expected")
+		http.Error(w, "POST method expected", 400)
+	}
+}
+
+func (sc *SafeConfig) reloadConfig(configFile string) (err error) {
+	sc.Lock()
+	sc.C, err = config.LoadFile(configFile)
```
brian-brazil: If there's an error in the config this sill updates the config. We should only update `sc.C` after checking `err`.

brian-brazil:
Thanks!

Issue #187, opened 2017-07-03
In pduValueAsString int64 should be uint64


brian-brazil:
The SNMP library can't produce a int64, but it can produce a uint64 (for Counter64s). This should never come up as counters do not make good labels, but it should be fixed.

PR #188, opened 2017-07-03
Handle SnmpPDUs with a value of type uint64

kongslund:
We encountered the following log message when trying to use Prometheus SNMP Exporter against a Cisco firewall:

    INFO[0014] Got PDU with unexpected type: Name: .1.3.6.1.4.1.9.9.392.1.3.21.1.35.8.107.114.105.115.116.105.97.110.9940993 Value: '%!s(uint64=522077)', Go Type: uint64 SNMP Type: %!s(gosnmp.Asn1BER=70)   source="collector.go:257"

This pull request fixes that by adding support for uint64.

Ping: @brian-brazil

Closes: https://github.com/prometheus/snmp_exporter/issues/187

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -226,8 +226,6 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ string) string {
 		return strconv.Itoa(pdu.Value.(int))
 	case uint:
 		return strconv.FormatUint(uint64(pdu.Value.(uint)), 10)
-	case int64:
-		return strconv.FormatInt(pdu.Value.(int64), 10)
 	case uint64:
 		return strconv.FormatUint(pdu.Value.(uint64), 10)
 	case string:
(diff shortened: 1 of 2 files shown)


collector.go
```diff
@@ -228,6 +228,8 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ string) string {
 		return strconv.FormatUint(uint64(pdu.Value.(uint)), 10)
 	case int64:
```
brian-brazil: Can you remove the int64 please? This isn't possible.

SuperQ: Are you sure we want to do that?  What if we change the upstream snmp and this becomes a valid case?

kongslund: For typ=="counter"? Or in all cases?

brian-brazil: I don't think SNMPv4 is on the horizon with support for int64 :)

brian-brazil: In all cases. The only thing a uint64 can be is a counter.

SuperQ: Ok, so it's an SNMP spec issue, not a library issue.  I misunderstood your IRC comment as implying it was an issue with the library. 👍

brian-brazil:
Thanks!

PR #193, opened 2017-07-13
Validate metric names

### LINKED ISSUE #184: Validate metric names and make them valid when possible
Some MIB files produce metric names that have dashes in them (like digital-sen1-1).

Should the snmp_exporter convert dashes to underscores to make these into valid metric names?

In this case, the problem metric is coming off of an older AVTECH TemPageR device. I'm currently building snmp_exporter from source (aa8c078) on a Linux machine.

Conorbro:
Fixes #184

### DIFF
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"regexp"
 	"sort"
 	"strings"
 
@@ -161,7 +162,7 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 			}
 
 			metric := &config.Metric{
-				Name:    n.Label,
+				Name:    sanitizeLabelName(n.Label),
 				Oid:     n.Oid,
 				Type:    t,
 				Indexes: []*config.Index{},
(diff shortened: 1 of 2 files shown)


generator/tree.go
```diff
@@ -160,6 +160,10 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 				return // Inaccessible metrics.
 			}
 
+			if strings.Contains(n.Label, "-") {
```
brian-brazil: You only need this change, plus a unittest for it.

Conorbro:
@brian-brazil

generator/tree_test.go
```diff
@@ -613,3 +613,44 @@ func TestGenerateConfigModule(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateMetricNames(t *testing.T) {
```
brian-brazil: This can go as part of TestGenerateConfigModule

Conorbro:
@brian-brazil

generator/tree.go
```diff
@@ -160,6 +160,10 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
...
+				n.Label = strings.Replace(n.Label, "-", "_", -1)
+			}
+
 			metric := &config.Metric{
 				Name:    n.Label,
```
brian-brazil: Replace it only down here, changing the actual n.Label could cause issues elsewhere.

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

PR #197, opened 2017-07-17
Rename default module

SuperQ:
To avoid confusion that `default` is special, rename it to `if_mib`
based on the primary MIB.
* Use `ifIndex` for default interface indexing instead of `ifDescr`.
* Make `ifDescr` its own module.
* Add `ifAlias` module.
* Update docs.
* Update `snmp.yml`.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -1,20 +1,26 @@
 modules:
-# If ifDesc is unique.
-  default:
+  # Default IF-MIB interfaces table with ifIndex.
+  if_mib:
+    walk: [sysUpTime, interfaces, ifXTable]
+  # Interfaces if ifAlias is unique.
+  if_mib_ifalias:
+    walk: [sysUpTime, interfaces, ifXTable]
+    lookups:
+      - old_index: ifIndex
+        new_index: ifAlias
+  # Interfaces if ifDescr is unique.
+  if_mib_ifdescr:
     walk: [sysUpTime, interfaces, ifXTable]
     lookups:
       - old_index: ifIndex
         new_index: ifDescr
-# If ifName is unique.
-  default_ifname:
+  # Interfaces if ifName is unique.
+  if_mib_ifname:
     walk: [sysUpTime, interfaces, ifXTable]
     lookups:
       - old_index: ifIndex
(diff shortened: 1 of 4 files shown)


brian-brazil:
I'm fine with the general thurst of this PR as we've previously discussed.

I'm not so sure about changing the name from `default` to `interfaces`, there's a lot of docs/references to that already.

We'll also need to do something about existing dashboards which presume the current default, at least the example console template (which we should probably just delete).

SuperQ:
I wanted to change the name, as I've had many cases where people confuse `default` with being a global defaults.

generator/README.md
```diff
@@ -41,14 +41,29 @@ Put the extracted mibs in a location NetSNMP can read them from. `$HOME/.snmp/mi
 
 https://github.com/librenms/librenms/tree/master/mibs can also be a good source of MIBs.
 
-
+## Generator file format
```
brian-brazil: This needs work generally, I'd leave it to another PR. For example all the auth stuff needs to move over here.

SuperQ:
@RichiH @brian-brazil PTAL

generator/generator.yml
```diff
@@ -1,20 +1,20 @@
 modules:
-# If ifDesc is unique.
-  default:
+  # Default IF-MIB interfaces table with ifIndex.
+  if-mib:
```
brian-brazil: as discussed, let's keep with underscores for consistency.

generator/generator.yml
```diff
@@ -1,20 +1,20 @@
...
+  # Interfaces if ifName is unique.
+  if-mib_ifname:
     walk: [sysUpTime, interfaces, ifXTable]
     lookups:
       - old_index: ifIndex
         # Use OID to avoid conflict with Netscaler NS-ROOT-MIB.
         new_index: 1.3.6.1.2.1.31.1.1.1.1 # ifName
-# If only ifIndex is unique.
-  default_ifindex:
+  # Interfaces if ifDescr is unique.
```
brian-brazil: We should add ifAlias too for completelness

SuperQ:
@brian-brazil Finished cleanup, added `snmp.yml` update.

PR #199, opened 2017-07-19
Expose loaded config on /config and elide secrets

Conorbro:


config/config.go
```diff
@@ -197,14 +217,31 @@ func (c *Lookup) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+}
+
+//UnmarshalYAML implements the yaml.Unmarshaler interface for Secrets.
+func (s *Secret) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type plain Secret
+	return unmarshal((*plain)(s))
+}
+
 type Auth struct {
 	Community     string `yaml:"community,omitempty"`
```
brian-brazil: This is a secret

config/config.go
```diff
@@ -55,6 +62,19 @@ type Module struct {
 	XXX map[string]interface{} `yaml:",inline"`
 }
 
+func (sc *SafeConfig) ReloadConfig(configFile string) (err error) {
```
brian-brazil: I think this should remain in the exporter code, as we'll be mixing in the generator config into this package at some point.

Conorbro:
@brian-brazil

PR #201, opened 2017-07-24
Keep non-accessible objects

### LINKED ISSUE #192: Reconsider dropping not-accessible objects
We recently stopped generating metrics for objects with an access level of not-accessible. This commonly includes (in accordance with the RFCs) numeric index objects in tables. While this is fine in theory, in reality there's other tables that might reference these by number (in the description, nothing we can pull out automatically), so we should probably generate these again.

We should continue to drop trap objects.

Conorbro:
Closes #192


brian-brazil:
Thanks!

PR #203, opened 2017-07-24
Rework docs

brian-brazil:
This moves the primary config docs to be those for the generator, as that's what users will be using in practice.

I also explicitly mentioned GETBULK to help reduce misunderstandings there.


Conorbro:
Looks good to me.

PR #204, opened 2017-07-25
Added test for pduToSample and corrected int to uint for metric value

### LINKED ISSUE #191: Negative value for ifHCOutUcastPkts
Hi,

I have the snmp_exporter running against a Juniper MX (Junos 17.1) with the default snmp.yaml and I'm getting a negative value for ifHCOutUcastPkts:

# HELP ifHCOutUcastPkts 
# TYPE ifHCOutUcastPkts counter
ifHCOutUcastPkts{ifDescr="ae0"} -1.241915e+06

I verified on the router that the value is indeed not negative:

teichhorn@DE-ES-001-03-10-01-re0> show snmp mib get 1.3.6.1.2.1.31.1.1.1.11.629
ifHCOutUcastPkts.629 = 18446744073708309388

I am not sure if this is a problem handling 64bit values, 
or howto debug this further - running the exporter with log level debug doesn't give any hints to what is happening behind the curtains.

If you need any help from my side, please don't hesitate to ask.

Thanks,
Tom

Conorbro:
Fixes #191

### DIFF
--- a/collector_test.go
+++ b/collector_test.go
@@ -6,8 +6,97 @@ import (
 
 	"github.com/soniah/gosnmp"
 
+	"github.com/prometheus/client_model/go"
 	"github.com/prometheus/snmp_exporter/config"
 )
+
+func TestPduToSample(t *testing.T) {
+	cases := []struct {
+		pdu                *gosnmp.SnmpPDU
+		indexOids          []int
+		metric             *config.Metric
+		oidToPdu           map[string]gosnmp.SnmpPDU
+		expectedMetric     string
+		expectedMetricDesc string
+	}{
+		{
+			pdu: &gosnmp.SnmpPDU{
+				Name:  "1.1.1.1.1",
+				Type:  gosnmp.Integer,
+				Value: 2,
+			},
+			indexOids: []int{},
+			metric: &config.Metric{
+				Name: "test_metric",
+				Oid:  "1.1.1.1.1",
+				Type: "counter",
+				Help: "Help string",
+			},
+			oidToPdu:           make(map[string]gosnmp.SnmpPDU),
+			expectedMetric:     "counter:<value:2 > ",
+			expectedMetricDesc: `Desc{fqName: "test_metric", help: "Help string", constLabels: {}, variableLabels: []}`,
+		},
+		{
+			pdu: &gosnmp.SnmpPDU{
+				Name:  "1.1.1.1.1",
+				Type:  gosnmp.Integer,
+				Value: 2,
+			},
+			indexOids: []int{},
+			metric: &config.Metric{
+				Name: "test_metric",
+				Oid:  "1.1.1.1.1",
+				Type: "gauge",
+				Help: "Help string",
+			},
+			oidToPdu:           make(map[string]gosnmp.SnmpPDU),
+			expectedMetric:     "gauge:<value:2 > ",
+			expectedMetricDesc: `Desc{fqName: "test_metric", help: "Help string", constLabels: {}, variableLabels: []}`,
+		},
+		{
+			pdu: &gosnmp.SnmpPDU{
+				Name:  "1.1.1.1.1",
+				Type:  gosnmp.Integer,
+				Value: -2,
+			},
+			indexOids: []int{},
+			metric: &config.Metric{
+				Name: "test_metric",
+				Oid:  "1.1.1.1.1",
+				Help: "Help string",
+			},
+			oidToPdu:           make(map[string]gosnmp.SnmpPDU),
+			expectedMetric:     `label:<name:"test_metric" value:"-2" > gauge:<value:1 > `,
+			expectedMetricDesc: `Desc{fqName: "test_metric", help: "Help string", constLabels: {}, variableLabels: [test_metric]}`,
+		},
+	}
+
+	for _, c := range cases {
+		m := pduToSample(c.indexOids, c.pdu, c.metric, c.oidToPdu)
+		metric := &io_prometheus_client.Metric{}
+		err := m.Write(metric)
+		if err != nil {
+			t.Fatalf("Error writing metric: %v", err)
+		}
+		if metric.String() != c.expectedMetric {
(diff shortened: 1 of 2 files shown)


collector_test.go
```diff
@@ -9,6 +9,72 @@ import (
...
+				Help: "Help string",
+			},
+			oidToPdu:       make(map[string]gosnmp.SnmpPDU),
+			expectedResult: "Desc{fqName: \"test_metric\", help: \"Help string\", constLabels: {}, variableLabels: [test_metric]}",
+		},
+	}
+
+	for _, c := range cases {
+		res := pduToSample(c.indexOids, c.pdu, c.metric, c.oidToPdu)
+		if res.Desc().String() != c.expectedResult {
```
brian-brazil: You'd probably be better off using `Write` rather than `Desc` so the values can be checked.

collector_test.go
```diff
@@ -9,6 +9,72 @@ import (
...
+		oidToPdu       map[string]gosnmp.SnmpPDU
+		expectedResult string
+	}{
+		{
+			pdu: &gosnmp.SnmpPDU{
+				Name:  "1.3.6.1.2.1.31.1.1.1.11.629",
+				Type:  0x02,
+				Value: 2,
+			},
+			indexOids: []int{1},
```
brian-brazil: This should be empty, as there's no index in the config.Metric

collector_test.go
```diff
@@ -9,6 +9,72 @@ import (
...
+	cases := []struct {
+		pdu            *gosnmp.SnmpPDU
+		indexOids      []int
+		metric         *config.Metric
+		oidToPdu       map[string]gosnmp.SnmpPDU
+		expectedResult string
+	}{
+		{
+			pdu: &gosnmp.SnmpPDU{
+				Name:  "1.3.6.1.2.1.31.1.1.1.11.629",
```
brian-brazil: Use dummy oids for testing, otherwise you have to make sure the data makes sense for this oid to have

Conorbro:
@brian-brazil

collector_test.go
```diff
@@ -6,9 +6,82 @@ import (
...
+				Oid:  "1.1.1.1.1",
+				Help: "Help string",
+			},
+			oidToPdu: make(map[string]gosnmp.SnmpPDU),
+		},
+	}
+
+	for _, c := range cases {
+		m := pduToSample(c.indexOids, c.pdu, c.metric, c.oidToPdu)
+		err := m.Write(&io_prometheus_client.Metric{})
```
brian-brazil: You should check what's put in the metric.

Conorbro:
@brian-brazil

collector_test.go
```diff
@@ -6,9 +6,90 @@ import (
...
+				Value: -2,
+			},
+			indexOids: []int{},
+			metric: &config.Metric{
+				Name: "test_metric",
+				Oid:  "1.1.1.1.1",
+				Help: "Help string",
+			},
+			oidToPdu:       make(map[string]gosnmp.SnmpPDU),
+			expectedMetric: "label:<name:\"test_metric\" value:\"-2\" > gauge:<value:1 > ",
```
brian-brazil: You can use backticks to have a raw value, so you don't need the \ for escaping

collector_test.go
```diff
@@ -6,9 +6,90 @@ import (
...
+		}
+		if metric.String() != c.expectedMetric {
+			t.Fatalf("Unexpected metric: got %v, want %v", metric.String(), c.expectedMetric)
+		}
+	}
+}
+
+func TestGetPduValue(t *testing.T) {
+	pdu := &gosnmp.SnmpPDU{
+		Value: -2,
```
brian-brazil: Signed integers are supported by SNMP, it's the Counter64 which are the uint64s that we need to be careful with.

collector_test.go
```diff
@@ -6,9 +6,90 @@ import (
...
+				Type: "counter",
+				Help: "Help string",
+			},
+			oidToPdu:       make(map[string]gosnmp.SnmpPDU),
+			expectedMetric: "counter:<value:2 > ",
+		},
+		{
+			pdu: &gosnmp.SnmpPDU{
+				Name:  "1.1.1.1.1",
+				Type:  0x02,
```
brian-brazil: Use the names here for clarity: https://github.com/soniah/gosnmp/blob/adbb107aeaa6c23e71056bfd7a095874a07c7d04/gosnmp.go#L134

Conorbro:
@brian-brazil

collector_test.go
brian-brazil: This test doesn't make sense, -2 isn't a valid value for a Counter64. Try 2^63.

collector_test.go
```diff
@@ -6,9 +6,91 @@ import (
...
+			},
+			indexOids: []int{},
+			metric: &config.Metric{
+				Name: "test_metric",
+				Oid:  "1.1.1.1.1",
+				Type: "counter",
+				Help: "Help string",
+			},
+			oidToPdu:       make(map[string]gosnmp.SnmpPDU),
+			expectedMetric: "counter:<value:2 > ",
```
brian-brazil: Should the help string not get propagated?

Conorbro: The `io_prometheus_client.Metric` type doesn't actually have a field for help strings.

brian-brazil: Hmm, we'll need to check the output of Desc as well then.

collector_test.go
```diff
@@ -1,14 +1,97 @@
...
+		}
+		if metric.String() != c.expectedMetric {
+			t.Fatalf("Unexpected metric: got %v, want %v", metric.String(), c.expectedMetric)
+		}
+	}
+}
+
+func TestGetPduValue(t *testing.T) {
+	pdu := &gosnmp.SnmpPDU{
+		Value: uint64(math.Pow(2, 63)),
```
brian-brazil: `uint64(1 << 63)` would be better, as it avoids floating point.

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

PR #205, opened 2017-07-25
Factored out WalkParams

Conorbro:
Factored out WalkParams into separate struct to be inlined in config.Module and ModuleConfig

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -24,28 +24,28 @@ func LoadFile(filename string) (*Config, error) {
 }
 
 var (
-	DefaultModule = Module{
-		Version:        2,
-		MaxRepetitions: 25,
-		Retries:        3,
-		Timeout:        time.Second * 20,
-	}
 	DefaultAuth = Auth{
 		Community:     "public",
 		SecurityLevel: "noAuthNoPriv",
 		AuthProtocol:  "MD5",
 		PrivProtocol:  "DES",
 	}
+	DefaultWalkParams = WalkParams{
+		Version:        2,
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -64,31 +70,31 @@ func (c *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := CheckOverflow(c.XXX, "module"); err != nil {
 		return err
 	}
-	if c.Auth == nil {
-		c.Auth = &DefaultAuth
+	if c.WalkParams.Auth == nil {
```
brian-brazil: All these checks should be done in the Unmarshal of WalkParams

config/config.go
```diff
@@ -64,31 +70,31 @@ func (c *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		c.WalkParams.Auth = &DefaultAuth
```
brian-brazil: Use the default logic we're using elsewhere.

config/config.go
```diff
@@ -97,19 +103,19 @@ func (c *Module) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // configureSNMP sets the various version and auth settings.
 func (c Module) ConfigureSNMP(g *gosnmp.GoSNMP) {
```
brian-brazil: This can change to be a method of WalkParams

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

PR #207, opened 2017-07-27
Switch to kingpin for flags

carlpett:
Continuing the conversions from `flag` to `kingpin`.


brian-brazil:
You missed the Dockerfile.

The generator also has subcommands, which should be switched over.

carlpett:
I can see the Dockerfile in the diff? Are there more of them?
Didn't see the generator since it didn't use the `flag` package. Will fix it now.

brian-brazil:
That's the only one I noticed.

carlpett:
I've updated to the new common/log as with the others, as well as converted the generator to use kingpin commands.

brian-brazil:
Thanks!

PR #212, opened 2017-08-02
Extract numeric time series from string values

### LINKED ISSUE #150: Allow overriding of MIB information
Some MIBs aren't quite right, particularly around types. Figure out what's a sane config syntax to allow for per-module overrides of bits of MIBs.

Conorbro:
Closes #150


generator/tree.go
```diff
@@ -215,6 +216,15 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 		}
 	}
 
+	// Apply module config overrides to their corresponding metrics.
+	for name, params := range cfg.Overrides {
+		for _, metric := range out.Metrics {
+			if name == metric.Name {
```
brian-brazil: Overrides may be provided by oid or name

config/config.go
```diff
@@ -256,6 +253,60 @@ func (c *Auth) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+type MetricOverrides struct {
+	StringMetrics map[string][]StringMetric `yaml:"string_metrics,omitempty"`
```
brian-brazil: At the collector level there should only be this setting, not the whole MetricOverrides as the rest of those will only exist in the generator.

brian-brazil: I don't think string_metrics quite captures it, we have other string metrics. regexp_extract or similar would be better.

collector.go
```diff
@@ -154,7 +154,10 @@ PduLoop:
 			}
 			if head.metric != nil {
 				// Found a match.
-				ch <- pduToSample(oidList[i+1:], &pdu, head.metric, oidToPdu)
+				samples := pduToSample(oidList[i+1:], &pdu, head.metric, oidToPdu)
```
brian-brazil: it's really pduToSamples now

collector.go
```diff
@@ -208,8 +215,27 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+	return res
+}
+
+func generateOverrideMetrics(metric *config.Metric, pduValue string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.Overrides.StringMetrics {
+		if len(strMetricSlice) > 0 {
+			for _, strMetric := range strMetricSlice {
+				value := strMetric.Regex.ReplaceAllString(pduValue, strMetric.Value)
+				v, _ := strconv.ParseFloat(value, 64)
```
brian-brazil: You need to check for an error here

collector.go
```diff
@@ -208,8 +215,27 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+				newMetric := prometheus.MustNewConstMetric(prometheus.NewDesc(metric.Name+"_"+name, metric.Help, []string{""}, nil),
```
brian-brazil: Don't add an underscore here

collector.go
```diff
@@ -192,6 +196,9 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
 		t = prometheus.GaugeValue
 		value = 1.0
 		stringType = true
+		if len(metric.Overrides.StringMetrics) > 0 {
+			res = generateOverrideMetrics(metric, pdu.Value.(string))
```
brian-brazil: You still need all the label logic below to apply. I'd reorder the code so this flows more naturally.

collector.go
brian-brazil: Why do you have an empty labelname?

collector.go
```diff
@@ -208,8 +215,27 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+	}
+	return res
+}
+
+func generateOverrideMetrics(metric *config.Metric, pduValue string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.Overrides.StringMetrics {
+		if len(strMetricSlice) > 0 {
+			for _, strMetric := range strMetricSlice {
+				value := strMetric.Regex.ReplaceAllString(pduValue, strMetric.Value)
```
brian-brazil: We need to check that the regex matches before doing subsitution.

collector.go
```diff
@@ -208,8 +215,27 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+		res = append(res, prometheus.MustNewConstMetric(prometheus.NewDesc(metric.Name, metric.Help, labelnames, nil),
+			t, value, labelvalues...))
+	}
+	return res
+}
+
+func generateOverrideMetrics(metric *config.Metric, pduValue string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.Overrides.StringMetrics {
+		if len(strMetricSlice) > 0 {
```
brian-brazil: No need for this, the for already handles this correctly

Conorbro:
@brian-brazil

collector.go
```diff
@@ -191,25 +201,41 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
-	for k, v := range labels {
-		labelnames = append(labelnames, k)
-		labelvalues = append(labelvalues, v)
-	}
-	// For strings we put the value as a label with the same name as the metric.
-	// If the name is already an index, we do not need to set it again.
-	if stringType {
+		// For strings we put the value as a label with the same name as the metric.
+		// If the name is already an index, we do not need to set it again.
 		if _, ok := labels[metric.Name]; !ok {
```
brian-brazil: This code doesn't apply if we're applying the regexes

collector.go
```diff
@@ -191,25 +201,41 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
 			labelnames = append(labelnames, metric.Name)
 			labelvalues = append(labelvalues, pduValueAsString(pdu, metric.Type))
 		}
+		if len(metric.Overrides) > 0 {
+			res = generateOverrideMetrics(metric, pdu.Value.(string), labelnames, labelvalues)
```
brian-brazil: I'd return here

config/config.go
```diff
@@ -156,24 +157,20 @@ func (c WalkParams) ConfigureSNMP(g *gosnmp.GoSNMP) {
...
-	Lookups []*Lookup `yaml:"lookups,omitempty"`
-
-	XXX map[string]interface{} `yaml:",inline"`
+	Name      string                     `yaml:"name"`
+	Oid       string                     `yaml:"oid"`
+	Type      string                     `yaml:"type"`
+	Help      string                     `yaml:"help"`
+	Indexes   []*Index                   `yaml:"indexes,omitempty"`
+	Lookups   []*Lookup                  `yaml:"lookups,omitempty"`
+	Overrides map[string][]RegexpExtract `yaml:"regexp_extract,omitempty"`
```
brian-brazil: The field should be called RegexpExtracts, not overrides

config/config.go
```diff
@@ -256,6 +253,43 @@ func (c *Auth) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	return nil
+}
+
+// Regexp encapsulates a regexp.Regexp and makes it YAML marshalable.
+type Regexp struct {
+	*regexp.Regexp
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
brian-brazil: Make sure to add an unmarshalling test for a config that's missing the regex, we've had issues there in the past

config/config.go
```diff
@@ -256,6 +253,43 @@ func (c *Auth) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+type RegexpExtract struct {
+	Regex Regexp `yaml:"regex"`
+	Value string `yaml:"value"`
```
brian-brazil: The value should default to ($1)

config/config.go
```diff
@@ -256,6 +253,43 @@ func (c *Auth) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+	XXX map[string]interface{} `yaml:",inline"`
+}
+
+func (c *RegexpExtract) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type plain RegexpExtract
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if err := CheckOverflow(c.XXX, "module"); err != nil {
```
brian-brazil: regexp_extract, not module

generator/config.go
```diff
@@ -20,11 +20,28 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+	XXX map[string]interface{} `yaml:",inline"`
+}
+
+func (c *MetricOverrides) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type plain MetricOverrides
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if err := config.CheckOverflow(c.XXX, "module"); err != nil {
```
brian-brazil: overrides

generator/tree.go
```diff
@@ -215,6 +215,15 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 		}
 	}
 
+	// Apply module config overrides to their corresponding metrics.
```
brian-brazil: This should be unittested on the generator side

collector.go
```diff
@@ -191,25 +201,41 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+
+func generateOverrideMetrics(metric *config.Metric, pduValue string, labelnames, labelvalues []string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.Overrides {
+		for _, strMetric := range strMetricSlice {
+			if strMetric.Regex.MatchString(pduValue) {
+				value := strMetric.Regex.ReplaceAllString(pduValue, strMetric.Value)
+				v, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					log.Errorf("Error parsing float64 from value: %v", value)
```
brian-brazil: This would be a debug log at most, and you need to `continue` when it happens and try the next regex

collector.go
```diff
@@ -191,25 +201,41 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+			t, value, labelvalues...))
+	}
+	return res
+}
+
+func generateOverrideMetrics(metric *config.Metric, pduValue string, labelnames, labelvalues []string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.Overrides {
+		for _, strMetric := range strMetricSlice {
+			if strMetric.Regex.MatchString(pduValue) {
```
brian-brazil: Look at https://github.com/prometheus/prometheus/blob/master/relabel/relabel.go#L57 for how to approach this

collector.go
```diff
@@ -191,25 +201,41 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+				value := strMetric.Regex.ReplaceAllString(pduValue, strMetric.Value)
+				v, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					log.Errorf("Error parsing float64 from value: %v", value)
+				}
+				newMetric := prometheus.MustNewConstMetric(prometheus.NewDesc(metric.Name+name, metric.Help, labelnames, nil),
+					prometheus.GaugeValue, v, labelvalues...)
+				results = append(results, newMetric)
```
brian-brazil: If we've gotten this far, we need to `break`

Conorbro:
@brian-brazil

collector.go
```diff
@@ -191,25 +201,43 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
-	labelvalues := make([]string, 0, len(labels)+1)
-	for k, v := range labels {
-		labelnames = append(labelnames, k)
-		labelvalues = append(labelvalues, v)
-	}
-	// For strings we put the value as a label with the same name as the metric.
-	// If the name is already an index, we do not need to set it again.
-	if stringType {
+		if len(metric.RegexpExtracts) > 0 {
+			return generateOverrideMetrics(metric, pdu.Value.(string), labelnames, labelvalues)
```
brian-brazil: You should have a test where the value is a number rather than a string.

collector.go
```diff
@@ -191,25 +201,43 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+		// If the name is already an index, we do not need to set it again.
 		if _, ok := labels[metric.Name]; !ok {
 			labelnames = append(labelnames, metric.Name)
 			labelvalues = append(labelvalues, pduValueAsString(pdu, metric.Type))
 		}
 	}
-	return prometheus.MustNewConstMetric(prometheus.NewDesc(metric.Name, metric.Help, labelnames, nil),
-		t, value, labelvalues...)
+
+	if len(res) == 0 {
```
brian-brazil: You don't need this check

collector.go
```diff
@@ -191,25 +201,43 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+		res = append(res, prometheus.MustNewConstMetric(prometheus.NewDesc(metric.Name, metric.Help, labelnames, nil),
```
brian-brazil: You know there's only one element in the list, you can create a literal here.

collector.go
```diff
@@ -191,25 +201,43 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+			t, value, labelvalues...))
+	}
+	return res
+}
+
+func generateOverrideMetrics(metric *config.Metric, pduValue string, labelnames, labelvalues []string) []prometheus.Metric {
```
brian-brazil: This is applying the regex extracts, not all overrides

collector.go
```diff
@@ -191,25 +201,43 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.RegexpExtracts {
+		for _, strMetric := range strMetricSlice {
+			if strMetric.Regex.MatchString(pduValue) {
+				value := strMetric.Regex.ReplaceAllString(pduValue, strMetric.Value)
```
brian-brazil: This needs to use the same logic as relabelling does so there's full regex functionality.

collector.go
```diff
@@ -191,25 +201,43 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+				v, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					log.Debugf("Error parsing float64 from value: %v", value)
+					continue
+				}
+				newMetric := prometheus.MustNewConstMetric(prometheus.NewDesc(metric.Name+name, metric.Help, labelnames, nil),
```
brian-brazil: In the Help you can mention that this was regex extracted

collector_test.go
```diff
@@ -2,26 +2,57 @@ package main
...
 			pdu: &gosnmp.SnmpPDU{
 				Name:  "1.1.1.1.1",
+				Value: "SomeStringValue",
+			},
+			indexOids: []int{},
+			metric: &config.Metric{
+				Name:           "snIfOpticalLaneMonitoringTemperature",
+				Oid:            "1.1.1.1.1",
+				Help:           "HelpText",
+				RegexpExtracts: strMetrics,
```
brian-brazil: No need for an extra variable, you can put this in literally

generator/config.go
```diff
@@ -20,11 +20,28 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
-type ModuleConfig struct {
-	Walk    []string  `yaml:"walk"`
-	Lookups []*Lookup `yaml:"lookups"`
+type MetricOverrides struct {
+	RegexpExtracts map[string][]config.RegexpExtract `yaml:"regexp_extract,omitempty"`
```
brian-brazil: We'd said to remove the p

Conorbro:
@brian-brazil

collector.go
```diff
@@ -191,25 +201,49 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+func applyRegexExtracts(metric *config.Metric, pduValue string, labelnames, labelvalues []string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.RegexpExtracts {
+		for _, strMetric := range strMetricSlice {
+			indexes := strMetric.Regex.FindStringSubmatchIndex(pduValue)
+			if indexes == nil {
+				continue
+			}
+			res := strMetric.Regex.ExpandString([]byte{}, strMetric.Value, pduValue, indexes)
+			if len(res) == 0 {
```
brian-brazil: You don't need this, the parsing should catch it

collector_test.go
```diff
@@ -2,26 +2,119 @@ package main
...
+	"github.com/prometheus/client_model/go"
 	"github.com/soniah/gosnmp"
 
-	"github.com/prometheus/client_model/go"
 	"github.com/prometheus/snmp_exporter/config"
 )
 
 func TestPduToSample(t *testing.T) {
+	var regexpMatch, regexpNoMatch config.Regexp
+	regexpMatch.Regexp, _ = regexp.Compile(".*")
```
brian-brazil: Put these inline in the struct, it's easier to understand the tests that way. You can use MustCompile

collector_test.go
```diff
@@ -2,26 +2,119 @@ package main
...
+				Help: "HelpText",
+				RegexpExtracts: map[string][]config.RegexpExtract{
+					"Extension": []config.RegexpExtract{
+						{
+							Regex: regexpMatch,
+							Value: "",
+						},
+					},
+				},
+			},
```
brian-brazil: It'd be a bit clearer if you put the nil result here explicitly

collector_test.go
```diff
@@ -2,26 +2,119 @@ package main
...
+				RegexpExtracts: map[string][]config.RegexpExtract{
+					"Status": []config.RegexpExtract{
+						{
+							Regex: regexpMatch,
+							Value: "5",
+						},
+					},
+				},
+			},
+			expectedMetrics: []string{},
```
brian-brazil: Can you test multiple outputted metrics, multiple extract rules and expansion?

Conorbro:
@brian-brazil

collector.go
```diff
@@ -191,25 +201,47 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
-		labelnames = append(labelnames, k)
-		labelvalues = append(labelvalues, v)
-	}
-	// For strings we put the value as a label with the same name as the metric.
-	// If the name is already an index, we do not need to set it again.
-	if stringType {
+		if len(metric.RegexpExtracts) > 0 {
+			v, ok := pdu.Value.(string)
+			if !ok {
+				log.Errorf("Invalid PDU value type: got %v, want string", reflect.TypeOf(pdu.Value))
```
brian-brazil: You could use %T here which will pull out the type

collector.go
```diff
@@ -191,25 +201,47 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+	for name, strMetricSlice := range metric.RegexpExtracts {
+		for _, strMetric := range strMetricSlice {
+			indexes := strMetric.Regex.FindStringSubmatchIndex(pduValue)
+			if indexes == nil {
+				continue
+			}
+			res := strMetric.Regex.ExpandString([]byte{}, strMetric.Value, pduValue, indexes)
+			v, err := strconv.ParseFloat(string(res), 64)
+			if err != nil {
+				log.Debugf("Error parsing float64 from value: %v", res)
```
brian-brazil: Including the metric name here would be useful, plus other log lines where you can

collector_test.go
```diff
@@ -2,26 +2,171 @@ package main
...
+						{
+							Regex: config.Regexp{
+								regexp.MustCompile(".*"),
+							},
+							Value: "",
+						},
+					},
+				},
+			},
+			expectedMetrics: []string{},
```
brian-brazil: Put this on the others too

collector_test.go
```diff
@@ -2,26 +2,171 @@ package main
...
+						{
+							Regex: config.Regexp{
+								regexp.MustCompile(".*"),
+							},
+							Value: "5",
+						},
+					},
+					"Blank": []config.RegexpExtract{
+						{
+							Regex: config.Regexp{
```
brian-brazil: Can you test multiple regexes too? We want to ensure first match wins.

Conorbro:
@brian-brazil

collector_test.go
```diff
@@ -2,26 +2,175 @@ package main
...
-		metric             *config.Metric
-		oidToPdu           map[string]gosnmp.SnmpPDU
-		expectedMetric     string
-		expectedMetricDesc string
+		pdu             *gosnmp.SnmpPDU
+		indexOids       []int
+		metric          *config.Metric
+		oidToPdu        map[string]gosnmp.SnmpPDU
+		expectedMetrics map[string]string
+		// expectedMetricDescs []string
```
brian-brazil: Why is this commented out?

collector_test.go
```diff
@@ -65,24 +212,28 @@ func TestPduToSample(t *testing.T) {
...
 		}
-		if m.Desc().String() != c.expectedMetricDesc {
-			t.Fatalf("Unexpected metric description: got %v, want %v", m.Desc().String(), c.expectedMetricDesc)
+		metric := &io_prometheus_client.Metric{}
+		for _, m := range metrics {
+			err := m.Write(metric)
+			if err != nil {
+				t.Fatalf("Error writing metric: %v", err)
+			}
+			if _, ok := c.expectedMetrics[metric.String()]; !ok {
```
brian-brazil: If you're going to do things that way, you also need to check that the length of the expected metrics matches what you actually got.

collector_test.go
```diff
@@ -2,26 +2,175 @@ package main
...
+					},
+					"Extension": []config.RegexpExtract{
+						{
+							Regex: config.Regexp{
+								regexp.MustCompile(".*"),
+							},
+							Value: "5",
+						},
+					},
+					"MultipleRegexes": []config.RegexpExtract{
```
brian-brazil: This is still only a single regex

collector.go
brian-brazil: You missed this comment

collector_test.go
Conorbro: That's checked on line 222.

Conorbro:
@brian-brazil

collector_test.go
```diff
@@ -2,26 +2,180 @@ package main
...
```
brian-brazil: I'd have three regexes, and make the second one match - but the 3rd one would also match. That way we know it's trying them in turn.

Conorbro:
@brian-brazil

collector.go
```diff
@@ -191,25 +200,48 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+		t, value, labelvalues...)}
+}
+
+func applyRegexExtracts(metric *config.Metric, pduValue string, labelnames, labelvalues []string) []prometheus.Metric {
+	results := []prometheus.Metric{}
+	for name, strMetricSlice := range metric.RegexpExtracts {
+		for _, strMetric := range strMetricSlice {
+			indexes := strMetric.Regex.FindStringSubmatchIndex(pduValue)
+			if indexes == nil {
+				log.Debugf("No match found for regexp: %v against value: %v for metric %v", strMetric.Regex.String(), pduValue, name)
```
brian-brazil: `name` is the name of the extract, not the metric. You want `metric.Name`.

generator/README.md
```diff
@@ -71,6 +71,18 @@ modules:
       # with that value.
       - old_index: bsnDot11EssIndex
         new_index: bsnDot11EssSsid
+     
+     overrides: # Allows for per-module overrides of bits of MIBs
+       metricName:
+         regexp_extract:
+           Temp: # A new metric will be created appending this to the metricName to become metricNameTemp.
+             - regex: '.*' # Regex to extract a value from the returned SNMP walks's value.
```
brian-brazil: For the example to make sense, this should be (.*) so that there is a $1

Also mention it's a float64 value

generator/config.go
```diff
@@ -20,11 +20,28 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
-type ModuleConfig struct {
-	Walk    []string  `yaml:"walk"`
-	Lookups []*Lookup `yaml:"lookups"`
+type MetricOverrides struct {
+	RegexpExtracts map[string][]config.RegexpExtract `yaml:"regex_extract,omitempty"`
```
brian-brazil: The field name here doesn't match the docs or unittest.

brian-brazil:
Format.md also needs updating.

Conorbro:
@brian-brazil

generator/README.md
```diff
@@ -71,6 +71,18 @@ modules:
       # with that value.
       - old_index: bsnDot11EssIndex
         new_index: bsnDot11EssSsid
+
+     overrides: # Allows for per-module overrides of bits of MIBs
+       metricName:
+         regex_extracts:
+           Temp: # A new metric will be created appending this to the metricName to become metricNameTemp.
+             - regex: '(.*)' # Regex to extract a value from the returned SNMP walks's value.
+               value: '$1' # Defaults to $1, float64.
```
brian-brazil: "the result will be parsed as a float64"

generator/FORMAT.md
```diff
@@ -45,4 +45,11 @@ module_name:
          oid: 1.3.6.1.2.1.2.2.1.2  # OID to look under.
          labelname: ifDescr        # Output label name.
          type: OctetString         # Type of output object.
+     # Overrides allows for per-module overrides of bits of MIBs
```
brian-brazil: This isn't the format of the exporter's yml, it's just the regex_extracts.

collector.go
```diff
@@ -191,25 +200,48 @@ func pduToSample(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oi
...
+		for _, strMetric := range strMetricSlice {
+			indexes := strMetric.Regex.FindStringSubmatchIndex(pduValue)
+			if indexes == nil {
+				log.Debugf("No match found for regexp: %v against value: %v for metric %v", strMetric.Regex.String(), pduValue, metric.Name)
+				continue
+			}
+			res := strMetric.Regex.ExpandString([]byte{}, strMetric.Value, pduValue, indexes)
+			v, err := strconv.ParseFloat(string(res), 64)
+			if err != nil {
+				log.Debugf("Error parsing float64 from value: %v for metric: %v", res, name)
```
brian-brazil: metric.Name here too

Conorbro:
@brian-brazil

brian-brazil:
Thanks!

PR #218, opened 2017-08-15
Remove Fortinet for defaults, the MIB isn't publicly available.

brian-brazil:
Also do general output refresh, changes verified as correct
with current code.

@Conorbro


Conorbro:
LGTM

PR #229, opened 2017-08-29
Use textual convention and handle more type hints

### LINKED ISSUE #164: Use tc_index to get text convention
Currently we rely on the display hint for things like DisplayString, however some MIBs depend on RFC1213-MIB which is too old to have a hint. The tc_index can be used to get the original text convention (SYNTAX) in the MIB, so we can still set the right type for such things.

### LINKED ISSUE #189: Support SnmpAdminString
This has a display hint of `255t` and is UTF-8.

brian-brazil:
Fixes #189 
Fixes #164 

@SuperQ

### DIFF
--- a/generator/FORMAT.md
+++ b/generator/FORMAT.md
@@ -19,7 +19,7 @@ module_name:
      #   gauge:   An integer with type gauge.
      #   counter: An integer with type counter.
      #   OctetString: A bit string, rendered as 0xff34.
-     #   DisplayString: An ASCII string.
+     #   DisplayString: An ASCII or UTF-8 string.
      #   PhysAddress48: A 48 bit MAC address, rendered as 00:01:02:03:04:ff.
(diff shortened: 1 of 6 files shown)


SuperQ:
Yup, this fixes @fjaeckel's test case:

```
test:
  walk:
  - 1.3.6.1.4.1.1991.1.1.3.3.6.1.2
  metrics:
  - name: snIfOpticalMonitoringTxPower
    oid: 1.3.6.1.4.1.1991.1.1.3.3.6.1.2
    type: DisplayString
    help: This object holds the value of the transmitter optical signal power for
      the interface, measured in dBm, followed by whether this is a normal value,
      or high/low warning or alarm - 1.3.6.1.4.1.1991.1.1.3.3.6.1.2
    indexes:
    - labelname: ifIndex
      type: gauge
```

PR #245, opened 2017-11-10
Fixed size index

brian-brazil:
Per RFC 1212 4.1.6(2)

@SuperQ

### DIFF
--- a/generator/FORMAT.md
+++ b/generator/FORMAT.md
@@ -32,6 +32,11 @@ module_name:
      indexes:
       - labelname: ifIndex
         type: gauge
+      - labelname: someString
+        type: OctetString
+        fixed_size: 8   # Only possible for OctetString/DisplayString types.
+                        # If only one length is possible this is it. Otherwise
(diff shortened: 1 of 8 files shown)


generator/FORMAT.md
```diff
@@ -32,6 +32,8 @@ module_name:
      indexes:
       - labelname: ifIndex
         type: gauge
+        fixed_size: 0   # Only possible for OctetString/DisplayString.
```
SuperQ: I would suggest making this a separate example, to make it more clear that this is for `type: DIsplayString`.  Also, comment that it's an optional field.

brian-brazil: Done

brian-brazil:
Looks like this doesn't work on OSX for some reason I can't determine, I'm presuming some compiler oddity there as netsnmp has a NULL ranges yet snmptranslate shows the range. This all works fine on Ubuntu.

brian-brazil:
It also works on Red Hat, so I'm going with OSX weirdness.

PR #258, opened 2018-01-08
Dockerfile for generator: also install snmp-mibs-downloader

huiyiqun:
For up-to-date mibs

### DIFF
--- a/generator/Dockerfile
+++ b/generator/Dockerfile
@@ -1,6 +1,9 @@
 FROM golang:latest
 
-RUN apt-get update && apt-get install -y libsnmp-dev
+# Enable non-free
+RUN sed -i -e 's/$/ contrib non-free/g' /etc/apt/sources.list
+
+RUN apt-get update && apt-get install -y libsnmp-dev snmp-mibs-downloader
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
That package doesn't contain the vendor MIBs that users typically want, and they may get in the way of the MIBs they're explicitly providing.

huiyiqun:
However, it provide standard MIBs for generic usage.

In fact, I couldn't event resolve oid such as `ifTable`. It's strange, but installing snmp-mibs-downloader solve this problem.

brian-brazil:
You need far more than what snmp-mibs-downloader provides for the included configuration to work.

rossigee:
FWIW, I needed to apply this in order to follow the instructions in the README.md, otherwise it wouldn't resolve 'sysUpTime' or any of the other oids mentioned in the example.

cloph:
definitely needed - while not enough to process the provided generator.yml (which is not the point), it is **required** to process custom MIBs, so I fully support this request.

ntavares:
@brian-brazil I don't understand why this patch hasn't simply been merged, if it solves people's problem. +1 for the merge

brian-brazil:
There's a makefile these days.

PR #263, opened 2018-02-01
WIP: Add override for metric type.

SuperQ:
# TODO
* [ ] Validate config for allowed types.
* [ ] Add tests.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -100,6 +100,7 @@ modules:
                value: '1'
              - regex: '.*'
                value: '0'
+         type: [counter, gauge, DisplayString, ...] # Override the metric type.
 ```
 
 ## Where to get MIBs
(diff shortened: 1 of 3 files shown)


generator/README.md
```diff
@@ -98,6 +98,7 @@ modules:
                value: '1'
              - regex: '.*'
                value: '0'
+         type: [counter, gauge, DisplayString, ...] # Override the metric type.
```
brian-brazil: You should link over to the FORMAT doc as to what these are

generator/tree.go
```diff
@@ -214,6 +214,9 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 					if !ok {
 						log.Fatalf("Unknown index type %s for %s", indexNode.Type, lookup.NewIndex)
 					}
+					if cfg.Overrides[sanitizeLabelName(indexNode.Label)].Type != "" {
```
brian-brazil: This is not the only place types are used, you need to produce a copy of the tree with this changed.

SuperQ:
Yes, I know this needs a bunch more work.

brian-brazil:
Obsoleted by #294

PR #268, opened 2018-02-19
update gosnmp and add support for opaque float/double

mschoch:
updated vendored version of gosnmp to latest, which now supports
returning values from opaque float and double

updated generator to recognize opaque float/double and configure
it to work as a gauge

updated collector to extract these values correctly

@brian-brazil

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -200,6 +200,8 @@ func pduToSamples(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, o
 		t = prometheus.CounterValue
 	case "gauge":
 		t = prometheus.GaugeValue
+	case "Float", "Double":
+		t = prometheus.GaugeValue
 	default:
 		// It's some form of string.
 		t = prometheus.GaugeValue
@@ -267,6 +269,10 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ string) string {
 		return strconv.FormatUint(uint64(pdu.Value.(uint)), 10)
 	case uint64:
 		return strconv.FormatUint(pdu.Value.(uint64), 10)
+	case float32:
+		return strconv.FormatFloat(float64(pdu.Value.(float32)), 'f', -1, 32)
+	case float64:
+		return strconv.FormatFloat(pdu.Value.(float64), 'f', -1, 64)
 	case string:
 		if pdu.Type == gosnmp.ObjectIdentifier {
 			// Trim leading period.
--- a/collector.go
+++ b/collector.go
@@ -172,6 +172,10 @@ func getPduValue(pdu *gosnmp.SnmpPDU) float64 {
 	switch pdu.Type {
 	case gosnmp.Counter64:
 		return float64(gosnmp.ToBigInt(pdu.Value).Uint64())
+	case gosnmp.OpaqueFloat:
+		return float64(pdu.Value.(float32))
+	case gosnmp.OpaqueDouble:
+		return pdu.Value.(float64)
 	default:
 		return float64(gosnmp.ToBigInt(pdu.Value).Int64())
 	}
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -95,12 +95,10 @@ func prepareTree(nodes *Node) map[string]*Node {
 		}
 	})
 
-	// Promote Opaque Float/Double textual convention to type
+	// Promote Opaque Float/Double textual convention to type.
 	walkNode(nodes, func(n *Node) {
-		if n.Type == "OPAQUE" {
-			if n.TextualConvention == "Float" || n.TextualConvention == "Double" {
-				n.Type = n.TextualConvention
-			}
+		if n.TextualConvention == "Float" || n.TextualConvention == "Double" {
+			n.Type = n.TextualConvention
 		}
 	})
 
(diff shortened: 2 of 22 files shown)


generator/tree.go
```diff
@@ -95,12 +95,24 @@ func prepareTree(nodes *Node) map[string]*Node {
 		}
 	})
 
+	// Promote Opaque Float/Double textual convention to type
+	walkNode(nodes, func(n *Node) {
+		if n.Type == "OPAQUE" {
+			if strings.EqualFold(n.TextualConvention, "float") {
```
brian-brazil: We do case-sensitive comparisons everywhere else

generator/tree.go
```diff
@@ -95,12 +95,24 @@ func prepareTree(nodes *Node) map[string]*Node {
...
+				n.Type = "FLOAT"
```
brian-brazil: `Float`, as that's what's in the MIB

generator/tree_test.go
```diff
@@ -782,6 +791,42 @@ func TestGenerateConfigModule(t *testing.T) {
 				},
 			},
 		},
+		// Opaque Float becomes gauge
```
brian-brazil: These go up with the metric type case

collector.go
```diff
@@ -172,6 +172,10 @@ func getPduValue(pdu *gosnmp.SnmpPDU) float64 {
 	switch pdu.Type {
```
brian-brazil: pduValueAsString also needs updating.

brian-brazil:
FORMAT.md will also need updating.

mschoch:
What is your preferred way for subsequent changes, additional commits to this branch? Or force push amended commit?

brian-brazil:
Additional commits tend to be easier to review.

mschoch:
Can you clarify how FORMAT.md is updated/used?  The text at the top indicates that it is automatically generated, but I don't see a description of how that is done.

brian-brazil:
FORMAT.md is the documentation for what snmp.yml is. You're adding new types, so you need to add them in there by hand.

mschoch:
Thanks, I'm still a bit confused, as this change results in Float and Double being generated as a gauge in the resulting snmp.yml.  My understanding was that at the prometheus level everything had to turn into gauge or counter anyway.  But I do see other types documented in this file, so maybe my understanding and part of this PR is wrong.

mschoch:
I think maybe I get it now, perhaps instead of turning Float and Double into a gauge, they should stay as their own type in the snmp.yml.  Then that additional type information might let you format or otherwise handle it a bit differently inside the snmp_exporter.  Then at the level of prometheus scraping it, they all become gauages or counters anyway.  Does this sound correct?

brian-brazil:
My question would be how does gosnmp know that a value is a float? That information is only available from the MIB afaik, which gosnmp does not have access to.

mschoch:
What I learned when I hacked the gosnmp portion was that the opaque type values are not just arbitrary bytes, but actually double-wrapped ASN.1 values.  The outer wrapper has type opaque (0x44) and the inner value has types 0x78 (float) or 0x78 (double).

It appears to me that opaque was included initially for extensibility, but it was under-specified and became confusing.  Later, they deprecated it, but there are a few types like Float, Double and possibly a few others that are still used.

- https://tools.ietf.org/html/draft-perkins-opaque-01
- https://tools.ietf.org/html/draft-perkins-float-00

Note, both of those are expired drafts, so it's possible this is not formal, but just a convention people happen to use (but following this did allow me to successfully parse the data I found in the real world).

As for your original question, both the MIB textual convention and type inside wrapped opaque value give us indication the value should be interpreted as float/double.

brian-brazil:
To be consistent let's work off the textualconvention like we do elsewhere then. This will avoid being over-eager in detecting such values.

mschoch:
Thanks for helping work through this, I'm still not exactly clear how you'd like it to proceed.  This PR does have the generator detect Float/Double based on the textual convention in the MIB.  I've changed it now to keep Float/Double as types in that snmp.yml.  I've documented these types in FORMAT.md.  However, I then also had to add some code to properly render these as a guage when scraped by prometheus.  I've also tried to incorporate your other feedback.  Hopefully this gets us a little closer.

generator/tree.go
```diff
@@ -95,6 +95,15 @@ func prepareTree(nodes *Node) map[string]*Node {
 		}
 	})
 
+	// Promote Opaque Float/Double textual convention to type
```
brian-brazil: Comments should end in a full stop.

generator/tree.go
```diff
@@ -95,6 +95,15 @@ func prepareTree(nodes *Node) map[string]*Node {
...
+	walkNode(nodes, func(n *Node) {
+		if n.Type == "OPAQUE" {
```
brian-brazil: I'd expect that checking the textual convention is enough

generator/tree.go
```diff
@@ -113,6 +122,10 @@ func metricType(t string) (string, bool) {
 		return "InetAddress", true
 	case "PhysAddress48", "DisplayString":
 		return t, true
+	case "Float":
```
brian-brazil: You can combine this into the previous case

mschoch:
Pushed commit addressing most recent review comments.

brian-brazil:
Thanks!

PR #272, opened 2018-03-08
Configure useful OIDs for pfSense.

smlx:
Add a usable generator configuration for pfSense, along with links for MIBs.


brian-brazil:
Using the node exporter would be preferred here over the snmp exporter.

brian-brazil:
That doesn't change that SNMP is not the best way to monitor this, and we don't want to encourage the use of SNMP where there's better options.

PR #275, opened 2018-03-28
Add default command line args for target and module

rguliyev:


brian-brazil:
This is not how the snmp exporter is intended to be used, this should be coming from URL parameters from Prometheus.

rguliyev:
This adds default parameters for people who want to rely on defaults, you can still pass them via URL to override the defaults. This change moves the hardcode of "if_mib" to a definable default, which makes the code cleaner.

SuperQ:
I would like the default module as a flag, that sounds like a good idea, but I don't think a default target is useful.

RichiH:
I can see this being used during manual testing etc, but what is the use
case with service discovering Prometheus?

Not opposed as such, just don't see the use case atm.

PR #277, opened 2018-03-31
add: MikroTik generator config and output

lukassup:


generator/generator.yml
```diff
@@ -218,3 +218,14 @@ modules:
...
+# MiktoTik
+#
+# http://download2.mikrotik.com/Mikrotik.mib
+#
+  mikrotik:
+    walk:
+      - sysUpTime
+      - interfaces
+      - ifXTable
+      - 1.3.6.1.4.1.14988.1.1
```
brian-brazil: What's this object's name?

lukassup: `snmptranslate` gives me this:
```
.iso.org.dod.internet.private.enterprises.mikrotik.mikrotikExperimentalModule.mtXRouterOs
```

brian-brazil: We don't usually pull in entire MIBs. Can you be more specific about which parts of it you want?

SuperQ:
Based on what I see in the MIB, there whole table is probably a bad idea.  There are some things indexed by full MAC address.  I think these walks should be safe / useful.

```
- mtxrWlStatTable
- mtxrWlApTable
- mtxrDnStatTable
- mtxrInterfaceStatsTable
- mtxrHealth
- mtxrLicense
- mtxrDHCPLeaseCount
```

brian-brazil:
This has gone stale, if you're still interested in working on this let us know.

PR #282, opened 2018-04-12
add support for module include of configuration file

0086:
Only "walk" and "metrics" will be included, SNMP parameters will not be included.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -66,22 +66,26 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
 	defer snmp.Conn.Close()
 
 	result := []gosnmp.SnmpPDU{}
-	for _, subtree := range config.Walk {
-		var pdus []gosnmp.SnmpPDU
-		log.Debugf("Walking target %q subtree %q", snmp.Target, subtree)
-		walkStart := time.Now()
-		if snmp.Version == gosnmp.Version1 {
-			pdus, err = snmp.WalkAll(subtree)
-		} else {
-			pdus, err = snmp.BulkWalkAll(subtree)
+	modules := append(config.IncludedModules, config)
+	for _, module := range modules {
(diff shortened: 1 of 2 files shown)


brian-brazil:
What are you trying to do?

0086:
I don't want to waste time copying oid between different modules.

0086:
sample configuration file:
```
default:
  walk:
  - 1.3.6.1.2.1.1.3
  metrics:
  - name: sysUpTime
    oid: 1.3.6.1.2.1.1.3
    type: gauge
    help: The time since the system was last re-initialized.

10.10.10.1:
  include:
  - default

  walk:
  - 1.3.6.1.2.1.1.5
  metrics:
  - name: sysName
    oid: 1.3.6.1.2.1.1.5
    type: DisplayString
    help: The name of system

```

brian-brazil:
You should be using the generator to create your snmp.yml. There's no plans to offer such a feature, as either your generator config is simple enough that it doesn't matter or it's sufficiently complex that you need configuration management anyway.

0086:
We have a large number of machines that need to be monitored and the machines are dynamically added and deleted. 
Some OIDs are common to all machines. Some OIDs are common to a group of machines. Each machine may also have some unique OIDs.

SuperQ:
This can already be done with YAML merge syntax.  There's no need for a code change.

```
sysuptime: &sysuptime
  walk:
  - 1.3.6.1.2.1.1.3
  metrics:
  - name: sysUpTime
    oid: 1.3.6.1.2.1.1.3
    type: gauge
    help: The time (in hundredths of a second) since the network management portion
      of the system was last re-initialized. - 1.3.6.1.2.1.1.3

some_other_module:
  <<: *sysuptime
  walk:
  - 1.3.6.1.2.1.2
  metrics:
    ...
```

0086:
@SuperQ Thanks a lot, I will try it later.

PR #284, opened 2018-04-14
Implement strict yaml unmarshalling.

manosf:
@brian-brazil I believe that this is the last Prometheus' component to require yaml unmarshalling changes.


brian-brazil:
Thanks!

PR #287, opened 2018-04-26
Fix SNMP GET in #283

sgissi:
snmp_exporter: Fix unsupported OIDs collection.
snmp_exporter: Added Debug log before collection.
generator: Rename OidDirect to OidScalar to reflect SNMP naming convention.
generator: Added trailing "0" index for scalar OIDs collection.


brian-brazil:
You seem to be missing DCO.

brian-brazil:
Thanks!

brian-brazil:
Something is still not right here, on my laptop I'm getting 0 for sysUpTime even though it should have a value

brian-brazil:
Yeah, it looks like it doesn't deal gracefully with just one variable in a get not existing.

PR #292, opened 2018-05-02
Fix SNMP GET and SNMPv3 auth configuration

sgissi:
Fixed SNMPv1 GET of unsupported OIDs.
Remove SNMPv3 default auth/priv protocols. Cannot use noAuthnoPriv if auth/priv protocols are set even if no keys are provided.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -68,8 +68,8 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
 	result := []gosnmp.SnmpPDU{}
 	getOids := config.Get
 	maxOids := int(config.WalkParams.MaxRepetitions)
-	// Max Repetition can be 0, maxOids cannot.
-	if maxOids == 0 {
+	// Max Repetition can be 0, maxOids cannot. SNMPv1 can only tell one OID error per call.
+	if maxOids == 0 || snmp.Version == gosnmp.Version1 {
 		maxOids = 1
 	}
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -28,8 +28,6 @@ var (
 	DefaultAuth = Auth{
 		Community:     "public",
 		SecurityLevel: "noAuthNoPriv",
-		AuthProtocol:  "MD5",
```
brian-brazil: This is a breaking change, users may not have these in their configs currently. I don't see a reason to remove them.

collector.go
```diff
@@ -68,8 +68,8 @@ func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error
 	result := []gosnmp.SnmpPDU{}
 	getOids := config.Get
 	maxOids := int(config.WalkParams.MaxRepetitions)
-	// Max Repetition can be 0, maxOids cannot.
-	if maxOids == 0 {
+	// Max Repetition can be 0, maxOids cannot. SNMPv1 can only tell one OID error per call.
```
brian-brazil: report

config/config.go
sgissi: Fair enough, I can open another issue/PR to discuss how to handle noAuthNoPriv as currently the default is broken (leads to error in gosnmp for using noAuthNoPriv with Authentication protocol defined).

PR #294, opened 2018-05-09
Generator metric type override

sgissi:


SuperQ:
Thanks, a good test case for this is under the `cisco_wlc` module.

`bsnAPName` should have an override to be `DisplayString`.

generator/tree.go
```diff
@@ -200,11 +200,54 @@ func getMetricNode(oid string, node *Node, nameToNode map[string]*Node) (*Node,
...
+		if params.Type == "" {
+			continue
+		}
+		_, ok := nameToNode[name]
+		if !ok {
+			log.Warnf("Could not find metric '%s' to override type", name)
+			continue
+		}
+		// Type validated at generator configuration.
+		typeOverrides[name], _ = metricType(params.Type)
```
brian-brazil: The type should already be the right value, no need to use `metricType`.

generator/tree.go
```diff
@@ -200,11 +200,54 @@ func getMetricNode(oid string, node *Node, nameToNode map[string]*Node) (*Node,
...
+	}
+	return &newNode
+}
+
 func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*Node) *config.Module {
 	out := &config.Module{}
 	needToWalk := map[string]struct{}{}
 	tableInstances := map[string][]string{}
 
+	// Gather all type overrides for the current module.
```
brian-brazil: I think this can be done in one pass

generator/tree_test.go
```diff
@@ -137,23 +137,30 @@ func TestGenerateConfigModule(t *testing.T) {
...
 
-	overrides := make(map[string]MetricOverrides)
-	metricOverrides := MetricOverrides{
+	regexpOverrides := make(map[string]MetricOverrides)
+	regexpOverrides["root"] = MetricOverrides{
 		RegexpExtracts: strMetrics,
 	}
-	overrides["root"] = metricOverrides
+
+	typeOverrides := make(map[string]MetricOverrides)
```
brian-brazil: You're working with base types here, so you can inline the config as usual.

generator/tree_test.go
```diff
@@ -396,6 +402,80 @@ func TestGenerateConfigModule(t *testing.T) {
...
+					{
+						Name: "node2",
+						Oid:  "1.2",
+						Type: "DisplayString",
+						Help: " - 1.2",
+					},
+				},
+			},
+		},
+		// Table with type override
```
brian-brazil: Full stop

generator/tree.go
sgissi: By gathering first we can make sure there are type overrides in the configuration and only duplicate the tree when actually needed.

sgissi: It is possible to do in one pass and do the duplication lazily, flow will suffer a bit though.

generator/tree.go
```diff
@@ -205,6 +205,34 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 	needToWalk := map[string]struct{}{}
 	tableInstances := map[string][]string{}
 
+	// Apply type overrides for the current module.
+	newTree := false
+	for name, params := range cfg.Overrides {
+		if params.Type == "" {
+			continue
+		}
+		// Make a copy of the tree if not done yet.
```
brian-brazil: Hmm, I think this would be simplest to always be done in the function that calls this one. Performance isn't the primary goal of the generator, it's a rarely run batch process.

brian-brazil:
Thanks!

PR #296, opened 2018-05-15
Unifi SNMP Version Fix

RustyBower:
The existing UAPs leverage 'tinysnmp' which is a lightweight SNMP daemon, which in turn only supports SNMPv1.

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -10219,3 +10219,4 @@ ubiquiti_unifi:
     oid: 1.3.6.1.4.1.41112.1.6.3.6
     type: DisplayString
     help: ' - 1.3.6.1.4.1.41112.1.6.3.6'
+  version: 1


sgissi:
Hi, two comments: The snmp.yml file is auto-generated, you want to update the generator/generator.yml file. In your commit you need the sign-off the patch (https://github.com/prometheus/prometheus/wiki/DCO-signing). Hope that helps!

SuperQ:
Which Unifi models only have v1?  I have several models that work fine with v2 and v3.  Are you sure your firmware is up to date?

RustyBower:
So specifically the ones I am testing against are "Unifi AP-AC v2", granted those are EOL come November.

Because of that, the firmware version that I am stuck on for these is 5.6.30.

SuperQ:
Yes, it looks like the original AC models are the only ones stuck on the 3.8 firmware series.  The 3.9 firmware works with SNMP v2/v3.  I don't think it's worth downgrading everything else to v1 by default for a soon-to-be EOL model.

RustyBower:
Makes sense. I'll close this out for now.

SuperQ:
If you find any other useful ubnt OIDs, please feel free to add them to the generator.yml.

Issue #286, opened 2018-04-20
Use the Makefile.common


krasi-georgiev:
https://github.com/prometheus/prometheus/blob/master/Makefile.common

PR #299, opened 2018-05-17
Update build

SuperQ:
* Update to `Makefile.common`.
* Update to CircleCI 2.0
* Cleanup unused vendor packages.

Closes: https://github.com/prometheus/snmp_exporter/issues/291
Closes: https://github.com/prometheus/snmp_exporter/issues/286


.circleci/config.yml
```diff
@@ -10,6 +10,7 @@ jobs:
 
     steps:
     - checkout
+    - run: sudo apt-get -y install build-essential libsnmp-dev
```
brian-brazil: When we template this all out later on, we should just install this everywhere

Issue #321, opened 2018-08-14
Support for RFC 2579 DateAndTime conversion


SuperQ:
It would be useful to convert RFC 2579 DateAndTime to a unix timestamp.

From SNMPv2-TC:
```
DateAndTime ::= TEXTUAL-CONVENTION
    DISPLAY-HINT "2d-1d-1d,1d:1d:1d.1d,1a1d:1d"
    STATUS       current
    DESCRIPTION
            "A date-time specification.

            field  octets  contents                  range
            -----  ------  --------                  -----
              1      1-2   year*                     0..65536
              2       3    month                     1..12
              3       4    day                       1..31
              4       5    hour                      0..23
              5       6    minutes                   0..59
              6       7    seconds                   0..60
                           (use 60 for leap-second)
              7       8    deci-seconds              0..9
              8       9    direction from UTC        '+' / '-'
              9      10    hours from UTC*           0..13
             10      11    minutes from UTC          0..59

            * Notes:
            - the value of year is in network-byte order
            - daylight saving time in New Zealand is +13

            For example, Tuesday May 26, 1992 at 1:30:15 PM EDT would be
            displayed as:

                             1992-5-26,13:30:15.0,-4:0

            Note that if only local time is known, then timezone
            information (fields 8-10) is not present."
    SYNTAX       OCTET STRING (SIZE (8 | 11))
```

brian-brazil:
Note that if there's no timezone, we can't properly convert.

RichiH:
That's true, but arguably a configuration error by the end user. If they put wrong or incomplete data in other config, we can not do anything but expose it as such.

brian-brazil:
It's not a configuration error of the user if the device is incapable of outputting timezones.

RichiH:
While that is true, the common case, in my experience with network devices, is that TZ info is optional, but always supported.

SuperQ:
The spec says we can basically fake it if the zone is not returned by the device. We can apply something like `time.Now().Zone()`. Documenting that "Time zones are not always supported" is probably good enough.

brian-brazil:
We'd have to presume UTC.

PR #322, opened 2018-08-14
Add support for RFC 2579 DateAndTime

SuperQ:
Extract a UNIX timestamp from RFC 2579 DateAndTime textual conversions.
* Add support to the generator.
* Add support to the collector.
* Update snmp.yml with new DateAndTime metrics.

Fixes: https://github.com/prometheus/snmp_exporter/issues/321

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"encoding/binary"
 	"fmt"
 	"net"
 	"strconv"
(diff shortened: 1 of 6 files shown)


generator/tree.go
```diff
@@ -130,6 +137,8 @@ func metricType(t string) (string, bool) {
 		return "IpAddr", true
 	case "PhysAddress48", "DisplayString", "Float", "Double":
 		return t, true
+	case "DateAndTime":
```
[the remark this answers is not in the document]
SuperQ: This comes from the definition in the MIB/RFC.
```
DateAndTime ::= TEXTUAL-CONVENTION
```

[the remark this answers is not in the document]
RichiH: Ah, good.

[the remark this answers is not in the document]
brian-brazil: The textual convention is what we use for these, so this is fine.

collector.go
```diff
@@ -234,6 +235,57 @@ func getPduValue(pdu *gosnmp.SnmpPDU) float64 {
...
+		return 0
+	}
+	pduLength := len(v)
+	// DateAndTime can be 8 or 11 bytes depending if the time zone is included.
+	switch pduLength {
+	case 8:
+		// No time zone included, assume UTC.
+		tz = time.UTC
+	case 11:
+		// Extract the timezone from hte last 3 bytes.
```
brian-brazil: the

collector.go
```diff
@@ -234,6 +235,57 @@ func getPduValue(pdu *gosnmp.SnmpPDU) float64 {
...
+			log.Debugf("error parsing location string: %q", locString)
+			return 0
+		}
+		tz = loc.Location()
+	default:
+		log.Debugf("invalid DateAndTime length %v", pduLength)
+		return 0
+	}
+	if err != nil {
+		log.Debugf("unable to parse DateAndTime %q", v)
```
brian-brazil: Can you log the error?

collector.go
```diff
@@ -234,6 +235,57 @@ func getPduValue(pdu *gosnmp.SnmpPDU) float64 {
...
+		tz  *time.Location
+		err error
+	)
+	// DateAndTime should be a slice of bytes.
+	switch pduType := pdu.Value.(type) {
+	case []byte:
+		v = pdu.Value.([]byte)
+	default:
+		log.Debugf("invalid DateAndTime type %v", pduType)
+		return 0
```
brian-brazil: Should we really be returning 0 here?

generator/tree.go
```diff
@@ -115,6 +115,13 @@ func prepareTree(nodes *Node) map[string]*Node {
 		}
 	})
 
+	// Convert RFC 2579 DateAndTime textual conversion to type.
+	walkNode(nodes, func(n *Node) {
```
brian-brazil: This and the previous two walks should really be combined

collector.go
SuperQ: I was debating how to do error handling here. The function that uses this, `pduToSamples`, doesn't really have any error handling.

generator/tree.go
SuperQ: I saw this, and was wondering why we did walkNode over more than once. I can refactor.

collector.go
brian-brazil: We can either fail the scrape, or return nothing for this metric and log. Returning bad data is not a good option imho.

generator/README.md
```diff
@@ -105,6 +105,7 @@ modules:
                              #   gauge:   An integer with type gauge.
                              #   counter: An integer with type counter.
                              #   OctetString: A bit string, rendered as 0xff34.
+                             #   DateAndTime: An RFC 2579 DateAndTime byte sequence.
```
brian-brazil: I'd mention that UTC is presumed if the tz is missing.

brian-brazil:
You need unittests for both the collector and generator changes.

SuperQ:
@brian-brazil Yup, tests were why I left it WIP.

generator/FORMAT.md
```diff
@@ -25,6 +26,10 @@ module_name:
      # as a label value on that gauge.
 
      # A metric that's part of a table, and thus has labels.
```
brian-brazil: This isn't part of a table. You don't need to change this file.

SuperQ: I was thinking it would be worth having an example in the FORMAT file. Don't care either way.

brian-brazil: The format file is more the general structure, the various types are over in the readme.

SuperQ: Got it, removed.

brian-brazil: The types used to be in here, but they got moved when the type override was added. I'd prefer not to having to maintain two lists of types, as it's more likely they'll get out of sync.

PR #329, opened 2018-09-06
Install snmp-mibs-downloader in the generator's Dockerfile

toliger:
There are no MIBs installed by default in the generator Dockerfile. This enables the non-free debian repository and installs the `snmp-mibs-downloader` in the Docker image.


brian-brazil:
Dupe of #258

PR #330, opened 2018-09-07
Inetaddress

### LINKED ISSUE #290: See what we can do about InetAddress
To understand it you need an InetAddressType, which is in another object. Our model is that each object is independent so making that work might not be practical.

brian-brazil:
Fixes #290

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -50,6 +50,14 @@ func oidToList(oid string) []int {
 	return result
 }
 
+func listToOid(l []int) string {
+	var result []string
+	for _, o := range l {
+		result = append(result, strconv.Itoa(o))
+	}
+	return strings.Join(result, ".")
+}
+
 func ScrapeTarget(target string, config *config.Module) ([]gosnmp.SnmpPDU, error) {
 	// Set the options.
 	snmp := gosnmp.GoSNMP{}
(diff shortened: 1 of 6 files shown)


collector.go
```diff
@@ -50,6 +50,17 @@ func oidToList(oid string) []int {
 	return result
 }
 
+func listToOid(l []int) string {
+	result := ""
```
SuperQ: I'm thinking this would be better done with a slice of strings.

```
var result []string
...
result = append(result strconv.Itoa(o))
...
return strings.Join(result, ":")
```

collector.go
```diff
@@ -315,14 +326,40 @@ func pduToSamples(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, o
 		// It's some form of string.
 		t = prometheus.GaugeValue
 		value = 1.0
+		metricType := metric.Type
+
+		// For InetAddress, lookup the type and substitute it in.
+		if metric.Type == "InetAddress" {
```
SuperQ: I think this should be `metricType`, as we just set it above.

collector.go
brian-brazil: Done, much cleaner.

collector.go
brian-brazil: Done

SuperQ:
Looks like some tab/space `go fmt` problems.

brian-brazil:
@SuperQ Could you give this another look? I needed to extend it to allow for Juniper issues.

SuperQ:
It would be useful to add an example of the RFC violation to the generator/snmp configs. But that can be done as a separate PR.

brian-brazil:
I don't think Juniper makes their MIBs available publicly unfortunately.

SuperQ:
Juniper MIBs are here: https://apps.juniper.net/mib-explorer/download.jsp

PR #332, opened 2018-09-12
Add MIB downloader to generator

SuperQ:
Add a Makefile to automatically download all the required MIBs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,23 @@ jobs:
         paths:
         - .build
 
+  generator:
+    docker:
+    - image: circleci/golang:1.10
+    working_directory: /go/src/github.com/prometheus/snmp_exporter/generator
+    resource_class: large
+
+    environment:
+      MIBDIRS: mibs:/usr/share/snmp/mibs
+    steps:
+    - checkout:
+        path: /go/src/github.com/prometheus/snmp_exporter
+    - run: sudo apt-get -y install build-essential diffutils libsnmp-dev
+    - run: go build
+    - run: make mibs
+    - run: ./generator generate
+    - run: diff -u ../snmp.yml snmp.yml
+
   docker_hub_master:
     docker:
     - image: circleci/golang:1.10
@@ -101,10 +118,15 @@ workflows:
         filters:
           tags:
             only: /.*/
+    - generator:
+        filters:
+          tags:
+            only: /.*/
     - docker_hub_master:
         requires:
         - test
         - build
+        - generator
         filters:
           branches:
             only: master
@@ -112,6 +134,7 @@ workflows:
         requires:
         - test
         - build
+        - generator
         filters:
           tags:
             only: /^v[0-9]+(\.[0-9]+){2}(-.+|[^-.]*)$/
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -1,0 +1,115 @@
+# Copyright 2018 The Prometheus Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+MIBDIR := mibs
+
+APC_URL        := 'https://download.schneider-electric.com/files?p_File_Name=powernet426.mib'
+ARISTA_URL     := https://www.arista.com/assets/data/docs/MIBS
+CISCO_URL      := 'ftp://ftp.cisco.com/pub/mibs/v2/v2.tar.gz'
+MIKROTIK_URL   := 'http://download2.mikrotik.com/Mikrotik.mib'
+PALOALTO_URL   := 'https://www.paloaltonetworks.com/content/dam/pan/en_US/assets/zip/technical-documentation/snmp-mib-modules/PAN-MIB-MODULES-7.0.zip'
+SERVERTECH_URL := 'ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3.mib'
+SYNOLOGY_URL   := 'https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip'
+UBNT_DL_URL    := http://dl.ubnt-ut.com/snmp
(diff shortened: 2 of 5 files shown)


generator/Makefile
```diff
@@ -0,0 +1,60 @@
...
+PALOALTO_URL   := 'https://www.paloaltonetworks.com/content/dam/pan/en_US/assets/zip/technical-documentation/snmp-mib-modules/PAN-MIB-MODULES-7.0.zip'
+
+
+.DEFAULT: all
+
+.PHONY: all
+all: mibs
+
+clean:
+	@rm -rv $(MIBDIR)
```
brian-brazil: I don't think it's wise to blow away the user's mibdir like this, it could contain other mibs.

SuperQ: Sounds reasonable.

SuperQ:
FYI, I left out UCD-SNMP-MIB, it's included in `/usr/share/snmp/mibs`.

SuperQ:
I've added a generator build and snmp.yml diff test to CircleCI.

generator/.gitignore
```diff
@@ -0,0 +1 @@
+mibs
```
brian-brazil: Can we put this in the main gitignore?

generator/Makefile
```diff
@@ -0,0 +1,115 @@
...
+	@curl -s -o $(MIBDIR)/ARISTA-SW-IP-FORWARDING-MIB -L "$(ARISTA_URL)/ARISTA-SW-IP-FORWARDING-MIB.txt"
+
+$(MIBDIR)/.cisco_v2:
+	@echo ">> Downloading cisco v2"
+	@curl -s -L $(CISCO_URL) \
+		| tar -C $(MIBDIR) --strip-components=3 -zxv
+	@touch $(MIBDIR)/.cisco_v2
+
+$(MIBDIR)/mikrotik.mib:
+	@echo ">> Downloading cisco v2"
```
brian-brazil: Message is wrong

.circleci/config.yml
```diff
@@ -28,6 +28,25 @@ jobs:
...
+    resource_class: large
+
+    environment:
+      MIBDIRS: mibs:/usr/share/snmp/mibs
+    steps:
+    - checkout:
+        path: /go/src/github.com/prometheus/snmp_exporter
+    - run: sudo sh -c "echo 'deb http://deb.debian.org/debian stable non-free' > /etc/apt/sources.list.d/non-free.list"
+    - run: sudo apt-get update
+    - run: sudo apt-get -y install build-essential diffutils libsnmp-dev snmp-mibs-downloader
```
brian-brazil: What do we need snmp-mibs-downloader for? I'd expect the MIBs to be self-contained.

generator/Makefile
```diff
@@ -0,0 +1,115 @@
...
+$(MIBDIR)/ARISTA-SMI-MIB:
+	@echo ">> Downloading ARISTA-SMI-MIB"
+	@curl -s -o $(MIBDIR)/ARISTA-SMI-MIB -L "$(ARISTA_URL)/ARISTA-SMI-MIB.txt"
+
+$(MIBDIR)/ARISTA-SW-IP-FORWARDING-MIB:
+	@echo ">> Downloading ARISTA-SW-IP-FORWARDING-MIB"
+	@curl -s -o $(MIBDIR)/ARISTA-SW-IP-FORWARDING-MIB -L "$(ARISTA_URL)/ARISTA-SW-IP-FORWARDING-MIB.txt"
+
+$(MIBDIR)/.cisco_v2:
+	@echo ">> Downloading cisco v2"
```
RichiH: `cisco_v2`

generator/Makefile
```diff
@@ -0,0 +1,115 @@
...
+  $(MIBDIR)/UBNT-MIB \
+  $(MIBDIR)/UBNT-UniFi-MIB \
+  $(MIBDIR)/UBNT-AirMAX-MIB.txt
+
+mib-dir:
+	@mkdir -p -v $(MIBDIR)
+
+$(MIBDIR)/apc-powernet-mib:
+	@echo ">> Downloading apc-powernet-mib"
+	@curl -s -o $(MIBDIR)/apc-powernet-mib -L $(APC_URL)
```
RichiH: No check if curl is installed.

generator/Makefile
```diff
@@ -0,0 +1,115 @@
...
+
+$(MIBDIR)/servertech-sentry3-mib:
+	@echo ">> Downloading servertech-sentry3-mib"
+	@curl -s -o $(MIBDIR)/servertech-sentry3-mib -L $(SERVERTECH_URL)
+
+$(MIBDIR)/.synology:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading synology to $(TMP)"
+	@curl -s -o $(TMP) -L $(SYNOLOGY_URL)
+	@unzip -j -d $(MIBDIR) $(TMP)
```
RichiH: No check if unzip is installed.

generator/Makefile
```diff
@@ -0,0 +1,115 @@
...
+	@rm -v $(TMP)
+	@touch $(MIBDIR)/.paloalto_panos
+
+$(MIBDIR)/servertech-sentry3-mib:
+	@echo ">> Downloading servertech-sentry3-mib"
+	@curl -s -o $(MIBDIR)/servertech-sentry3-mib -L $(SERVERTECH_URL)
+
+$(MIBDIR)/.synology:
+	$(eval TMP := $(shell mktemp))
```
RichiH: If that fails, `$TMP` is just empty.

RichiH:
> FYI, I left out UCD-SNMP-MIB, it's included in `/usr/share/snmp/mibs`.

Which mibs dir? I doubt there is a unix-wide standard to rely on.

generator/Makefile
SuperQ: Make should fail if mktemp returns a non-zero code.

generator/Makefile
SuperQ: I consider this out of scope, Make will handle the command failure.

generator/Makefile
SuperQ: Also out of scope, Make handles this.

SuperQ:
@RichiH It turns out we don't need `UCD-SNMP-MIB` at all for the current generator.yml.

RichiH:
Generally speaking, I would like to try and avoid depending on any mib which has not been downloaded.

SuperQ:
@RichiH Yes, previously we used the `snmp-mibs-downloader` package, but it turns out to be unnecessary. As we add new features to the generator.yml, we can add more MIBs to the downloader.

RichiH:
I think we can merge current state.

General comment: MIB names are repeated so often, I wonder if it wouldn't make sense to loop through them.

PR #335, opened 2018-09-18
Added RHEL package list for installation instructions

czhanacek:
Signed-off-by: czhanacek <charlie@czhanacek.com>
@brian-brazil

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -9,6 +9,8 @@ Due to the dynamic dependency on NetSNMP, you must build the generator yourself.
 
 ```
 sudo apt-get install build-essential libsnmp-dev # Debian-based distros
+OR
+sudo yum install gcc gcc-g++ make gcc-go net-snmp net-snmp-utils net-snmp-libs
 go get github.com/prometheus/snmp_exporter/generator
 cd ${GOPATH-$HOME/go}/src/github.com/prometheus/snmp_exporter/generator
 go build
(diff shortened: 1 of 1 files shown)


generator/README.md
```diff
@@ -9,6 +9,8 @@ Due to the dynamic dependency on NetSNMP, you must build the generator yourself.
 
 ```
 sudo apt-get install build-essential libsnmp-dev # Debian-based distros
+OR
```
brian-brazil: This reads like it's a command, use a comment like above.

czhanacek:
Both issues should be fixed in my latest commit

generator/README.md
brian-brazil: There's still the OR here, and no comment saying the other is for RH distros

czhanacek:
Forgot to add RHEL comment, it's in there now.

generator/README.md
```diff
@@ -10,7 +10,7 @@ Due to the dynamic dependency on NetSNMP, you must build the generator yourself.
 ```
 sudo apt-get install build-essential libsnmp-dev # Debian-based distros
 # OR
```
brian-brazil: Can you remove this line?

brian-brazil: The line with the OR

czhanacek: What if we did
# Debian-based distros
apt-get command
# RHEL-based distros
yum command

That way, it doesn't read like both commands should be executed

SuperQ: That sounds reasonable to me.

brian-brazil:
Thanks!

Issue #180, opened 2017-06-23
[Discussion] Appropriate level of normalization in snmp_exporter and the generator


RichiH:
This should serve as the canonical place for this recurring discussion to collect links to related issues and to keep the overall discussion somewhat concise.

brian-brazil:
This should follow the same principles as all other non-distinguishing labels, per the examples at https://www.robustperception.io/how-to-have-labels-for-machine-roles/ and https://www.robustperception.io/exposing-the-software-version-to-prometheus/

For why non-distinguishing labels shouldn't be added to time series, see https://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas/ (replace target with time series, same logic applied).

RichiH:
The obvious counter-point is the same as in all normalization discussions:

Striking the right balance between pristine design and usability.

Having everything in referenced time series quickly descends into a join-fest, which highly complicates promql queries for no end-user benefit.

As per @fabxc , there's no storage penalties either way.

Also, while it's not a Prometheus project, Grafana is the default suggestion for dashboarding and there's no way to handle joins in either templates nor in legends.

As mentioned, this situation has led to people not adapting snmp_exporter or even Prometheus; or to writing their own custom exporters.

All in all, the trade-off seems to be more than valid, to me. I would not object to allowing the snmp_exporter to do both, but this opens the question of what the default should be.

brian-brazil:
The argument you're making is that client libraries should be permitted to allow adding constant labels to a set of exposed metrics for things like version numbers, as that is the same semantic question as what's being discussed here.

We've decided that's not the way we're going there (and we call it out in the exporter guidelines), and I see no reason to make a different decision here.

> Having everything in referenced time series quickly descends into a join-fest, which highly complicates promql queries for no end-user benefit.

That it does not benefit your exact use case does not mean there is no end-user benefit. What you suggest breaks common use cases due to discontinuity of time series, and makes life harder for other use cases due to superfluous labels.

> Grafana is the default suggestion for dashboarding and there's no way to handle joins in either templates nor in legends.

An exact type of tooling support not currently existing is not an argument I accept, as that leads to technical and design debt when that feature comes along. It's also not required, you'd do this in the expression as we do with version numbers (which users appear to be fine with).

I'd suggest filing a bug with Grafana if this is something you want, though I'm unsure how this would work. I think this might require a custom plugin.

[...]

PR #339, opened 2018-09-25
Multi index

### LINKED ISSUE #227: Support generating multi-index lookups
Some OIDs have multiple index values that need to be used to perform lookups.  This is supported by the exporter's config, but not the generator's config.

Example:

```yaml
servertech_sentry3:
  walk:
  - 1.3.6.1.4.1.1718.3.2.2.1 # infeedEntry
  metrics:
  - name: infeedStatus
    oid: 1.3.6.1.4.1.1718.3.2.2.1.5
    type: gauge
    indexes:
    - labelname: towerIndex
      type: gauge
    - labelname: infeedIndex
      type: gauge
    lookups:
    - labels: [towerIndex, infeedIndex]
      labelname: infeedName
      oid: 1.3.6.1.4.1.1718.3.2.2.1.3
      type: DisplayString
```

brian-brazil:
Fixes #227 #180

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -23,6 +23,7 @@ SERVERTECH_URL := 'ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3.mib'
 SYNOLOGY_URL   := 'https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip'
 UBNT_DL_URL    := http://dl.ubnt-ut.com/snmp
 UBNT_AIROS_URL := https://dl.ubnt.com/firmwares/airos-ubnt-mib/ubnt-mib.zip
+UCD_URL        := http://www.net-snmp.org/docs/mibs/UCD-SNMP-MIB.txt
 
 
 .DEFAULT: all
@@ -50,7 +51,8 @@ mibs: mib-dir \
   $(MIBDIR)/.synology \
   $(MIBDIR)/UBNT-MIB \
   $(MIBDIR)/UBNT-UniFi-MIB \
-  $(MIBDIR)/UBNT-AirMAX-MIB.txt
+  $(MIBDIR)/UBNT-AirMAX-MIB.txt \
+  $(MIBDIR)/UCD-SNMP-MIB.txt
 
 mib-dir:
 	@mkdir -p -v $(MIBDIR)
@@ -119,3 +121,7 @@ $(MIBDIR)/UBNT-AirMAX-MIB.txt:
 	@curl -s -o $(TMP) -L $(UBNT_AIROS_URL)
 	@unzip -j -d $(MIBDIR) $(TMP) UBNT-AirMAX-MIB.txt
 	@rm -v $(TMP)
+
+$(MIBDIR)/UCD-SNMP-MIB.txt:
+	@echo ">> Downloading UCD-SNMP-MIB.txt"
+	@curl -s -o $(MIBDIR)/UCD-SNMP-MIB.txt -L "$(UCD_URL)"
--- a/generator/README.md
+++ b/generator/README.md
@@ -86,15 +86,16 @@ modules:
                             # Required if context is configured on the device.
 
     lookups:  # Optional list of lookups to perform.
-              # If keep_old is the default of false, this must only be used when the new index is unique.
+              # The default for `keep_source_indexes` is false. Indexes must be unique for this option to be used.
 
       # If the index of a table is bsnDot11EssIndex, usually that'd be the label
       # on the resulting metrics from that table. Instead, use the index to
       # lookup the bsnDot11EssSsid table entry and create a bsnDot11EssSsid label
       # with that value.
-      - old_indexes: [bsnDot11EssIndex]
-        new_index: bsnDot11EssSsid
-        keep_old: false  # If true, don't delete old index label for this lookup.
+      - source_indexes: [bsnDot11EssIndex]
+        lookup: bsnDot11EssSsid
+        keep_source_indexes: false  # If true, don't delete source index labels for this lookup.
+                                    # This is useful when the new index has duplicates.
 
      overrides: # Allows for per-module overrides of bits of MIBs
        metricName:
--- a/generator/README.md
+++ b/generator/README.md
@@ -86,7 +86,7 @@ modules:
                             # Required if context is configured on the device.
 
     lookups:  # Optional list of lookups to perform.
-              # This must only be used when the new index is unique.
+              # If keep_old is the default of false, this must only be used when the new index is unique.
 
       # If the index of a table is bsnDot11EssIndex, usually that'd be the label
       # on the resulting metrics from that table. Instead, use the index to
@@ -94,6 +94,7 @@ modules:
       # with that value.
       - old_indexes: [bsnDot11EssIndex]
         new_index: bsnDot11EssSsid
+        keep_old: false  # If true, don't delete old index label for this lookup.
 
      overrides: # Allows for per-module overrides of bits of MIBs
        metricName:
--- a/generator/README.md
+++ b/generator/README.md
@@ -94,8 +94,8 @@ modules:
       # with that value.
       - source_indexes: [bsnDot11EssIndex]
         lookup: bsnDot11EssSsid
-        keep_source_indexes: false  # If true, don't delete source index labels for this lookup.
-                                    # This is useful when the new index has duplicates.
+        drop_source_indexes: false  # If true, delete source index labels for this lookup.
+                                    # This avoids label clutter when the new index is unique.
 
      overrides: # Allows for per-module overrides of bits of MIBs
        metricName:
--- a/generator/README.md
+++ b/generator/README.md
@@ -92,7 +92,7 @@ modules:
       # on the resulting metrics from that table. Instead, use the index to
       # lookup the bsnDot11EssSsid table entry and create a bsnDot11EssSsid label
       # with that value.
-      - old_index: bsnDot11EssIndex
+      - old_indexes: [bsnDot11EssIndex]
         new_index: bsnDot11EssSsid
 
      overrides: # Allows for per-module overrides of bits of MIBs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
       working_dir: ~/project/generator
 
     environment:
-      MIBDIRS: mibs:/usr/share/snmp/mibs
+      MIBDIRS: mibs
     steps:
     - checkout:
         path: ~/project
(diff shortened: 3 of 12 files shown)


RichiH:
Wait, what, w00t!

generator/README.md
```diff
@@ -82,14 +82,15 @@ modules:
                             # Required if context is configured on the device.
 
     lookups:  # Optional list of lookups to perform.
-              # This must only be used when the new index is unique.
+              # If keep_old is the default of false, this must only be used when the new index is unique.
```
SuperQ: Should `keep_old` be `keep_sources_indexes`?

generator/README.md
```diff
@@ -82,14 +82,15 @@ modules:
...
 
       # If the index of a table is bsnDot11EssIndex, usually that'd be the label
       # on the resulting metrics from that table. Instead, use the index to
       # lookup the bsnDot11EssSsid table entry and create a bsnDot11EssSsid label
       # with that value.
-      - old_index: bsnDot11EssIndex
-        new_index: bsnDot11EssSsid
+      - source_indexes: [bsnDot11EssIndex]
+        lookup: bsnDot11EssSsid
+        keep_sources_indexes: false  # If true, don't delete source index labels for this lookup.
```
SuperQ: Typo, `keep_source_indexes`.

brian-brazil: Fixed

generator/README.md
brian-brazil: Done

generator/README.md
```diff
@@ -82,14 +82,17 @@ modules:
                             # Required if context is configured on the device.
 
     lookups:  # Optional list of lookups to perform.
-              # This must only be used when the new index is unique.
+              # If keep_source_indexes is the default of false, this must only
```
SuperQ: Hrm, this is now more awkward.

```
The default for `keep_source_indexes` is false. Indexes must be unique for this option to be used.
```

I also think we should make the default be `true`, in order to provide safer behavior.

SuperQ: @RichiH Do you have any wording suggestions here? Also what do you think about the default value?

brian-brazil: Changed. The default value should be false, as adding additional labels should only be done on a case-by-case basis - not blindly.

generator/README.md
```diff
@@ -82,14 +82,15 @@ modules:
...
-              # This must only be used when the new index is unique.
+              # If keep_old is the default of false, this must only be used when the new index is unique.
 
       # If the index of a table is bsnDot11EssIndex, usually that'd be the label
       # on the resulting metrics from that table. Instead, use the index to
       # lookup the bsnDot11EssSsid table entry and create a bsnDot11EssSsid label
       # with that value.
       - old_indexes: [bsnDot11EssIndex]
         new_index: bsnDot11EssSsid
+        keep_old: false  # If true, don't delete old index label for this lookup.
```
RichiH: I think this should default to true. I tried running without the indices and it's a pain pit of duplicate label sets.

brian-brazil: When that happens the setting can be changed by the user, we shouldn't be adding labels blindly.

SuperQ: We must add labels, rather than assume the new index is unique. The only true unique key for SNMP provides is the index. In order to provide end user safe defaults, this needs to be `true.`

brian-brazil: The default is not to have any lookups, which is perfectly safe.

RichiH:
The default in this option should be safe as well.

brian-brazil:
The default discourages users from having surplus labels, and it was agreed previously that use of such a thing should only be on a case-by-case basis - which you are proposing is not in line with.

dannyk81:
Apologies for barging in 😄 

But I came across this, while looking for a way to add both the `ifDescr` and `ifAlias` labels to the various `if*` metrics from Network devices, currently we are replacing `ifIndex` with `ifDescr`, but I would also like to have the ifAlias as a label (indeed we solve this today with group modifiers, but that really complicates things for use when querying).

This change looks interesting, but I don't think it's going to provide a solution for this particular use case, right?

SuperQ:
@dannyk81 Yes! This will do exactly that. We're currently arguing over the final detail about having add or remove be the default behavior.

RichiH:
FYI, we are currently deadlocked on the last open question; before that's resolved, I don't think we can merge this PR.

* @brian-brazil strongly believes ifIndex should be dropped by default
* I strongly believe ifIndex must be kept by default

dannyk81:
Ohhh! @SuperQ great news! I guess I misunderstood how this works - apologies for the noise and hope to see this soon 🙏

Bu to be honest, I'm still not quite sure how the lookup should be defined with this to achieve it (wanted to take this branch for a spin):

```yaml
    lookups:
    - source_indexes: [ifIndex, ifAlias]
      lookup: ifDescr
      keep_source_indexes: true
```
?

The result I'm looking for:
```
ifHCInOctets{ifIndex="<index>",ifDescr="EthernetX/Y",ifAlias="# <some text here> #",instance="<some cisco switch>",job="cisconxos-snmp-exporter"}
```

My 2 cents on your debate (though I don't presume it will have too much weight on your final decision 😅) is that the `ifIndex` should be kept around, since for me it provides a unique ID of the resource the metric represents and removing it by default seems not the best approach.

Having said that, I'm perfectly fine tuning this parameter to `true` myself if it means we get this thing merged and released 👍 


I agree with @brian-brazil that `snmp.yml` should not be generated/tweaked manually and we as users try to abide by the rule/best-practice, however the reality is that we have cases where this is unavoidable, my example is the CheckPoint FW devices.

```
  - name: fwDropPcktsOut
    oid: 1.3.6.1.4.1.2620.1.1.25.5.1.10
    type: counter
    help: ' - 1.3.6.1.4.1.2620.1.1.25.5.1.10'
    indexes:
    - labelname: fwIfIndex
      type: gauge
    - labelname: fwIfDummyIndex
      type: gauge
    lookups:
    - labels:
      - fwIfIndex
      - fwIfDummyIndex
      labelname: fwIfName
      oid: 1.3.6.1.4.1.2620.1.1.25.5.1.2
      type: DisplayString
```

We modify the generated snmp.yml as described above, since the results returned from the device are incorrect and not supported by the generator (the solution was provided by Brian and Ben --> https://groups.google.com/forum/#!msg/prometheus-users/NDyO8qRlXoM/mUculDc-BQAJ)

I believe this is not a unique case, and SNMP is full of such examples, so the point is that we need to find a balance between keeping things clean and standardized on the one hand and on the other hand provide the users (us) with tools to handle the abnormal cases.

/end of rant ☮️

SuperQ:
For `ifIndex` you need a configuration like this when using this patch:

```yaml
    lookups:
    - source_indexes: [ifIndex]
      lookup: ifDescr
      keep_source_indexes: true
    - source_indexes: [ifIndex]
      lookup: ifAlias
      keep_source_indexes: true
    - source_indexes: [ifIndex]
      lookup: ifName
      keep_source_indexes: true
```

The explanation here is that the "source index" is the index on the MIB entry.

You can see that in the MIB:
```
ifEntry OBJECT-TYPE
    SYNTAX      IfEntry
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION
            "An entry containing management information applicable to a
            particular interface."
    INDEX   { ifIndex }
    ::= { ifTable 1 }
```

The big reason for this change is we wanted to be able to support objects that have multi-dimension indexes. For example, here's one from a wifi controller:

```
   wlsxWlanAPBssidEntry OBJECT-TYPE
      SYNTAX       WlanAPBssidEntry
      MAX-ACCESS   not-accessible       
      STATUS       current
      DESCRIPTION
             "BSSID Entry"
      INDEX {wlanAPMacAddress, wlanAPRadioNumber, wlanAPBSSID}
      ::= { wlsxWlanAPBssidTable 1 }
```

dannyk81:
Ahhh, I see! @SuperQ, that's clear now. so the `ifIndex` lookup can be specified several times (each with a different lookup), I wasn't sure about that.

Thanks, this is clear now.

dannyk81:
Hope to see this soon 🙏

SuperQ:
@fjaeckel We're stuck on agreeing on the default of `keep_source_indexes`.

fjaeckel:
@SuperQ as long as we're documenting the default, whatever it may be, and its implications - we should be able to just roll with it. I think for myself `ifIndex`, if thats the only label, should be kept by default.

SuperQ:
That's the issue. If you use a lookup, for `ifDescr`, `ifIndex` is dropped with the current default of `false`.

fjaeckel:
> That's the issue. If you use a lookup, for ifDescr, ifIndex is dropped with the current default of false.

I don't see it as a blocker, it can be overridden and should just be explained why some think its a good default.
I see a bigger issue in not merging this PR than setting a sane default, people maintain their own patch sets for snmp_exporter (and generator) - I for one do.

dannyk81:
@SuperQ @RichiH @brian-brazil 

Sorry for @ing like this, but is there any chance we can have closure on this? I desperately need this functionality and I've been avoiding using a private build of this PR until now... but seems like we're at an impasse 😞

I completely agree with @fjaeckel, as long as we have the option to change the default behavior, it should be a reasonable approach for everyone.

Please 🙏

SuperQ:
Can you please rebase this against master?

brian-brazil:
This is updated. Do we know which if any of the existing lookup examples are non-unique?

SuperQ:
It's not the lookups that cause issues, it's the devices when you use them. I know HP devices will generate duplicate `ifDescr="HP ProCurve Switch software loopback interface"` labels when you have management IPs configured on more than one vlan. The only way to have `ifDescr` is by having both `ifIndex` and `ifDescr`.

brian-brazil:
ifInterfaces I've already covered, it's the others I'm wondering about.

generator/Makefile
```diff
@@ -50,7 +51,8 @@ mibs: mib-dir \
   $(MIBDIR)/.synology \
   $(MIBDIR)/UBNT-MIB \
   $(MIBDIR)/UBNT-UniFi-MIB \
-  $(MIBDIR)/UBNT-AirMAX-MIB.txt
+  $(MIBDIR)/UBNT-AirMAX-MIB.txt \
+	$(MIBDIR)/UCD-SNMP-MIB.txt
```
SuperQ: Tab vs space issue here.

SuperQ:
I think that should cover it for now.

generator/Makefile
brian-brazil: Fixed.

SuperQ:
Looks like there's a minor difference to some metrics in the snmp.yml.

brian-brazil:
Odd, yet another run fixed it.

RichiH:
LGTM. Ideally, I would love to have a release right after merging this.

brian-brazil:
Yes, this requires an immediate release due to the nature of the changes.

RichiH:
Just to make sure: @brian-brazil is this ready to merge from your end?

brian-brazil:
Yes, I'll do it when I'm ready to release.

Issue #345, opened 2018-10-17
fatal error: net-snmp/net-snmp-config.h: No such file or directory


ykfq:
I encountered the error below when generating `snmp.yml`
  ```
  go/src/github.com/prometheus/snmp_exporter/generator/net_snmp.go:19:38: fatal error: net-snmp/net-snmp-config.h: No such file or directory
   #include <net-snmp/net-snmp-config.h>
  ```
  
It is because I need to install the `net-snmp-devel` module, so the Building step in the README of generator should be like this:
  ```
  sudo yum install gcc gcc-g++ make net-snmp net-snmp-utils net-snmp-libs net-snmp-devel
  ```

wbh1:
I experienced the same issue today on RHEL7 and just submitted a PR for it.

RichiH:
Fixed with https://github.com/prometheus/snmp_exporter/pull/347

PR #347, opened 2018-10-25
Add net-snmp-devel as a dependency for building the generator on RHEL

wbh1:
cc: @SuperQ 

This is a real small fix to the generator README to address issue #345, which I also ran into today (RHEL 7.5)


RichiH:
Thanks!

Issue #350, opened 2018-10-31
Option to mask counters to 53 bits?


candlerb:
Prometheus stores float64 (double) values, which have a [53-bit mantissa](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).

For counters which exceed 2<sup>53</sup>, this means there is a loss of precision when taking the difference between adjacent samples to calculate rates.

A link running at 100Gbps could reach 2<sup>53</sup> bytes in 8.3 days.  After this, the accuracy will degrade, eventually to the point where counters are rounded to the nearest 2<sup>11</sup> bytes.

This is arguably a minor issue - but it could be avoided entirely if there were an option to truncate SNMP counter values to 53 bits by ANDing them with `0x001f_ffff_ffff_ffff`.  Such counters would then wrap more frequently, but maintain full accuracy.

Such masking must not be applied to gauges of course, so it would have to be a per-OID option.

SuperQ:
Thanks for opening this issue. I've talked with people about doing this, I think it's a good idea. The simple method would be to do `value % 2^53`.

The question is, should we do this automatically for counters? Should it be a configuration option for the generator? Should it be per-OID or per module?

I could imagine something like this:
```yaml
overrides:
  ifHCInOctets:
    wrap_large_values: true
  ifHCOutOctets:
    wrap_large_values: true
```

SuperQ:
Another question, for users that care about the current value of the counter, how/if do we expose the number of wraps applied to the value? Do we have a metric like `ifHCInOctets_wraps{...} 0`

RichiH:
I thought this was done already with counter resets. I know Brian and me
talked about this after the Prometheus dev summit 2016 in Berlin.


Sent by mobile; please excuse my brevity.

RichiH:
LGTM.

RichiH:
Maybe with one or two sentences in the docs to make this obvious.

PR #351, opened 2018-11-02
Add automatic Conter64 wrapping

SuperQ:
Automatically wrap Counter64 PDU values every 2^53 to avoid float64
precision loss.

Add a command line flag to enable/disable this feature, enable by
default.

Closes: https://github.com/prometheus/snmp_exporter/issues/350

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -24,6 +24,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"github.com/soniah/gosnmp"
+	"gopkg.in/alecthomas/kingpin.v2"
 
 	"github.com/prometheus/snmp_exporter/config"
 )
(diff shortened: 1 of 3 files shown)


collector.go
```diff
@@ -28,6 +28,9 @@ import (
 	"github.com/prometheus/snmp_exporter/config"
 )
 
+// 64-bit float mantissa: https://en.wikipedia.org/wiki/Double-precision_floating-point_format
+const float64Mantissa = 9007199254740992
```
squeed: Careful, go constants are "special" and try and do all kinds of [magic](https://blog.golang.org/constants#TOC_10.). Better to use an explicitly typed variable here.

SuperQ: Good point, thanks!

main.go
```diff
@@ -37,6 +37,7 @@ import (
 var (
 	configFile    = kingpin.Flag("config.file", "Path to configuration file.").Default("snmp.yml").String()
 	listenAddress = kingpin.Flag("web.listen-address", "Address to listen on for web interface and telemetry.").Default(":9116").String()
+	wrapCounters  = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
```
brian-brazil: this should be snmp. or similar

SuperQ: Good idea.

SuperQ: Also, I guess this should go in collector.go.

RichiH:
I would state at what value you wrap. Else, lgtm given the other comments.

PR #360, opened 2018-11-27
Fix codespell issues.

mjtrangoni:
Hi @brian-brazil @SuperQ ,

I fixed the misspellings issues reported by [codespell](https://github.com/codespell-project/codespell).

What is about adding this to CircleCI too? There is only one word to be excluded `uinit` and one file `snmp.yml` (because of the upstream mibs misspellings issues).
It should be something like `$ codespell -S './vendor*,./.git*' -L uint -x snmp.yml`.

Signed-off-by: Mario Trangoni <mjtrangoni@gmail.com>

### DIFF
--- a/generator/net_snmp.go
+++ b/generator/net_snmp.go
@@ -126,7 +126,7 @@ var (
 	}
 )
 
-// Initialise NetSNMP. Returns MIB parse errors.
+// Initialize NetSNMP. Returns MIB parse errors.
 //
 // Warning: This function plays with the stderr file descriptor.
 func initSNMP() string {
--- a/generator/net_snmp.go
+++ b/generator/net_snmp.go
@@ -126,7 +126,7 @@ var (
 	}
 )
 
-// Initilise NetSNMP. Returns MIB parse errors.
+// Initialise NetSNMP. Returns MIB parse errors.
 //
 // Warning: This function plays with the stderr file descriptor.
 func initSNMP() string {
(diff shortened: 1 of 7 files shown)


SuperQ:
We use [codespell in circleci](https://github.com/prometheus/node_exporter/blob/master/.circleci/config.yml#L16-L23) for other projects. I have no objection to adding it here. I've been intending to add it upstream in the main Prometheus repo.

SuperQ:
My only problem with codespell is it's a bit too sensitive on Go code with short variable names, and the uint thing bothers me.

mjtrangoni:
@SuperQ  Yes, you are right. As always there is some exceptions, and that is why I didn't add it directly.

generator/net_snmp.go
```diff
@@ -126,7 +126,7 @@ var (
 	}
 )
 
-// Initilise NetSNMP. Returns MIB parse errors.
+// Initialise NetSNMP. Returns MIB parse errors.
```
brian-brazil: Didn't we say we were going with US spellings?

mjtrangoni: @brian-brazil +1
I can go for it, and replace every `Initialise` with  `Initialize`. Should I?

brian-brazil: Yes please

mjtrangoni:
@brian-brazil PTAL. `snmp.yml` has to remain as it is.

PR #376, opened 2019-01-28
Update generator readme.md to include full make

RaceFPV:
Centos 7 has an issue with the generator parsing the default MIBs when using just 'make mibs'. Instead running the full 'make' command inside generator resolves the unmet dependency and allows the generator to run correctly. Without running full make the generator will always fail on generating MIBs, even when more are added in later as one of the defaults is incorrectly functioning.


RaceFPV:
updated with signoff

brian-brazil:
If there's an issue with the Makefile, I'd prefer to fix the Makefile.

brian-brazil:
This has gone stale, and there's not a solid confirmation that there's an actual issue here.

PR #378, opened 2019-02-07
 Add EnumAsInfo and EnumAsStateSet type overrides.

brian-brazil:
@SuperQ

### DIFF
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -163,6 +163,8 @@ func metricType(t string) (string, bool) {
 		return t, true
 	case "DateAndTime":
 		return t, true
+	case "EnumAsInfo", "EnumAsStateSet":
+		return t, true
 	default:
 		// Unsupported type.
 		return "", false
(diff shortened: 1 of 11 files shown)


SuperQ:
Nice.

In tree.go, maybe we should only populate the final metric `EnumValues` if `Type` is not "gauge". This way it avoid filling the snmp.yml with enums that don't get used.

generator/tree.go
```diff
@@ -322,12 +324,13 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
...
-				Help:    n.Description + " - " + n.Oid,
-				Indexes: []*config.Index{},
-				Lookups: []*config.Lookup{},
+				Name:       sanitizeLabelName(n.Label),
+				Oid:        n.Oid,
+				Type:       t,
+				Help:       n.Description + " - " + n.Oid,
+				Indexes:    []*config.Index{},
+				Lookups:    []*config.Lookup{},
+				EnumValues: n.EnumValues,
```
SuperQ:
```go
e := make(map[int]string)
if t != "gauge" {
  e = n.EnumValues
}
...
  EnumValues: e
...
```

brian-brazil:
> In tree.go, maybe we should only populate the final metric EnumValues if Type is not "gauge". This way it avoid filling the snmp.yml with enums that don't get used.

I actually did that on purpose, I figure we may want to e.g. put it in the HELP for enums or something in the future and this avoids having to do a new generator run.

SuperQ:
Interesting idea. Could possibly be huge, like ifType is.

brian-brazil:
> Interesting idea. Could possibly be huge, like ifType is.

Yeah, I also had that concern. That's the only that big though.

Issue #375, opened 2019-01-23
ADP PDU only gets interface info


qrpike:
### What device/snmpwalk OID are you using?
I am using an ADP PDU AP86XX

I can get the network interface information, however I cannot get the power usage and any other stats other than net inferface details. 

I found this article: https://www.apc.com/us/en/faqs/FA156179/ however, I'm not sure how to configure the generator. Any help would be greatly appreciated.

Thanks,

daenney:
Look in `generator.yml` for `apcups`. There's a `walk` block there with names and OIDs to walk. As you can see the `1.3.6.1.4.1.318.1.1.26.x.y.z....` etc listed in the article you linked aren't covered, so none of the OIDs for the Rack PDU 2G projects are getting fetched.

While you're at it, also update the `Makefile` to fetch `428` of the APC MIB would be nice.

PR #382, opened 2019-02-14
Add rPDU2BankStatusTable to generator example.

SuperQ:
Also updates with latest keepalived MIB.

Closes: https://github.com/prometheus/snmp_exporter/issues/375
Closes: https://github.com/prometheus/snmp_exporter/issues/383

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -11,20 +11,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-MIBDIR := mibs
+MIBDIR   := mibs
+MIB_PATH := 'mibs'
 
 APC_URL        := 'https://download.schneider-electric.com/files?p_File_Name=powernet426.mib'
 ARISTA_URL     := https://www.arista.com/assets/data/docs/MIBS
 CISCO_URL      := 'ftp://ftp.cisco.com/pub/mibs/v2/v2.tar.gz'
+FRAMEWORK_URL  := http://www.net-snmp.org/docs/mibs/SNMP-FRAMEWORK-MIB.txt
 KEEPALIVED_URL := 'https://raw.githubusercontent.com/acassen/keepalived/master/doc/KEEPALIVED-MIB.txt'
 MIKROTIK_URL   := 'http://download2.mikrotik.com/Mikrotik.mib'
+NET_SNMP_URL   := http://www.net-snmp.org/docs/mibs/NET-SNMP-MIB.txt
+NET_TC_URL     := http://www.net-snmp.org/docs/mibs/NET-SNMP-TC.txt
 PALOALTO_URL   := 'https://www.paloaltonetworks.com/content/dam/pan/en_US/assets/zip/technical-documentation/snmp-mib-modules/PAN-MIB-MODULES-7.0.zip'
 SERVERTECH_URL := 'ftp://ftp.servertech.com/Pub/SNMP/sentry3/Sentry3.mib'
 SYNOLOGY_URL   := 'https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip'
 UBNT_DL_URL    := http://dl.ubnt-ut.com/snmp
 UBNT_AIROS_URL := https://dl.ubnt.com/firmwares/airos-ubnt-mib/ubnt-mib.zip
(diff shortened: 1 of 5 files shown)


brian-brazil:
The generator test is failing.

SuperQ:
I get different number of parse errors on my machine. Very strange.

brian-brazil:
Did you do MIBDIRS=mibs?

SuperQ:
Yes, I did, but the whole mibs dir was a mess. I've re-done the way we fetch Cisco's broken set of MIBs by only copying what we need out of it.

SuperQ:
Ok, now the generator parse_errors runs without any errors.

generator/Makefile
```diff
@@ -73,19 +89,45 @@ $(MIBDIR)/ARISTA-SW-IP-FORWARDING-MIB:
...
-
-$(MIBDIR)/KEEPALIVED-MIB.txt:
-	@echo ">> Downloading keepalived"
-	@curl -s -o $(MIBDIR)/KEEPALIVED-MIB.txt -L $(KEEPALIVED_URL)
-
-$(MIBDIR)/mikrotik.mib:
-	@echo ">> Downloading mikrotik"
-	@curl -s -o $(MIBDIR)/mikrotik.mib -L $(MIKROTIK_URL)
+		| tar -C $(MIBDIR)/cisco_v2 --strip-components=3 -zxv
+	cp mibs/cisco_v2/AIRESPACE-REF-MIB.my mibs/AIRESPACE-REF-MIB
```
brian-brazil: A for loop would be a bit clearer here I think

generator/Makefile
```diff
@@ -34,18 +37,31 @@ all: mibs
...
 		$(MIBDIR)/.synology
 
+generator:
+	go build
+
+generate: generator mibs
+	MIBDIRS=$(MIB_PATH) ./generator generate
+
+parse_errors: generator mibs
+	MIBDIRS=$(MIB_PATH) ./generator parse_errors
```
brian-brazil: You know this will exit 0 even if there are parse errors?

generator/Makefile
SuperQ: Maybe, I wanted to just keep it simple.

generator/Makefile
SuperQ: Yup, it's just a shortcut for manual testing and visibility at this point. I was going to open an issue to add a "fail on errors" flag to the generator so we can include it as an actual part of the tests.

SuperQ: I added the flag to allow failure code on parse errors.

generator/main.go
```diff
@@ -98,10 +99,15 @@ func main() {
 	command := kingpin.Parse()
 
 	parseErrors := initSNMP()
+	parseErrorCount := len(parseErrors)
 
-	if len(parseErrors) != 0 {
+	if parseErrorCount != 0 {
 		log.Warnf("NetSNMP reported %d parse errors", len(strings.Split(parseErrors, "\n")))
 	}
+	if !*failOnParseErrors {
```
brian-brazil: I'd expect the process to exit here

generator/main.go
```diff
@@ -124,4 +130,5 @@ func main() {
 				n.Oid, n.Label, t, n.TextualConvention, n.Hint, n.Indexes, implied, n.EnumValues, n.Description)
 		})
 	}
+	os.Exit(parseErrorCount)
```
brian-brazil: This is unsafe, what if there's 256 parse errors?

generator/main.go
```diff
@@ -86,6 +86,7 @@ func generateConfig(nodes *Node, nameToNode map[string]*Node) {
 }
 
 var (
+	failOnParseErrors  = kingpin.Flag("fail-on-parse-errors", "Exit with a non-zero status if there are parse errors").Default("false").Bool()
```
brian-brazil: I'd put MIB in here somewhere, as it could be generator.yml you're talking about.

generator/main.go
SuperQ: I thought about that, but then it exits without printing the actual parse errors.

generator/main.go
SuperQ: Ahh, good point. I'll fix.

generator/main.go
SuperQ: Ok.

PR #385, opened 2019-02-16
Update generator Dockerfile

SuperQ:
* Use a single RUN to build as a single layer.
* Bake MIBs into the image.
* Setup MIBDIRS ENV.

### DIFF
--- a/generator/Dockerfile
+++ b/generator/Dockerfile
@@ -1,17 +1,16 @@
 FROM golang:latest
 
-RUN apt-get update && apt-get install -y libsnmp-dev
-
-RUN go get github.com/prometheus/snmp_exporter/generator
-RUN cd /go/src/github.com/prometheus/snmp_exporter/generator \
+RUN apt-get update && \
+    apt-get install -y libsnmp-dev && \
+    go get github.com/prometheus/snmp_exporter/generator && \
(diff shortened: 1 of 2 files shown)


daenney:
Haha, I just pushed #386 which has a change to the `docker run` example instead.

brian-brazil:
:+1:

generator/Dockerfile
```diff
@@ -1,17 +1,19 @@
...
+    go install && \
+    cp generator.yml /opt/ && \
+    make mibs && \
+    mv mibs /opt/
 
 WORKDIR "/opt"
 
 ENTRYPOINT ["/go/bin/generator"]
 
+ENV MIBDIRS mibs
```
brian-brazil: Actually, this values doesn't match the docs. This should work off the user's mibs, not the ones we use.

SuperQ: I've update the README to make things easier and match our build.

SuperQ:
I've also removed the `make mibs` part of the Dockerfile.

What do you think about removing the `generator.yml` from the Dockerfile as well?

brian-brazil:
It's make sense on consistency grounds.

SuperQ:
Done, PTAL.

brian-brazil:
:+1:

PR #386, opened 2019-02-16
generator: Update README.md

### LINKED ISSUE #384: EnumAsInfo vs EnumAsStateSet
I'm a bit confused as to what the difference between the two is. I know what the README says but I'm not sure what that means. Is there some heuristic for when to pick one over the other? I'm currently revising #379 and a bunch of fields are enum-like.

daenney:
* Explain the difference between EnumAsInfo and EnumAsStateSet

Closes #384

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -129,6 +129,23 @@ modules:
                              #   EnumAsStateSet: An enum with a time series per state. Good for variable low-cardinality enums.
 ```
 
+### EnumAsInfo and EnumAsStateSet
+
+SNMP contains the concept of integer indexed enumerations (enums). There are two ways
+to represent these strings in Prometheus. They can be "info" metrics, or they can be
+"state sets". SNMP does not specify which should be used, and it's up to the use case
+of the data. Some users may also prefer the raw integer value, rather than the string.
+
+In order to set enum integer to string mapping, you must use one of the two overrides.
+
+`EnumAsInfo` should be used for properties that provide inventory-like data. For example
+a device type, the name of a colour etc. It is important that this value is constant.
+
+`EnumAsStateSet` should be used for things that represent state or that you might want
+to alert on. For example the link state, is it up or down, is it in an error state,
+whether a panel is open or closed etc. Please be careful to not use this for high
+cardinality values as it will generate 1 time series per possible value.
+
 ## Where to get MIBs
 
 Some of these are quite sluggish, so use wget to download.


generator/README.md
```diff
@@ -43,6 +43,11 @@ docker run -ti \
   snmp-generator generate
 ```
 
+Please note: In CI we don't run the generator with the MIBs in `/usr/share/snmp`
```
SuperQ: Yea, I think we should use https://github.com/prometheus/snmp_exporter/pull/385 to make it work by default.

daenney: Aye, I'll remove that.

generator/README.md
```diff
@@ -127,6 +127,17 @@ modules:
                              #   EnumAsStateSet: An enum with a time series per state. Good for variable low-cardinality enums.
 ```
 
+### EnumAsInfo and EnumAsStateSet
+
+Sometimes it'll be necessary to specify an override for a metric because they are
```
SuperQ: I would prefer this sentence:

```
SNMP contains the concept of integer indexed enumerations (Enums). There are two ways to represent these strings in Prometheus. They can be "info" metrics, or they can be "state sets". SNMP does not specify which should be used, and it's up to the use case of the data. Some users may also prefer the raw integer value, rather than the string.

In order to set enum integer to string mapping, you must use one of the two overrides.
```

generator/README.md
```diff
@@ -127,6 +127,17 @@ modules:
...
+enum-like, i.e they provide a fixed number of possible options for a value.
+
+The difference between `EnumAsInfo` and `EnumAsStateSet` is that the first one should
```
SuperQ: Let's split this into two paragraphs to avoid mixing information about each type. This will be less confusing for users to parse.

generator/README.md
```diff
@@ -127,6 +127,22 @@ modules:
...
+### EnumAsInfo and EnumAsStateSet
+
+SNMP contains the concept of integer indexed enumerations (Enums). There are two ways
+to represent these strings in Prometheus. They can be "info" metrics, or they can be
+"state sets". SNMP does not specify which should be used, and it's up to the use case
+of the data. Some users may also prefer the raw integer value, rather than the string.
+
+In order to set enum integer to string mapping, you must use one of the two overrides.
+
+`EnumAsInfo` should be used for properties that provide inventory-like data. For example
```
brian-brazil: This should mention that it should be constant

generator/README.md
```diff
@@ -127,6 +127,22 @@ modules:
...
+a device type, the name of a colour etc.
+
+`EnumAsStateSet` should be used for things that represent state or that you might want
```
brian-brazil: I'd add a cardinality warning here

generator/README.md
daenney: Updated.

generator/README.md
daenney: Done.

generator/README.md
```diff
@@ -127,6 +127,23 @@ modules:
                              #   EnumAsStateSet: An enum with a time series per state. Good for variable low-cardinality enums.
 ```
 
+### EnumAsInfo and EnumAsStateSet
+
+SNMP contains the concept of integer indexed enumerations (Enums). There are two ways
```
brian-brazil: s/Enum/enum/, it's not a proper noun

PR #379, opened 2019-02-11
Add support for Printer MIB v2

daenney:
This adds support for PRINTER-MIB-v2 as defined in RFC 3805. It allows for the monitoring of printers including their supply levels (toners/ink cartridges etc).

Signed-off-by: Daniele Sluijters <daenney@users.noreply.github.com>

PR #387, opened 2019-02-16
generator: Get ifType MIB from IANA

daenney:
We used to use the ifType MIB that ships with the Cisco package. However that one is outdated and has at least one type in it. Instead, this now uses the MIB from the authoritative source, IANA.

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -19,6 +19,7 @@ ARISTA_URL       := https://www.arista.com/assets/data/docs/MIBS
 CISCO_URL        := 'ftp://ftp.cisco.com/pub/mibs/v2/v2.tar.gz'
 FRAMEWORK_URL    := http://www.net-snmp.org/docs/mibs/SNMP-FRAMEWORK-MIB.txt
 IANA_CHARSET_URL := https://www.iana.org/assignments/ianacharset-mib/ianacharset-mib
+IANA_IFTYPE_URL  := https://www.iana.org/assignments/ianaiftype-mib/ianaiftype-mib
 IANA_PRINTER_URL := https://www.iana.org/assignments/ianaprinter-mib/ianaprinter-mib
 KEEPALIVED_URL   := 'https://raw.githubusercontent.com/acassen/keepalived/master/doc/KEEPALIVED-MIB.txt'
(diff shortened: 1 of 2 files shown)


daenney:
This is a follow-up from #379 where the problem with the Cisco version of ifType MIB was noticed.

SuperQ:
So many ifTypes!

generator/Makefile
```diff
@@ -118,6 +119,10 @@ $(MIBDIR)/IANA-CHARSET-MIB.txt:
 	@echo ">> Donwloading IANA charset MIB"
 	@curl -s -o $(MIBDIR)/IANA-CHARSET-MIB.txt -L $(IANA_CHARSET_URL)
 
+$(MIBDIR)/IANA-IFTYPE-MIB.txt:
+	@echo ">> Donwloading IANA ifType MIB"
```
brian-brazil: Why we're not using the cisco version deserves a comment

SuperQ: I would say we should document the opposite, why we are using Cisco MIBs that are not written by Cisco. :)

generator/Makefile
```diff
@@ -118,6 +119,14 @@ $(MIBDIR)/IANA-CHARSET-MIB.txt:
 	@echo ">> Donwloading IANA charset MIB"
 	@curl -s -o $(MIBDIR)/IANA-CHARSET-MIB.txt -L $(IANA_CHARSET_URL)
 
+$(MIBDIR)/IANA-IFTYPE-MIB.txt:
+# We donwload the ifType MIB from IANA instead of using the one that comes
```
SuperQ: IMO this comment is unnecessary, as this is the official source. We only used the one in the Cisco tree out of laziness/convenience.

daenney: Ya, I think that makes more sense. I'll wait until you and @brian-brazil reach consensus and update accordingly.

generator/Makefile
brian-brazil: True, though this isn't a Cisco mib.

generator/Makefile
brian-brazil: Go with SuperQ here.

PR #388, opened 2019-02-16
generator: Upgrade APC to 4.2.8

daenney:


snmp.yml
```diff
@@ -1118,6 +1141,66 @@ apcups:
...
+  - name: upsHighPrecBatteryPackOnlyTemperature
+    oid: 1.3.6.1.4.1.318.1.1.1.2.3.10.4.1.4
+    type: gauge
+    help: The battery pack temperature measured in 10ths of degree Celcius - 1.3.6.1.4.1.318.1.1.1.2.3.10.4.1.4
+    indexes:
+    - labelname: upsHighPrecBatteryPackOnlyIndex
+      type: gauge
+  - name: upsHighPrecBatteryPackOnlyStatus
+    oid: 1.3.6.1.4.1.318.1.1.1.2.3.10.4.1.5
+    type: OctetString
```
daenney: Wondering whether this should be overridden to `DisplayString`

snmp.yml
```diff
@@ -1118,6 +1141,66 @@ apcups:
...
+    help: The battery status for the pack only - 1.3.6.1.4.1.318.1.1.1.2.3.10.4.1.5
+    indexes:
+    - labelname: upsHighPrecBatteryPackOnlyIndex
+      type: gauge
+  - name: upsHighPrecBatteryPackOnlyHealth
+    oid: 1.3.6.1.4.1.318.1.1.1.2.3.10.4.1.6
+    type: OctetString
```
daenney: Wondering whether this should be overridden to `DisplayString`

snmp.yml
brian-brazil: Is it always an ASCII/UTF-8 string?

daenney: Mmm, according to the MIB docs:

```
The battery status for the pack only.
bit 0 Disconnected
bit 1 Overvoltage
bit 2 NeedsReplacement
bit 3 OvertemperatureCritical
bit 4 Charger
bit 5 TemperatureSensor
bit 6 BusSoftStart
bit 7 OvertemperatureWarning
```

snmp.yml
daenney: According to the MIB docs:

```
The battery pack health.
bit 0 Battery lifetime okay
bit 1 Battery lifetime near end, order replacement pack
bit 2 Battery lifetime exceeded, replace battery
bit 3 Battery lifetime near end acknowledged, order replacement pack
bit 4 Battery lifetime exceeded acknowledged, replace battery
bit 5 Battery measured lifetime near end, order replacement pack
bit 6 Battery measured lifetime near end acknowledged, order replacement pack
```

snmp.yml
SuperQ: That seems like it returns a single byte and you have to look at the bits yourself.

daenney: Yeah. I guess so. So we leave it as such?

SuperQ:
`upsAdvBatteryChargerStatus` would make a good StateSet.

snmp.yml
SuperQ: Yea, I would leave it, it doesn't seem like a string, and without a device to test with it's hard to say.

daenney:
Yap. Added.

snmp.yml
brian-brazil: You might be able to do something with regex extracts.

PR #389, opened 2019-02-16
printer: Add prtMarkerSuppliesType label

daenney:
Without `prtMarkerSuppliesType` there's no way to know to what kind of
supply (toner, vs paper etc) the `prtMarkerSuppliesLevel` and `prtMarkerSuppliesMaxCapacity`
refer to which makes collecting them pretty useless.

I can't add an override to `EnumAsInfo` on this one since then the
lookup won't work.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -309,6 +309,10 @@ modules:
       - prtCoverStatus
       - prtMarkerSuppliesLevel
       - prtMarkerSuppliesMaxCapacity
+      - prtMarkerSuppliesType
+    lookups:
+      - source_indexes: [hrDeviceIndex, prtMarkerSuppliesIndex]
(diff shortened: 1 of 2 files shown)


generator/generator.yml
```diff
@@ -309,6 +309,11 @@ modules:
       - prtCoverStatus
       - prtMarkerSuppliesLevel
       - prtMarkerSuppliesMaxCapacity
+      - prtMarkerSuppliesType
+    lookups:
+      - source_indexes: [hrDeviceIndex, prtMarkerSuppliesIndex]
+        lookup: prtMarkerSuppliesType
+        drop_source_indexes: false
```
brian-brazil: No need to mention defaults

PR #406, opened 2019-04-25
add external label support for each metric

ShingYang:
add external label support for each metric. 
We now move from zabbix to promethues, we need cutomize some label key/value for each metric, though it can be completed by metric_relabel_configs of prometheus, it's complicated and I'm afraid promethues cannot handle them when many tagets there.

external labels for each metric gives an option for snmp exporter user to customize their own label key values.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -637,6 +637,9 @@ func indexesToLabels(indexOids []int, metric *config.Metric, oidToPdu map[string
 	labels := map[string]string{}
 	labelOids := map[string][]int{}
 
+	for k, v := range metric.ExternalLabels {
+		labels[k] = v
+	}
 	// Covert indexes to useful strings.
 	for _, index := range metric.Indexes {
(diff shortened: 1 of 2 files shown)


brian-brazil:
What exactly are you trying to do?

ShingYang:
add external for each metric.
Eg:
```
  metrics:
  - external_labels:
      app: Power
      name: Power Supply {#SNMPINDEX} Status
    help: app:Power name:Power Supply {#SNMPINDEX} Status  oid:1.3.6.1.4.1.674.10892.5.4.600.12.1.5.1
    indexes:
    - labelname: PowerSuppliesIndex
      type: gauge
    name: PowerSupplyStatus
    oid: 1.3.6.1.4.1.674.10892.5.4.600.12.1.5.1
    type: gauge
```

brian-brazil:
We generally only accept things that are determinable from a MIB, and such custom labels are not from a MIB. I'd suggest doing without these, as these seem to be presentation-layer issues rather than fundamentally interesting labels.

ShingYang:
understood, thanks.

PR #413, opened 2019-05-17
Updating Ubuntu deps to match all requirements to actually build

shairozan:
Signed-off-by: DJ Breeden <dbreeden@thinq.com>

Was trying to build a docker image for doing the snmp generation and ran into all kinds of issues with Ubuntu. Updated the readme to include all the packages I had to use in order to get it to build from an empty container.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -9,7 +9,7 @@ Due to the dynamic dependency on NetSNMP, you must build the generator yourself.
 
 ```
 # Debian-based distributions.
-sudo apt-get install unzip build-essential libsnmp-dev snmp libsnmp-base snmp-mibs-downloader golang git curl # Debian-based distros
+sudo apt-get install unzip build-essential libsnmp-dev snmp libsnmp-base golang git curl # Debian-based distros
 # Redhat-based distributions.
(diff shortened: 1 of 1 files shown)


generator/README.md
```diff
@@ -9,7 +9,7 @@ Due to the dynamic dependency on NetSNMP, you must build the generator yourself.
 
 ```
 # Debian-based distributions.
-sudo apt-get install unzip build-essential libsnmp-dev # Debian-based distros
+sudo apt-get install unzip build-essential libsnmp-dev snmp libsnmp-base snmp-mibs-downloader golang git curl # Debian-based distros
```
brian-brazil: The mibs downloader is not required.

shairozan: Updated it to remove it. Waiting on the build to queue / finish.

SuperQ:
I would generally recommend against using the deb packages for `golang`. Ubuntu does not update them, and the available version in Ubuntu (1.10) is too old as we have switched to Go modules which requires >= 1.11.

SuperQ:
If you look at the generator Dockerfile, we do not use Ubuntu, but the official `golang` image that is still Debian-based.

Issue #414, opened 2019-05-19
Drop metrics that are looked up in generator?


candlerb:
### Host operating system: output of `uname -a`

```
Linux prometheus 4.15.0-43-generic #46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

### snmp_exporter version: output of `snmp_exporter --version`

```
snmp_exporter, version 0.15.0 (branch: HEAD, revision: 92a3da4467f8bc6759cf197e7442b8c43e890b13)
  build user:       root@8c3a7c03d455
  build date:       20190212-13:40:02
  go version:       go1.11.5
```

### What device/snmpwalk OID are you using?

`if_mib` from default supplied `snmp.yml`

### What did you expect to see?

A single timeseries declaring each binding of (ifAlias, ifDescr, ifName, ifIndex), apart from the interface metrics themselves.

### What did you see instead?

There are three separate metrics which as far as I can see are identical:

```
ifAlias{ifAlias="",ifDescr="bridge",ifIndex="14",ifName="bridge"} 1
ifAlias{ifAlias="",ifDescr="ether1",ifIndex="16",ifName="ether1"} 1
ifAlias{ifAlias="",ifDescr="ether2",ifIndex="2",ifName="ether2"} 1
ifAlias{ifAlias="",ifDescr="ether3",ifIndex="3",ifName="ether3"} 1
ifAlias{ifAlias="",ifDescr="ether4",ifIndex="4",ifName="ether4"} 1
ifAlias{ifAlias="",ifDescr="ether5",ifIndex="5",ifName="ether5"} 1
ifAlias{ifAlias="",ifDescr="pppoe-out1",ifIndex="17",ifName="pppoe-out1"} 1
ifAlias{ifAlias="",ifDescr="sfp1",ifIndex="6",ifName="sfp1"} 1
ifAlias{ifAlias="",ifDescr="vlan254",ifIndex="10",ifName="vlan254"} 1
ifAlias{ifAlias="",ifDescr="vlan255",ifIndex="9",ifName="vlan255"} 1
ifAlias{ifAlias="Hurricane Electric IPv6 Tunnel Broker",ifDescr="sit1",ifIndex="15",ifName="sit1"} 1
...
ifDescr{ifAlias="",ifDescr="bridge",ifIndex="14",ifName="bridge"} 1
[...]

PR #415, opened 2019-05-19
Drop metrics for label lookups

SuperQ:
Remove metrics in if_mib that are part of the lookups.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -11,6 +11,12 @@ modules:
         # Use OID to avoid conflict with Netscaler NS-ROOT-MIB.
         lookup: 1.3.6.1.2.1.31.1.1.1.1 # ifName
     overrides:
+      ifAlias:
+        ignore: true # Lookup metric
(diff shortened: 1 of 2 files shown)


SuperQ:
Partial fix for https://github.com/prometheus/snmp_exporter/issues/414

CC @candlerb

SuperQ:
`ifIndex` is also a bit useless, possibly worth dropping.

candlerb:
I have tested the new generated `snmp.yml`, and I confirm this drops the ifName/ifDescr/ifAlias metrics as desired - thank you!

I don't have a particular use for ifIndex.  It *does* give a static mapping between names and indexes which you can depend on independently of any particular metric (e.g. if you needed to join on it).  But since every metric is already labelled with ifIndex, ifName, ifDescr and ifAlias, I don't think it's needed in practice.

brian-brazil:
We should leave ifIndex, some other tables tend to have the value and users may need to do == on it.

PR #437, opened 2019-07-21
Fix default value in generator documentation

erdnaxe:
In generator, auth_protocol defaults to MD5 rather than SHA.
This can be confusing for people trying to collect data with this exporter for the first time.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -78,7 +78,7 @@ modules:
                                     # Can be noAuthNoPriv, authNoPriv or authPriv.
       password: pass  # Has no default. Also known as authKey, -A option to NetSNMP.
                       # Required if security_level is authNoPriv or authPriv.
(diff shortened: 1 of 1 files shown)


generator/README.md
```diff
@@ -78,7 +78,7 @@ modules:
                                     # Can be noAuthNoPriv, authNoPriv or authPriv.
       password: pass  # Has no default. Also known as authKey, -A option to NetSNMP.
                       # Required if security_level is authNoPriv or authPriv.
-      auth_protocol: SHA  # MD5 or SHA, defaults to SHA. -a option to NetSNMP.
+      auth_protocol: SHA  # MD5 or SHA, defaults to MD5. -a option to NetSNMP.
```
brian-brazil: It should say MD5 on the left too.

erdnaxe: As this is a small edit, I just amended the last commit.

brian-brazil:
Thanks!

PR #438, opened 2019-07-24
Add support for SNMP community string as a parameter

visago:
The whole idea is to allow configurations of target/community/module from prometheus instead of snmp_exporter's snmp.yml. Currently you can only specify target/module in snmp_exporter. This adds the community string (So its only for v1/v2) 

I've updated the README.md with an example of having a list of targets via a JSON file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,8 @@ releases](https://github.com/prometheus/snmp_exporter/releases) page.
 
 Visit http://localhost:9116/snmp?target=1.2.3.4 where 1.2.3.4 is the IP of the
 SNMP device to get metrics from. You can also specify a `module` parameter, to
-choose which module to use from the config file.
+choose which module to use from the config file. You can also specify a
+'community' paramter, to override the community string from the config file.
 
 ## Configuration
 
@@ -47,21 +48,48 @@ scrape_configs:
     static_configs:
       - targets:
         - 192.168.1.2  # SNMP device.
+    file_sd_configs:
+      - files:
+        - '/data/*.json'
     metrics_path: /snmp
     params:
       module: [if_mib]
+      community: [public]
     relabel_configs:
       - source_labels: [__address__]
         target_label: __param_target
       - source_labels: [__param_target]
         target_label: instance
+      - source_labels: [community]
+        target_label: __param_community
+      - source_labels: [module]
+        target_label: __param_module
       - target_label: __address__
         replacement: 127.0.0.1:9116  # The SNMP exporter's real hostname:port.
+```
+
+Optionally, provide a json file as a file based service discovery. Only
+community and module are needed. Other labels are optional for snmp_exporter
+purposes.
+
+```JSON
+[
+  {
+    "targets": [ "192.168.1.2" ],
+    "labels": {
+      "name": "router",
+      "community": "publicCommunity",
+      "module": "ddwrt"
+    }
+  }
+]
 ```
 
 This setup allows Prometheus to provide scheduling and service discovery, as
 unlike all other exporters running an exporter on the machine from which we are
 getting the metrics from is not possible.
+
+
 
 ## Large counter value handling
 
--- a/README.md
+++ b/README.md
@@ -20,8 +20,9 @@ releases](https://github.com/prometheus/snmp_exporter/releases) page.
 
 Visit http://localhost:9116/snmp?target=1.2.3.4 where 1.2.3.4 is the IP of the
 SNMP device to get metrics from. You can also specify a `module` parameter, to
-choose which module to use from the config file. You can also specify a
-'community' paramter, to override the community string from the config file.
+choose which module to use from the config file. You can also specify an
+(base64 encoded AES-GCM encrypted) 'community' paramter, to override the 
+community string from the config file.
 
 ## Configuration
 
@@ -69,8 +70,8 @@ scrape_configs:
 ```
 
 Optionally, provide a json file as a file based service discovery. Only
-community and module are needed. Other labels are optional for snmp_exporter
-purposes.
+(base64 encoded AES-GCM encrypted) community and module are needed. Other 
+labels are optional for snmp_exporter purposes.
 
 ```JSON
 [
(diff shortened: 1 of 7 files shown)


brian-brazil:
URL parameters are not considered secure in Prometheus, and in addition you can already do this via multiple modules.

visago:
> URL parameters are not considered secure in Prometheus, and in addition you can already do this via multiple modules.

That explains why this feature wasn't around there !  How about if used with HTTPS terminated exporters ?

Agreed that this is doable via multiple modules, but it means we have to do service discovery on the exporters too.

brian-brazil:
> That explains why this feature wasn't around there ! How about if used with HTTPS terminated exporters ?

It'd still be exposed on the Prometheus UI.

> Agreed that this is doable via multiple modules, but it means we have to do service discovery on the exporters too.

That's a configuration management issue. I'd suggest having fewer passwords.

CRCinAU:
I just stumbled across this because we use an SNMP community other than 'public' - so nothing seems to work.

Is the only way to get this working by creating a new snmp.yml file? There seems to be no way to set the community to anything other than public :\

visago:
> I just stumbled across this because we use an SNMP community other than 'public' - so nothing seems to work.
> 
> Is the only way to get this working by creating a new snmp.yml file? There seems to be no way documented to set the community to anything other than public :\
> 
> Considering the snmp community to be 'secure' is probably a mistake - as its transmitted plain over the network, and also available multiple other places.

Yes, the only 'secure' way is to have it hidden in the snmp.yml file. This way each object you don't see the community string in the URL or prometheus target list (as a label)

My PR adds in community as a parameter (i.e. http://localhost:9116/snmp?target=1.2.3.4&community=superpublic) but is considered insecure (Esp if its a non-https end point) The good thing is you can then add your targets on the prometheus side instead of snmp.yml.

CRCinAU:
SNMP communities aren't secure at all - its transmitted in plain text over the network, and is in no way, shape or form considered a secret - lets not confuse where its stored in a config file or parameter as a security feature.

I totally agree that the community should be able to be set from prometheus via the parameters. It's a good functional improvement that shouldn't be handwaved over for some misguided attempt at security.

brian-brazil:
Secrets are secrets, no matter how they're transmitted. This is the stance in terms of general configuration management anyway, it'd be the same response for any other setting.

CRCinAU:
For clarity, a snmp community is not a secret.

It seems to be a different stance than every other NMS in existence which has a standard config set of IP, Port, Version, community.

cleonello:
This would be useful for real world management of SNMP devices.  

@CRCinAU is 100% correct.  the SNMP community is a name grouping agents and applications together, it is not a "secret".  It does, as a side effect of being a user defined string, provide the most trivial form of authentication.  However, devices are free to treat all SNMP messages as authentic messages and almost always do. 

This feature would make it easier for us to use Prometheus and do our jobs without impacting the complete absence of security in SNMP.  Please consider implementing.

xaque208:
I think you can get around this by just creating a new group in the generator config with the community string you want for the given group.  Name it something useful, and then include it with the module on the Prometheus side configuration.

brian-brazil:
> I think you can get around this by just creating a new group in the generator config with the community string you want for the given group. Name it something useful, and then include it with the module on the Prometheus side configuration.

That's the expected way to handle devices with different configurations.

CRCinAU:
Its not a convenient way to do things though... Changing the community via prometheus.yml is a much better way to do things rather than probably duplicating mass parts of SNMP.yml just to change a community.

xaque208:
If you authenticate to devices using separate credentials, then perhaps their differences warrant separate modules.  Also the config modules in generator.yaml still allow you to expand the portions of the OID tree to walk with the `walk` field, which at least helps reduce the duplication of the config.

cleonello:
> 
> 
> > I think you can get around this by just creating a new group in the generator config with the community string you want for the given group. Name it something useful, and then include it with the module on the Prometheus side configuration.
> 
> That's the expected way to handle devices with different configurations.

Yes, and this hi how I do things currently but it isn't the most appropriate way in my opinion.


> 
> 
> If you authenticate to devices using separate credentials, then perhaps their differences warrant separate modules. Also the config modules in generator.yaml still allow you to expand the portions of the OID tree to walk with the `walk` field, which at least helps reduce the duplication of the config.

The reason this is not desirable is that the community string is an attribute of the target device, not an attribute of the metric (oid).  It does not belong in the generator.yml/snmp.yml describing the oids, it belongs in the json file describing the target devices.  It is valid to have different community strings for each and every device.  It is also valid that these strings may change periodically based on some outside business practice.  These changes can be easily picked up in the json file through the file_sd_config.

cleonello:
> > > I think you can get around this by just creating a new group in the generator config with the community string you want for the given group. Name it something useful, and then include it with the module on the Prometheus side configuration.
> > 
> > 
> > That's the expected way to handle devices with different configurations.
> 
> Yes, and this hi how I do things currently but it isn't the most appropriate way in my opinion.

Not to walk back my comment but to clarify, the existing configuration method works fine for many use cases.  For those not wanting to expose the community string through http it is beneficial.  However, the feature being proposed by @visago provides flexibility which, in my opinion, is needed for the reasons stated in my previous comment.

visago:
Thank you all for the comments. Apologies for the slow update on this. Its good to see that others have the same need as me (ability to handle discovery and management via prometheus.yml) 

I've added AES-GCM encryption for the community string. There's a folder with a simple cryptotool to encrypt (or decrypt to verify) your community string with a passphrase. The passphrase should be passed to the snmp_exporter via commandline (Either direct or preferably via a file containing the passphrase)

I hope this satisfy the security requirements.

brian-brazil:
> I've added AES-GCM encryption for the community string. 

We don't invent security stuff, any secrets should be in plain text for us. It's the user's responsibility to decode it from whatever they're using.

> This is the stance in terms of general configuration management anyway, it'd be the same response for any other setting.

This still applies.

cameronkerrnz:
With all this mention of community-strings, please remember that method used needs to be generalisable to SNMPv3 as well, which _does_ have a notion of secret credentials.

brian-brazil:
As indicated above, this change will not be accepted as it is not in line with either our configuration approach or security policies.

PR #447, opened 2019-09-03
Switch logging to go-kit

brian-brazil:
@SuperQ


generator/tree.go
```diff
@@ -42,7 +44,7 @@ func walkNode(n *Node, f func(n *Node)) {
 }
 
 // Transform the tree.
-func prepareTree(nodes *Node) map[string]*Node {
+func prepareTree(nodes *Node, logger log.Logger) (map[string]*Node, error) {
```
SuperQ: I don't see where this `error` return is non-nil in this function.

brian-brazil: Line 148

SuperQ: Line 148 is a `nil` return.

brian-brazil: Sorry, I'm not awake yet. Fixed.

brian-brazil:
Did you have a chance to look at this again?

collector_test.go
```diff
@@ -534,9 +534,8 @@ func TestGetPduValue(t *testing.T) {
 }
 
 func TestGetPduLargeValue(t *testing.T) {
-	// Setup default flags and suppress logging.
-	log.AddFlags(kingpin.CommandLine)
-	_, err := kingpin.CommandLine.Parse([]string{"--log.level", "fatal"})
+	// Setup default.
```
SuperQ: Maybe just drop the comment.

generator/tree_test.go
```diff
@@ -1881,8 +1882,8 @@ func TestGenerateConfigModule(t *testing.T) {
 			}
 		}
 
-		nameToNode := prepareTree(c.node)
-		got := generateConfigModule(c.cfg, c.node, nameToNode)
+		nameToNode := prepareTree(c.node, log.NewNopLogger())
+		got, _ := generateConfigModule(c.cfg, c.node, nameToNode, log.NewNopLogger())
```
SuperQ: Are you sure we should ignore the error here? Previously these were `Fatal` exits.

brian-brazil: The got will be empty, so it'll be caught that way

collector_test.go
brian-brazil: Removed

generator/tree_test.go
SuperQ: But it means that the Fatal errors printed previously will be lost as they're not saved and printed.

Maybe instead of returning the error in tree.go, we do this:
```go
level.Error(logger)...
return nil
```

brian-brazil: I've updated the code, PTAL

PR #452, opened 2019-09-23
Fixed extra wrapping of regexp

arch1tect0r:
I have device which return some metrics as multiline string
```
D1 : 5973 R
D2 : 6053 R
...
```
I wrote regexp which allows to me extract numeric value using override:
```
overrides:
   myStrangeString:
     regex_extracts:
       D1:
         - regex: D1 . (\d*) R
           value: $1
```
In generated snmp.yml I found that my expression was wraped:
```
    regex_extracts:
      D1:
      - value: $1
        regex: ^(?:D1 . (\d*) R)$
```
I tested this expression with my device and got nothing.
After that i tried to see debug info and found extra wrapping:
```
level=debug ts=2019-09-23T13:27:17.223Z caller=collector.go:410 module=mydevice target=192.168.250.54 msg="No match found for regexp" metric=myStrangeString value="D1 : 5973 R\nD2 : 6053 R\n" regex="^(?:^(?:D1 . (\\d*) R')$)$"
```
After fixing the code I got working collector and got expected result from regexp.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -257,7 +257,7 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal(&s); err != nil {
 		return err
 	}
-	regex, err := regexp.Compile("^(?:" + s + ")$")
+	regex, err := regexp.Compile(s)
 	if err != nil {
 		return err
 	}


brian-brazil:
The wrapping is meant to be a noop, if it's not then that's what needs fixing.

arch1tect0r:
It's not noop. Every time when reading snmp.yml it's added extra wrapping for every regexp. It's wrong because it's unexpected logic.

brian-brazil:
The extra wrapping should have no effect though.

arch1tect0r:
I prepare sample when wrapping is bad.

https://play.golang.org/p/MVO9pNNZy96

brian-brazil:
That's working as I'd expect, the regex is being fully anchored.

arch1tect0r:
So, it's normal that any regexp from generator.yml will be wraped twice in `^(?:)$`?

Could you explain me such unexpected behavior?

I prepared sample above because it's real answer which can't be parsed by current version of exporter. Double wrapping or single wrapping brakes regexp logic. 
I don't understand wraping of regexp during loading of snmp.yml. As for me it's unexpected behavior, because it's change logic of original regexp.

brian-brazil:
> So, it's normal that any regexp from generator.yml will be wraped twice in ^(?:)$?

Yes. It's not perfect, but it's not a problem afaik.

> Double wrapping or single wrapping brakes regexp logic.

Our regexes are documented as being fully anchored, and this is how we implement that.

arch1tect0r:
> > So, it's normal that any regexp from generator.yml will be wraped twice in ^(?:)$?
> 
> Yes. It's not perfect, but it's not a problem afaik.
> 
> > Double wrapping or single wrapping brakes regexp logic.
> 
> Our regexes are documented as being fully anchored, and this is how we implement that.

Ok. Could you implement regexp for my situation with double anchores https://play.golang.org/p/1W_AsxxiFS8 ?

SuperQ:
Because you have a multi-line string, you need to ignore the second half of the output.

```
regex: 'D1 . (\d+) R.*'
```

arch1tect0r:
> Because you have a multi-line string, you need to ignore the second half of the output.
> 
> ```
> regex: 'D1 . (\d+) R.*'
> ```

It's not working because exporter adds anchores. I tried to show this in the examples above.

SuperQ:
Yes, you're correct, there's no way to get this to work when the OID contains a newline.

@brian-brazil I think we need to not apply anchoring in SNMP, like we do elsewhere in Prometheus.

SuperQ:
I would extend this no-anchor policy to the generator code as well. We just can't trust SNMP data enough to make assumptions that anchoring is going to work.

arch1tect0r:
> The issue here is matching a newline. You can use `(?s)` or `(\n|.)*`.
It's not about matching new line in common. It's more about that values can't be extracted by regexp because in code exists logic which adds unexpected anchores. 

I know that it's enouth to use such regexp `D1 . (\d+) R.*`, but exporter changed it to `^(?:^(?: D1 . (\d*) R.*)$)$`.

SuperQ:
It works if you're explicit about the newline matching:

```
regex 'D1 . (\d*) R\n.*\n'
```

arch1tect0r:
> It works if you're explicit about the newline matching:
> 
> ```
> regex: 'D1 . (\d*) R\n.*\n'
> ```

But it's work only in current case. If text will have non constant amount of lines it wil not work.

brian-brazil:
This is a usage question, so no need to change anything.

PR #456, opened 2019-09-29
collector: move out of main package

SaveTheRbtz:
Moving collector out of main package allows to use `snmp_exporter` as a library in other programs.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package collector
 
 import (
 	"encoding/binary"
@@ -198,6 +198,10 @@ type collector struct {
 	target string
 	module *config.Module
 	logger log.Logger
+}
+
+func New(target string, module *config.Module, logger log.Logger) *collector {
+	return &collector{target: target, module: module, logger: logger}
 }
 
 // Describe implements Prometheus.Collector.
--- a/collector/collector_test.go
+++ b/collector/collector_test.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package collector
 
 import (
 	"errors"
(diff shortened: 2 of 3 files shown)


brian-brazil:
How do you plan on using this?

SaveTheRbtz:
Currently, to run this in production we need to subprocess the collector, scrape it, and re-emit metrics into our monitoring system, this will simplify this so that we can write our custom `main.go`  and import `github.com/prometheus/snmp_exporter/collector` from it.

brian-brazil:
Most of the main monitoring systems out there support scraping the Prometheus metrics format directly, and running it in-process would be quite odd. Which monitoring system is this?

SaveTheRbtz:
This is an in-house monitoring used at @Dropbox.  It also supports scraping for Prometheus metrics but this adds couple of levels of indirection to the system, which is a bit cumbersome to maintain:
internal configs -> snmp_exporter configs -> prom_metrics -> internal metrics.

This also spans multiple processes (address spaces) which adds additional overhead for operations.

brian-brazil:
Running exporters as separate processes is pretty normal, I'm not sure trying to get all exporters into one process is wise.

SaveTheRbtz:
The question for this diff is mostly how do you want other people to customize snmp_exporter's behaviour if they need to.  Do you want them 1) to be able to use it as a library, 2) do you want them to fork it, or 3) do you want them to wrap it with a separate binary that will generate the config and scrape the results?  Our current approach is 3), but we would really like to move to 1).

brian-brazil:
It's intended to be used as a binary, if there's features we're missing we can talk about adding them.

SuperQ:
I think it would be a good thing to move the functionality to a collector package, we do this across several of our exporters, for example node_exporter and mysqld_exporter.

brian-brazil:
In those cases there's large numbers of collectors, and users wanted to augment them iirc. That's not the case here, there's only one collector.

SuperQ:
Yes, but the abstraction of collectors as a Go package still makes sense.

SaveTheRbtz:
Closing this, but sadly this would likely result in @dropbox forking snmp_exporter.

PR #457, opened 2019-09-30
generator: add --compressed to curl options

SaveTheRbtz:
Plain text MIBs' download speed will benefit from `Accept-Encoding: deflate, gzip`.  This also workarounds currently broken Synology CDN since they return 3xx redirect.

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -14,7 +14,7 @@
 MIBDIR   := mibs
 MIB_PATH := 'mibs'
 
-CURL_OPTS ?= -s --retry 3 --retry-delay 3
+CURL_OPTS ?= -s --retry 3 --retry-delay 3 --compressed --location
 
 DOCKER_IMAGE_NAME ?= snmp-generator
 DOCKER_IMAGE_TAG  ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))


brian-brazil:
Hmm, CI isn't liking that. Let's retry.

SuperQ:
Synology is 404'ing

SaveTheRbtz:
Hm.. Seems like it depends on the geographic location of the build box (CloudFlare has a lot of PoPs and they don't seem to share the cache.)
Does generator build for you locally on master?  Otherwise synology needs to be disabled or we need to find a different URL for it.

SuperQ:
I did some quick digging, it looks like Synology deleted the MIB zip file. I'm not able to find a new location for it either. :frowning:

SaveTheRbtz:
Yeah, the original purpose of this diff was to workaround this with `--compressed` (because of `vary: Accept-Encoding`) since it is still in CF's cache, but seems like this is not true for all of their PoPs:
```
𝛌 ~ curl -s -I https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip | head -1
HTTP/2 404
```
```
𝛌 ~ curl -s --compressed -I https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip | head -1
HTTP/2 200
```

SuperQ:
It looks like Synology updated the URL to a 301. Can you add `--location` to the flags?

brian-brazil:
Thanks!

PR #465, opened 2019-12-03
Support BITS values.

### LINKED ISSUE #443: Handle BITS textual conversion
The [ServerTech Sentry 4](https://cdn10.servertech.com/assets/documents/documents/815/original/Sentry4.mib) uses a textual conversion of a `BITS` type. This looks like it's used as an enum of sort.

For example:
```
    EventNotificationMethods ::= TEXTUAL-CONVENTION
        STATUS      current
        DESCRIPTION
            "Bits to enable event notification methods."
        SYNTAX      BITS {
                        snmpTrap(0),     -- enables notifications by SNMP traps
                        email(1)         -- enables notifications by Email
                    }

    st4UnitNotifications OBJECT-TYPE
        SYNTAX      EventNotificationMethods
        MAX-ACCESS  read-write
        STATUS      current
        DESCRIPTION
            "The notification methods enabled for unit events."
        ::= { st4UnitEventConfigEntry 1 }
```

Another `BITS` example is status enum:

```
    st4UnitCapabilities OBJECT-TYPE
        SYNTAX      BITS {
                        dc(0),         -- unit is DC
                        phase3(1),     -- unit is 3 phase
                        wye(2),        -- unit has Wye circuits
                        delta(3)       -- unit has Delta circuits
                    }
        MAX-ACCESS  read-only
        STATUS      current
        DESCRIPTION
            "The capabilities of the unit."
        ::= { st4UnitConfigEntry 8 }
```

It looks like part of the problem is with `BITS`, we turn them into OctetStrings by default, and they turn into invalid UTF8.

```yaml
modules:
  servertech_sentry4:
    walk:
      - st4UnitCapabilities
```

```yaml
servertech_sentry4:
  walk:
  - 1.3.6.1.4.1.1718.4.1.2.2.1.8
  metrics:
  - name: st4UnitCapabilities
    oid: 1.3.6.1.4.1.1718.4.1.2.2.1.8
    type: OctetString
    help: The capabilities of the unit. - 1.3.6.1.4.1.1718.4.1.2.2.1.8
    indexes:
    - labelname: st4UnitIndex
      type: gauge
    enum_values:
      0: dc
      1: phase3
      2: wye
      3: delta
```

brian-brazil:
Fixes #443

@SuperQ 

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -353,6 +353,8 @@ func pduToSamples(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, o
 		return enumAsInfo(metric, int(value), labelnames, labelvalues)
 	case "EnumAsStateSet":
 		return enumAsStateSet(metric, int(value), labelnames, labelvalues)
+	case "Bits":
+		return bits(metric, pdu.Value, labelnames, labelvalues)
 	default:
 		// It's some form of string.
 		t = prometheus.GaugeValue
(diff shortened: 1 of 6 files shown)


SuperQ:
Wouldn't this be better as a state set, with the various BITS entries as labels, rather than separate metrics?

brian-brazil:
We don't know that they're related, and some of them are just single values (e.g. cIeee8021PaeCapV12R0233SXHPCat6K).

SuperQ:
We do know that they're related by OID. I don't see the single value as an argument either. The enum could change in the future, adding additional label values.

brian-brazil:
It could merely be someone trying to put a pile of unrelated bits into one place for efficiency, and there's more than one that's a singleton and I suspect unlikely to be extended.

SuperQ:
Sure, but the most useful case is a bitset of labels for a single OID, aka a single metric.

brian-brazil:
> is a bitset of labels for a single OID

You mean object, not OID.

> the most useful case

That doesn't mean most common, or what makes sense given the balance of all actual uses.

StateSet was created as a way to handle enums, and this isn't an enum, and not all uses are states so I'm wary of using it.

SuperQ:
The examples we already have in the generator are state sets. Here's some other random examples from MIBs I found.

```
CLSecEncryptType ::= TEXTUAL-CONVENTION
        STATUS  current
        DESCRIPTION
                "This textual convention defines the type of
                encryption to be applied to a WLAN.

                The semantics are as follows:

                tkip - This value indicates that TKIP encryption
                is configured for data protection.

                aes  - This value indicates that AES encryption
                is configured for data protection. "
        SYNTAX   BITS {
                        tkip(0),
                        aes(1)
                 }
```

```
    sleOamFunctionsSupported OBJECT-TYPE
      SYNTAX BITS
        {
        unidirectionalSupport(0),
        loopbackSupport(1),
        eventSupport(2),
        variableSupport(3)
        }
      MAX-ACCESS read-only
      STATUS current
      DESCRIPTION
        "The OAM functions supported on this
        Ethernet like interface."
      ::= { sleEfmOamEntry 7 }
```

They all appear to be simple state sets to me, completely appropriate as labels.

brian-brazil:
Okay, updated.

PR #467, opened 2019-12-10
Sanitize snmp probe params

SuperQ:
Sanitize the snmp probe query params by only allowing them to be
specified once.

### DIFF
--- a/main.go
+++ b/main.go
@@ -68,13 +68,21 @@ func init() {
 }
 
 func handler(w http.ResponseWriter, r *http.Request, logger log.Logger) {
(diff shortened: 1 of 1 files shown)


stintel:
This change breaks our Prometheus snmp_exporter scraping. We used to query multiple mibs in a single scrape config:

```
    params:
      module:
        - if_mib
        - ubiquiti_unifi
```

If you are going to introduce breaking changes, please elaborate on the *why* in the commit message.

brian-brazil:
That query only used if_mib, the purpose of this PR was to make such misconfigurations more obvious.

PR #473, opened 2020-01-04
Add support for Raritan PDUs.

jsharkey:
This change downloads the latest PDU-MIB definition from Raritan,
which was last updated February 22, 2016, and then defines a handful
of useful values for monitoring.  This includes the current, voltage,
and wattage of both inlets and outlets on the PDU, and the current
switch status of each outlet.

This configuration has been tested and verified to work with a
Raritan DPXR8-15 unit in production.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -361,3 +361,25 @@ modules:
       - picoMobileMIB
       - picoIPv4MIB
       - picoIPv6MIB
+
+  raritan:
+    walk:
+      - sysUpTime
+      - 1.3.6.1.4.1.13742.4.1.20.2.1.7 # inletCurrent
+      - 1.3.6.1.4.1.13742.4.1.20.2.1.8 # inletVoltage
+      - 1.3.6.1.4.1.13742.4.1.20.2.1.9 # inletActivePower
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.31 # outletWattHours
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.3  # outletOperationalState
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.4  # outletCurrent
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.5  # outletMaxCurrent
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.6  # outletVoltage
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.7  # outletActivePower
+      - 1.3.6.1.4.1.13742.4.1.3.1.5    # unitCpuTemp
+    lookups:
+      - source_indexes: [outletIndex]
+        lookup: outletLabel
+    overrides:
+      outletOperationalState:
+        type: EnumAsStateSet
+    auth:
+      community: raritan_public
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -36,6 +36,7 @@ SERVERTECH_URL   := 'https://cdn10.servertech.com/assets/documents/documents/817
 SYNOLOGY_URL     := 'https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip'
 UBNT_AIROS_URL   := https://dl.ubnt.com/firmwares/airos-ubnt-mib/ubnt-mib.zip
 UBNT_DL_URL      := http://dl.ubnt-ut.com/snmp
+RARITAN_URL      := http://cdn.raritan.com/download/PX/v1.5.20/PDU-MIB.txt
 
 .DEFAULT: all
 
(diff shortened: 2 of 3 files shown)


generator/generator.yml
```diff
@@ -361,3 +361,19 @@ modules:
...
+      - 1.3.6.1.4.1.13742.4.1.20.2.1.7 # inletCurrent
+      - 1.3.6.1.4.1.13742.4.1.20.2.1.8 # inletVoltage
+      - 1.3.6.1.4.1.13742.4.1.20.2.1.9 # inletActivePower
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.3  # outletOperationalState
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.4  # outletCurrent
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.5  # outletMaxCurrent
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.6  # outletVoltage
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.7  # outletActivePower
+    auth:
+      community: raritan_private
```
brian-brazil: Is this the default for this hardware?

jsharkey: Good point, I could probably move this to "raritan_public" since I'm not doing write access.  Here's where the community string is documented, and it matches the factory reset defaults:

http://support.raritan.com/PX/version-1.5.7/quick-setup-guides/qsg-dpx-0n-v1-5-5-e.pdf

SuperQ: Yes, `raritan_public` would be better.

generator/generator.yml
```diff
@@ -361,3 +361,19 @@ modules:
       - picoMobileMIB
       - picoIPv4MIB
       - picoIPv6MIB
+
+  raritan:
+    walk:
+      - sysUpTime
+      - interfaces
+      - 1.3.6.1.4.1.13742.4.1.3.1.5    # unitCpuTemp
```
SuperQ: Typically we like to use the strings, rather than the OID numbers here.

generator/generator.yml
```diff
@@ -361,3 +361,19 @@ modules:
       - picoMobileMIB
       - picoIPv4MIB
       - picoIPv6MIB
+
+  raritan:
+    walk:
+      - sysUpTime
+      - interfaces
```
SuperQ: I'm thinking maybe we should leave `interfaces` off this module. People can get this from `if_mib`.

generator/generator.yml
jsharkey: Done  :)

generator/generator.yml
jsharkey: Done  :)

SuperQ:
A couple more minor suggestions, how about adding `outletWattHours`? Also, you could add a lookup like this:

```yaml
    lookups:
      - source_indexes: [outletIndex]
        lookup: outletLabel
```

jsharkey:
Bleck, now it seems to be pulling in outletVoltage from an unrelated MIB.  :/  I'll try hard-coding the OID to work around that?

SuperQ:
Yes, using the OID numbers for MIB conflicts is totally fine. It looks like `outletVoltage` is also in the servertech MIB. :-(

SuperQ:
One more suggestion.  `outletOperationalState` can have an override to map the states like this:

```yaml
    overrides:
      outletOperationalState:
        type: EnumAsStateSet
```

generator/generator.yml
```diff
@@ -361,3 +361,25 @@ modules:
...
+      - unitCpuTemp
+      - inletCurrent
+      - inletVoltage
+      - inletActivePower
+      - outletOperationalState
+      - outletCurrent
+      - outletMaxCurrent
+      - outletActivePower
+      - outletWattHours
+      - 1.3.6.1.4.1.13742.4.1.2.2.1.6 # outletVoltage
```
brian-brazil: We usually have the names, with the oids as a comment.

jsharkey: See comment thread earlier where using "outletVoltage" conflicted with another MIB and was causing build errors, which is why I switched to using the raw OID.

Or are you asking for everything to be in a format like this (depending on which variant needs to be listed first)?

  - name # OID
  - OID # name

brian-brazil: Put everything as the oid then I'd say, with the name as a comment.

jsharkey: I changed all Raritan-specific names to be OIDs instead, but I left the general "sysUpTime" in place unchanged.  This seems to mirror how other devices have been configured in this file, such as ubiquiti_airmax and ddwrt.

brian-brazil:
Thanks!

PR #481, opened 2020-02-14
Use contexts with SNMP.

### LINKED ISSUE #17: Timeout definition for snmp_exporter

brian-brazil:
When Prometheus times out the scrape and closes the http connection,
this will now make the snmp walk stop.

Fixes #17


RichiH:
I thought it did that already; very good idea.

SuperQ:
LGTM, https://github.com/prometheus/snmp_exporter/pull/482 should take care of the generator error.

Issue #487, opened 2020-03-04
Some 'context canceled' error in logs


shdxiang:
### Host operating system: output of `uname -a`
Docker container: Linux d8827dda937a 4.19.95-flatcar #1 SMP Sat Feb 8 07:25:12 -00 2020 x86_64 GNU/Linux

### snmp_exporter version: output of `snmp_exporter -version`

version=0.17.0, branch=HEAD, revision=f0ad4551a5c2023e383bc8dde2222f47dc760b83
f0ad4551a5c2023e383bc8dde2222f47dc760b83

### What device/snmpwalk OID are you using?
generator.yml
```yaml
nas:
    version: 2
    auth:
      community: public
    walk:
      - 1.3.6.1.4.1.24681.1.3.1 # systemCPU-UsageEX
      - 1.3.6.1.4.1.24681.1.3.11 # systemHdTableEX
      - 1.3.6.1.4.1.24681.1.3.15 # systemFanTable
      - 1.3.6.1.4.1.24681.1.3.17 # systemVolumeTable
```

### If this is a new device, please link to the MIB(s).
Download from QNAP itself.

### What did you do that produced an error?
I can get the metrics, but sometimes prometheus says that target is unreachable. And there are several error in `snmp-exporter` logs.

### What did you expect to see?
No error in logs.

### What did you see instead?
```bash
snmp-exporter_1       | level=info ts=2020-03-04T07:20:52.964Z caller=collector.go:216 module=dell_idrac target=172.30.60.220:161 msg="Error scraping target" err="error walking target 172.30.60.220: context canceled"
snmp-exporter_1       | level=info ts=2020-03-04T07:21:28.405Z caller=collector.go:216 module=nas target=172.30.50.104:161 msg="Error scraping target" err="error getting target 172.30.50.104: context canceled"
snmp-exporter_1       | level=info ts=2020-03-04T07:22:01.730Z caller=collector.go:216 module=dell_idrac target=172.30.60.209:161 msg="Error scraping target" err="error walking target 172.30.60.209: context canceled"
snmp-exporter_1       | level=info ts=2020-03-04T07:24:28.404Z caller=collector.go:216 module=nas target=172.30.50.104:161 msg="Error scraping target" err="error getting target 172.30.50.104: context canceled"
[...]

PR #488, opened 2020-03-04
Improve canceled context messages

SuperQ:
Catch SNMP context canceled errors so they are displayed to the user as
scrape canceled.

Fixes: https://github.com/prometheus/snmp_exporter/issues/487

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -109,6 +109,9 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 	// Do the actual walk.
 	err := snmp.Connect()
 	if err != nil {
+		if err == context.Canceled {
+			return nil, fmt.Errorf("scrape canceled (possible timeout) connecting to target %s", snmp.Target)
+		}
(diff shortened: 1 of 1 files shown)


collector.go
```diff
@@ -109,6 +109,9 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 	// Do the actual walk.
 	err := snmp.Connect()
 	if err != nil {
+		if err == context.Canceled {
+			return nil, fmt.Errorf("timeout connecting to target %s", target)
```
brian-brazil: This is incorrect, it's the scrape request's tcp connection being closed - not a timeout.

SuperQ: Ahh, ok.

brian-brazil: You're still using the word timeout. Any timeout is on the Prometheus side, not in the SNMP exporter.

SuperQ: How about `scrape canceled (possible timeout) ...`?

brian-brazil: That could work

Issue #405, opened 2019-04-25
Support chained lookups


tarko:
Hey,

Given the following metrics (coming straight from CISCO-CLASS-BASED-QOS-MIB):

```
ifIndex{ifName="Fa4"} 5

cbQosIfIndex{cbQosPolicyIndex="82"} 5

cbQosCMDropPkt64{cbQosObjectsIndex="1472497",cbQosPolicyIndex="82"} 116235
cbQosCMDropPkt64{cbQosObjectsIndex="14967793",cbQosPolicyIndex="82"} 116235
cbQosCMDropPkt64{cbQosObjectsIndex="5207969",cbQosPolicyIndex="82"} 0
cbQosCMDropPkt64{cbQosObjectsIndex="8687729",cbQosPolicyIndex="82"} 0

cbQosConfigIndex{cbQosObjectsIndex="1472497",cbQosPolicyIndex="82"} 1593
cbQosConfigIndex{cbQosObjectsIndex="14967793",cbQosPolicyIndex="82"} 1593
cbQosConfigIndex{cbQosObjectsIndex="5207969",cbQosPolicyIndex="82"} 296545
cbQosConfigIndex{cbQosObjectsIndex="8687729",cbQosPolicyIndex="82"} 18785

cbQosCMName{cbQosCMName="CS6",cbQosConfigIndex="296545"} 1
cbQosCMName{cbQosCMName="EF",cbQosConfigIndex="18785"} 1
cbQosCMName{cbQosCMName="class-default",cbQosConfigIndex="1593"} 1
```

and wanting to get cbQosCMDropPkt64 for specific ifName with cbQosCMName label added, it is pretty straightforward to go from ifIndex -> cbQosIfIndex -> cbQosCMDropPkt64 so one ends up with:

```
cbQosCMDropPkt64{cbQosObjectsIndex="1472497",cbQosPolicyIndex="82"}
cbQosCMDropPkt64{cbQosObjectsIndex="14967793",cbQosPolicyIndex="82"}
cbQosCMDropPkt64{cbQosObjectsIndex="5207969",cbQosPolicyIndex="82"}
cbQosCMDropPkt64{cbQosObjectsIndex="8687729",cbQosPolicyIndex="82"}
```

Getting the cbQosCMName is a problematic bit as this would require joining cbQosObjectsIndex value (for example 1593) with cbQosConfigIndex label on cbQosCMName metric.

[...]

PR #489, opened 2020-03-04
WIP: allow lookup chaining in a basic way (#405)

debugloop:
I've added the basic ability of chaining lookups as discussed in #405 to the exporter, but I am unsure wether there should be support in the generator for this type of lookup and how it should look like. I've thought about marking the lookup as `intermediate` or something via the `type:` config directive, but having lookups reusable by default seems okay to me. Thoughts?


brian-brazil:
This doesn't look right, and my previous investigations indicated this was a generator thing.

debugloop:
Yup, you said [here](https://github.com/prometheus/snmp_exporter/issues/405#issuecomment-560390426) that the exporter can handle that, but I can not reproduce this without my patch.

I am testing with this:

```
cpu:
  walk:
  - 1.3.6.1.4.1.9.9.109.1.1.1.1 # ciscos cpu table thingie
  - 1.3.6.1.2.1.47.1.1.1.1.7 # entity mib physical names
  metrics:
  - name: cpmCPUTotal1minRev
    oid: 1.3.6.1.4.1.9.9.109.1.1.1.1.7
    type: gauge
    indexes:
    - labelname: cpmCPUTotalIndex
      type: gauge
    lookups:

    # get physical entity id from some metric
    - labelname: entPhysicalIndex
      labels: [cpmCPUTotalIndex]
      oid: 1.3.6.1.4.1.9.9.109.1.1.1.1.2
      type: DisplayString

    # reuse that id to lookup the entity name
    - labelname: entPhysicalName
      labels: [entPhysicalIndex]
      oid: 1.3.6.1.2.1.47.1.1.1.1.7
      type: DisplayString
```

brian-brazil:
I think you're right. Can you add a unittest for this?

debugloop:
I've added a basic one to this function's test case list, not sure if it is enough. Gives the appropriate result for both, this version and `v0.17.0`

PR #496, opened 2020-04-05
the feature which allows to upload config and reload, was developed to snmp_exporter.

bkayranci:
Thanks to the feature, config file can be upload and reload. Also, uploaded config can be permanent if prefer it. @brian-brazil

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,3 +69,12 @@ In order to provide accurate counters for large Counter64 values, the exporter w
 wrap the value every 2^53 to avoid 64-bit float rounding.
 
 To disable this feature, use the command line flag `--no-snmp.wrap-large-counters`.
+
+## Upload dynamic config
+
+Snmp exporter allow to upload config file and reload it.
+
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for your PR. By design the only way to change Prometheus config files is by changing them on disk, if you need this sort of feature you can do so outside of the exporter.

bkayranci:
> Thanks for your PR. By design the only way to change Prometheus config files is by changing them on disk, if you need this sort of feature you can do so outside of the exporter.

Thank you for fast review.

The feature can already change config on disk. It was developed for the scenario following,
Prometheus was configured dynamic target for example file_sd_configs. However snmp_exporter's config can not allow change config file via http request. Thanks to the feature, we can it.

Am i thinking wrong?

brian-brazil:
This feature is explicitly something that we don't want. The only place configuration comes from is disk.

bkayranci:
Thank you for your review. I closed the PR.

Issue #432, opened 2019-07-15
Pull specific interfaces with snmp.yml


freedomwarrior:
Hello.
How can I pull only ethernetCsmacd interfaces from device? For example, I got network switch with around 1k l2vlan and I don`t need to pull them via snmp-exporter. How can I do that?

SuperQ:
I don't think there's a way to do that kind of filtering right now. But it's an interesting idea.

There have been some ideas on adding filtering based on OID index ranges, but those are also not implemented yet.

brian-brazil:
You can specify specific oids though within a table.

freedomwarrior:
> You can specify specific oids though within a table.

Can you show me some example?

freedomwarrior:
A good option will be ifType filter, for vlan interface:` l2vlan`, for ethernet interface: `ethernetCsmacd`

freedomwarrior:
Some solution: rfc1271, etherStatsDataSource

freedomwarrior:
Hello again.
Any update?

SuperQ:
@freedomwarrior PRs welcome.

orgito:
For network devices, it would be great if we could filter based on interface description. For some networks that could mean the difference between polling half a million interfaces or 20 thousand interfaces.

brian-brazil:
I'm not sure this is something that we should have in the snmp exporter itself, as it means we'd need a two pass walk. I think figuring out what OIDs you want offline and putting them in the config would be better.

brian-brazil:
Per @SuperQ, we may want to be able to notice a bunch of gets for adjacent oids and use a walk instead for efficiency.

xkilian:
The current method, thanks to #283 can now reference a specific index in the walk section of the generator.yml:      - 1.3.6.1.2.1.31.1.1.1.6.40  # Instance of "ifHCInOctets" with index "40"
I understand that the generator is being kept as simple as possible and would avoid two pass collection on each run. Here are two suggestions for a solution:

Extend what was done in #283 to permit some kind of syntax with a list or range: 1.3.6.1.2.1.31.1.1.1.6.[40,41,42,45,199] # 40,41,42,45,199 being instances. But I think the yml config fille will not like this
This puts all the hard work outside generator.yml.

[...]

PR #497, opened 2020-04-06
constant label parameter added

bkayranci:
@brian-brazil 

Here the use case is as follows;
* We have thousands of interfaces on the network elements that we need to monitor (say 10000)
* We're only interested in a subset of these interfaces (say 100)
* We know the "ifIndex" of these specific interfaces along with their descriptions etc. (we have network inventory)
* If snmp_exporter allows us to "specify" "constant labels" for the metrics, we're able to define "metrics" and assign "constant labels" to each metric.

Example config:

```yaml
10.0.0.2:
  get:
  - 1.3.6.1.2.1.31.1.1.1.6.1
  - 1.3.6.1.2.1.31.1.1.1.6.10
  metrics:
  - name: ifHCInOctets
    oid: 1.3.6.1.2.1.31.1.1.1.6.1
    type: counter
    const_labels:
      lag_name: LAG00
      port_name: Ge 1/1/0
      region_name: NY
      city_name: NYC
      ifIndex: 1 

  - name: ifHCInOctets
    oid: 1.3.6.1.2.1.31.1.1.1.6.10
    type: counter
    const_labels:
      lag_name: LAG01
      port_name: Ge 2/0
      region_name: NY
      city_name: NYC
      ifIndex: 10
```

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -327,12 +327,30 @@ func pduToSamples(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, o
 
 	value := getPduValue(pdu)
 	t := prometheus.UntypedValue
+	
+	labels_length := len(labels) + len(metric.Labels) + 1
 
-	labelnames := make([]string, 0, len(labels)+1)
-	labelvalues := make([]string, 0, len(labels)+1)
+	labelnames := make([]string, 0, labels_length)
+	labelvalues := make([]string, 0, labels_length)
+
+	// Add labels returned from "indexes to labels"
 	for k, v := range labels {
 		labelnames = append(labelnames, k)
 		labelvalues = append(labelvalues, v)
+	}
+
+	// Add constant labels if any
+	if len(metric.Labels) > 0 {
+		level.Debug(logger).Log("msg", "Metric:", "Name", metric.Name, "OID", metric.Oid)
+
+		for ln, lv := range metric.Labels {
+			tmp_label_name := string(ln)
+			tmp_label_value := string(lv)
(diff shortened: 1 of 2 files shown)


brian-brazil:
Thanks for your PR. Business logic like this is out of scope for the snmp exporter, it exposes SNMP data rather than trying to be a general network database.

You can specify the oids you want in the walk section already, and I'd suggest using ifDescr and friends for the rest.

karanlik:
Hi @brian-brazil ,

We're working with @bkayranci on this issue and we believe this solution can help tickets like following;
https://github.com/prometheus/snmp_exporter/issues/432

Here the main idea is "allowing" snmp_exporter users to add "some constant label" to the metrics that they want to. This "const_labels" is optional. People like us (with network inventory in hand) can create their own "generator" projects, read information from their inventory DB, and create "snmp.yml" files accordingly. I see this is the only way to allow people not to walk all those thousands of interfaces and specify only a subset of these interfaces. 

Since Prometheus metrics are unique with name + labels, providing a "get" OID list won't help only. Here the alternatives would be;
1) define the metric names to bu unique as ifHCInOctets_if_123
2) definte the metric names to be the same as "ifHCInOctets" AND let the "const_labels" account for the "uniqueness".

Can you please share your comments in this respect ?

thanks,

brian-brazil:
#432 is separate, and all the support needed within the snmp exporter to is already there for you to build on.

> Since Prometheus metrics are unique with name + labels, providing a "get" OID list won't help only.

I think you're getting confused. I'm talking about the walk config option of the generator config, which already handles this correctly.

brian-brazil:
This is adding things which don't make sense in snmp-exporter terms, and I believe we already support what you need in the generator.

PR #511, opened 2020-04-27
Reduce and fix timeout/docs for SNMP requests

### LINKED ISSUE #507: SNMP Exporter Timeout
### Host operating system: output of `uname -a`

Linux hostname 4.9.0-4-amd64 #1 SMP Debian 4.9.65-3+deb9u1 (2017-12-23) x86_64 GNU/Linux

### snmp_exporter version: output of `snmp_exporter -version`
0.17.0

We are walking lots of different devices/manufacturers (Mikrotik devices, Huawei switches, Ruckus wireless switches, Ruckus Wireless APs, Huawei APs, etcetera)

We are using curl to execute the query ourselves.

curl 'http://127.0.0.1:9116/snmp?target=10.1.32.110&module=mikrotik'
curl 'http://127.0.0.1:9116/snmp?target=10.1.32.110&module=huawei_switch'

We expect to see SNMP results. Most of the time, the queries succeed and we get back the data. Now and then (there's no correlation as to when it succeeds or fails), the SNMP query times out. Actually after taking packet captures, we see that the query is reaching the device, the device answers, but the response gets lost somewhere along the way. (These queries/replies are being sent over VPN tunnels over the public Internet.)

Due to the connectionless nature of UDP, we expect the SNMP exporter to send another query (after timeout interval of 10s, retries this 3 times; default values). 

### What did you see instead?

However, as you can see below, the snmp exporter doesn't send another query for 60 seconds:

Screenshot 2020-04-24 at 14 40 41

brian-brazil:
Upstream changed how their timeout&retries interacted, so
we no longer need to adjust things.

Our docs and code were also out of sync, but let's just reduce timeouts
to 2s as that should be plenty of time for a single GETBULK

Fixes #507

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -47,7 +47,7 @@ var (
 		Version:        2,
 		MaxRepetitions: 25,
 		Retries:        3,
-		Timeout:        time.Second * 20,
+		Timeout:        time.Second * 5,
 		Auth:           DefaultAuth,
 	}
 	DefaultModule = Module{
--- a/collector.go
+++ b/collector.go
@@ -88,9 +88,8 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 	snmp := gosnmp.GoSNMP{}
 	snmp.Context = ctx
 	snmp.MaxRepetitions = config.WalkParams.MaxRepetitions
-	// User specifies timeout of each retry attempt but GoSNMP expects total timeout for all attempts.
 	snmp.Retries = config.WalkParams.Retries
-	snmp.Timeout = config.WalkParams.Timeout * time.Duration(snmp.Retries)
+	snmp.Timeout = config.WalkParams.Timeout
 
 	snmp.Target = target
 	snmp.Port = 161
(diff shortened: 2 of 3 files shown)


config/config.go
```diff
@@ -47,7 +47,7 @@ var (
 		Version:        2,
 		MaxRepetitions: 25,
 		Retries:        3,
-		Timeout:        time.Second * 20,
+		Timeout:        time.Second * 2,
```
SuperQ, line 50: I'd suggest leaving this a little bit higher, maybe 5s. Some devices/networks are particularly slow.

brian-brazil: For a single GETBULK that's close on the network?

There is also an exponential option (which is what broke us) now that we could use.

brian-brazil: This is also a default, so if there's slow devices they could be configured by with a longer value without slowing down everyone who only loses a packet.

SuperQ: Yes, I think the default should be relatively relaxed for this kind of thing.

brian-brazil: Changed.

PR #514, opened 2020-05-05
update port to variable from hard code

cawamata:
## Issue
Snmp port number is hard code (161).
  - https://github.com/prometheus/snmp_exporter/blob/v0.17.0/collector.go#L96

So I cannot use this exporter when not using default port by snmpd

## Proposal
User can set port number on snmp.yml (via generator)

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -93,7 +93,7 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 	snmp.Timeout = config.WalkParams.Timeout * time.Duration(snmp.Retries)
 
 	snmp.Target = target
-	snmp.Port = 161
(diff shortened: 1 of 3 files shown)


SuperQ:
This is unnecessary, it is simply a default. If you specify the port in your target, it will be parsed from the target input.

See [the next several lines of code](https://github.com/prometheus/snmp_exporter/blob/87c61e2698a4c7a9d436c8885109ebdbf8756660/collector.go#L97-L104).

For example `/snmp?module=if_mib&target=10.0.0.1:11611` would use port 11611.

cawamata:
@SuperQ san
thank you for your information.

PR #522, opened 2020-05-20
Treat an OBJID in an index as an OctetString.

brian-brazil:
We've not seen a non-null one of these yet so rendering is unclear, so
this is just enough to unbreak tables using them.

@SuperQ


SuperQ:
LGTM, looks like the snmp.yml needs updating with this.

PR #523, opened 2020-05-20
Updating to APC powernet432.mib

chewrocca:
Changing url to reflect new powernet432.mib.


brian-brazil:
That's a fairly old branch you've got there, you'll want to rebase and recompile.

There's also something weird going on with the Ubiquiti MIBs that I haven't figured out yet - may I ask which OS version you're on?

chewrocca:
I tried generate on MacOS, Ubuntu 16.04, and Docker.  I think you're right about the Ubiquiti MIBs.  Each time I run generate depending on os or Docker it seems to produce different results.

chewrocca:
I’m not sure what you mean about needing to rebase my branch. Can you explain? My branch adds 4 commits on top of the current HEAD of the upstream master branch (https://github.com/prometheus/snmp_exporter/commit/52c305c26faff4b9d9bd492d82d4c3d14a705b49) and should be a fast-forward.

brian-brazil:
I'd avoid OS X with the generator, I've seen weird stuff there with netsnmp and indexes that I never got to the bottom of (and it's not an OS I have easy access to).

brian-brazil:
Thanks!

PR #527, opened 2020-06-08
Allow lookup chaining for uints

tarko:
I don't see how `gosnmp.ToBigInt` could return anything else than ints so I think it's safe to do it without any additional checks. Also added testcase that failed before but do not fail any more.


brian-brazil:
Thanks!

PR #544, opened 2020-06-30
Add support for mini_snmpd

NightTsarina:
Some small devices running openWRT can't afford proper SNMP daemons and use mini_snmpd instead, this adds support for this basic daemon, which does not implement if_mib.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -241,6 +241,25 @@ modules:
       ifType:
         type: EnumAsInfo
 
+# mini_snmpd / OpenWRT
+#
+# Small SNMP daemon included in OpenWRT (among others).
+# The list of SNMP OIDs to care about can be found here: https://openwrt.org/docs/guide-user/services/snmp/mini_snmpd
+#
+# Tested on TP-LINK TL-WR703N v1 with OpenWRT 17.01.7.
+#
+  mini_snmpd:
+    walk: [hrSystemUptime, sysUpTime, interfaces]
+    lookups:
+      - source_indexes: [ifIndex]
+        # Uis OID to avoid conflict with PaloAlto PAN-COMMON-MIB.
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'm confused, `interfaces` is IF-MIB - so the existing module should cover it. It'd also be advised to use the node exporter variant over the snmp exporter on something like OpenWrt

NightTsarina:
> I'm confused, `interfaces` is IF-MIB - so the existing module should cover it. It'd also be advised to use the node exporter variant over the snmp exporter on something like OpenWrt

Not exactly.. if-mib introduced ifName and ifAlias, which are not supported by this daemon. It corresponds with pattern 7 [here](https://web.archive.org/web/20180830150814/www.irobert.info/display/IRINFO/Understanding+ifDescr,+ifName+and+ifAlias). The result of using if-mib with this daemon is having empty ifName and ifAlias labels, which just add clutter and confusion.

Agree about running the node exporter, but it is not available for all routers that are stuck with older versions of openWRT, like mine. And I am not even sure this box will have enough memory for it.

brian-brazil:
> The result of using if-mib with this daemon is having empty ifName and ifAlias labels, which just add clutter and confusion.

That's harmless, an empty label is the same as no label.

SuperQ:
Yes, Prometheus will ignore the empty labels and not store them.

Another nice option for OpenWRT users is the [LUA node_exporter](https://github.com/openwrt/packages/tree/master/utils/prometheus-node-exporter-lua) provided by the OpenWRT packages repo. I use this on my Turris router.

NightTsarina:
Well, as you wish. It seemed to me this is the correct way to implement scraping of a snmp that does not implement if-mib. Feel free to close

SuperQ:
The example snmp.yml is just an example. It's not meant to handle every case for every device. Users are free to, and meant to, modify the generator.yml for their own needs.

PR #565, opened 2020-10-08
Add a way to update configuration by passing http request body

ahuret:
Renamed current `updateConfiguration` handler to `reloadConfiguration`.
Added a new route '/-/update' and its handler function `updateConfiguration` that allows to update the configuration by passing the config via the http request body (request POST)
Changed few related things.
Doesn't break BC.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. As a matter of general policy we only read configuration files from disk, and don't attempt to do any fancy configuration management or config database features ourselves. If you need this you should handle this by having a separate process update the file on disk.

ahuret:
7-1

PR #571, opened 2020-10-22
Add support for Kemp LoadMaster

tumd:
I haven't tested this extensively, but I've verified that all the relevant fields seem to export correctly, and added suitable overrides to the values that was in need of conversion.

Based on a blogpost by Bas Roovers.
https://onpremisys.nl/2020/01/06/monitoring-kemp-loadmaster-using-snmp-and-prometheus

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -27,6 +27,7 @@ IANA_CHARSET_URL  := https://www.iana.org/assignments/ianacharset-mib/ianacharse
 IANA_IFTYPE_URL   := https://www.iana.org/assignments/ianaiftype-mib/ianaiftype-mib
 IANA_PRINTER_URL  := https://www.iana.org/assignments/ianaprinter-mib/ianaprinter-mib
 KEEPALIVED_URL    := https://raw.githubusercontent.com/acassen/keepalived/v2.1.5/doc/KEEPALIVED-MIB.txt
+KEMP_LM_URL       := https://kemptechnologies.com/files/packages/current/LM_mibs.zip
 MIKROTIK_URL      := 'http://download2.mikrotik.com/Mikrotik.mib'
 NEC_URL           := https://jpn.nec.com/univerge/ix/Manual/MIB
 NET_SNMP_URL      := https://raw.githubusercontent.com/net-snmp/net-snmp/v5.9/mibs
@@ -53,7 +54,8 @@ clean:
 		$(MIBDIR)/.cisco_v2 \
 		$(MIBDIR)/.net-snmp \
 		$(MIBDIR)/.paloalto_panos \
-		$(MIBDIR)/.synology
+		$(MIBDIR)/.synology \
+		$(MIBDIR)/.kemp-lm
 
 generator: *.go
 	go build
@@ -87,6 +89,7 @@ mibs: mib-dir \
   $(MIBDIR)/IANA-IFTYPE-MIB.txt \
   $(MIBDIR)/IANA-PRINTER-MIB.txt \
   $(MIBDIR)/KEEPALIVED-MIB \
+  $(MIBDIR)/.kemp-lm \
   $(MIBDIR)/MIKROTIK-MIB \
   $(MIBDIR)/.net-snmp \
   $(MIBDIR)/.paloalto_panos \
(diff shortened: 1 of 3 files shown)


generator/generator.yml
```diff
@@ -362,6 +362,64 @@ modules:
...
+      - ssCpuIdle
+      - memory
+      - 1.3.6.1.4.1.12196.13.0 # VSdesc
+      - 1.3.6.1.4.1.12196.13.1 # VSentry
+      - 1.3.6.1.4.1.12196.13.2 # RSentry
+    lookups:
+      - source_indexes: [vSidx]
+        lookup: 1.3.6.1.4.1.12196.13.1.1.13 # vSName
+    overrides:
+      daemonState:
```
brian-brazil, line 386: Does this change over time? Usually things called "state" would be statesets, not info.

There's other states too below.

brian-brazil (commented):
It looks like the MIB is invalid per netsnmp.

generator/generator.yml
tumd: I suppose the "state"'s won't change over time in the MIB's. Not entirely sure however. It seem like some of the "state"s has changed in the past. `vSState` for example, seem to have removed one of it's elements at some point (3 is missing);
```
vSState OBJECT-TYPE
  SYNTAX    INTEGER {
    inService(1),
    outOfService(2),
    disabled(4),
    sorry(5),
    redirect(6),
    errormsg(7),
    securityDown(8)
  }
```
`rSState` is similar to `vSState` in missing the 3:rd element.
`daemonState` & `haState` are simpler in that regard.

What is your recommendation?

tumd:
What would you suggest regarding the invalid timestamps? I guess the preferred would be to ask the vendor to correct the MIB's upstream.
Is there any other way around this?

generator/generator.yml
SuperQ: The "change over time" referrers to the device, not the MIB. Think about ethernet link state (changes over time when you plug/un-plug a cable) vs ethernet port type (gige, 10gige, sfp) that doesn't really change over time.

tumd: Ah, I see. :)
In that case, they all may change on the device over time.

SuperQ:
Strange, I don't get a parse error on net-snmp 5.8.

brian-brazil:
We'd first have to check the RFCs to see what's going on here, this isn't an error I've come across before.

tumd:
What I've found the [RFC2578](https://tools.ietf.org/html/rfc2578#section-2) (ExtUTCTime) quite clearly states the format of the value for `LAST-UPDATED` & `REVISION`, which the upstream vendor MIB's doesn't comply with. I haven't found any newer RFC's replacing this section in RFC2578.
I'll reach out to the vendor and ask them to revise.

brian-brazil:
That looks to be it alright, they've included second which aren't permitted.

SuperQ:
One option would be to add a small patch script to the Makefile to fix up the strings.

generator/Makefile
```diff
@@ -155,6 +158,15 @@ $(MIBDIR)/KEEPALIVED-MIB:
 	@echo ">> Downloading KEEPALIVED-MIB"
 	@curl $(CURL_OPTS) -o $(MIBDIR)/KEEPALIVED-MIB $(KEEPALIVED_URL)
 
+$(MIBDIR)/.kemp-lm:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading Kemp LM MIBs to $(TMP)"
+	@curl $(CURL_OPTS) -o $(TMP) $(KEMP_LM_URL)
+	@unzip -j -d $(MIBDIR) $(TMP) *.txt
+	@sed -i -E "s/([0-9]{12})[0-9]{2}Z/\1Z/" $(MIBDIR)/*.RELEASE-B100-MIB.txt
```
SuperQ, line 166: Would you mind adding a comment here that this is a workaround? Plus a minor suggestion for the regexp.

```suggestion
	# Workaround invalid timestamps.
	@sed -i -E 's/"([0-9]{12})[0-9]{2}Z"/"\1Z"/' $(MIBDIR)/*.RELEASE-B100-MIB.txt
```

brian-brazil:
Was there any work back from the vendor? I'd rather not get into the game of altering vendor-provided MIBs if we don't have to.

generator/Makefile
tumd: I agree on your suggestions and have updated the commit.
Initially I had a similar comment included locally, but was annoyed/confused of the fact that `make mibs` printed the comment, thus wasn't confident enough to include it.

tumd:
At last Friday (13'th) the vendor said;
> The MIB issue has be recognized as a bug and I'm just waiting on the development team to provide a date for the fix.
In other words a fix may come tomorrow or in 6 months.

I totally understand that you don't want such workarounds in the code. Let me know and I'll revert it.

SuperQ:
I think the sed patch is small and simple enough that we can keep it. We're not changing any functionality of the MIB, just working around a syntax issue.

SuperQ (approved):
LGTM

brian-brazil:
Thanks!

PR #573, opened 2020-10-31
Properly parse the value of PhysAddress48 string metrics

Setheck:
I discovered what appears to be a small bug. After generating an snmp.yml that contained following section
```yaml
  - name: ifPhysAddress
    oid: 1.3.6.1.2.1.2.2.1.6
    type: PhysAddress48
    help: The interface's address at its protocol sub-layer - 1.3.6.1.2.1.2.2.1.6
    indexes:
    - labelname: ifIndex
      type: gauge
```

Not sure if it matters, but my case was retrieving physical addresses from mac minis. I did some digging and as it turns out the snmp_exporter only handles the `PhysAddress48` type if it is returned as a `[]byte` but it appears that in some cases it is actually returned from gosnmp as a `string`. Included is the simple logic and test to handle this case.  
When a user such as I encounters this case, the snmp_exporter will scrape but not return any values scraped, instead errors generated by [this line](https://github.com/prometheus/snmp_exporter/blob/master/collector.go#L409) show. 

The one case I did not handle is that it is possible for the string to come back as `""` in this case the resulting mac address is `00:00:00:00:00:00`.  

@brian-brazil small fix as per contributing instructions.

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -552,6 +552,14 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ string) string {
 			// Trim leading period.
 			return pdu.Value.(string)[1:]
 		}
+		if typ == "PhysAddress48" {
+			parts := make([]int, len([]byte(pdu.Value.(string))))
+			for i, o := range []byte(pdu.Value.(string)) {
+				parts[i] = int(o)
+			}
+			str, _, _ := indexOidsAsString(parts, typ, 0, false, nil)
+			return str
+		}
 		// DisplayString.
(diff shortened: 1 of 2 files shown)


brian-brazil:
Hmm, I think this was a bug introduced by https://github.com/gosnmp/gosnmp/pull/202 which changed the return type of OctectStrings from []byte to string. This should be fixed upstream.

Setheck:
Awesome, Thanks!

SuperQ:
I've merged the upstream vendor fix.

PR #576, opened 2020-11-02
Add TLS support

coderigo:
Allow exporter to listen over https

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,9 +37,12 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks for your PR. The plan is to have one central TLS implementation with all the required features, rather than re-ad-hoc implementing partial versions in each repository.

coderigo:
ok thanks

Issue #578, opened 2020-11-04
Add Juniper MIBs


loganmc10:
Just a feature request to add Juniper MIBs by default, they are available for download here:

https://apps.juniper.net/mib-explorer/download.jsp#product=Junos%20OS

brian-brazil:
If they're wgettable, feel free to send a PR for devices.

PR #579, opened 2020-11-05
Add JunOS MIBs

loganmc10:
Add vendor provided JunOS MIBs. Fixes https://github.com/prometheus/snmp_exporter/issues/578

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -41,6 +41,7 @@ UBNT_DL_URL       := http://dl.ubnt-ut.com/snmp
 RARITAN_URL       := http://cdn.raritan.com/download/PX/v1.5.20/PDU-MIB.txt
 INFRAPOWER_URL    := https://www.austin-hughes.com/support/software/infrapower/IPD-MIB.7z
 LIEBERT_URL       := https://www.vertiv.com/492204/contentassets/b00273585e0a453a9c983523e8a0d6ff/lgpmib-unix_rev16.tar
+JUNOS_URL         := https://www.juniper.net/techpubs/software/junos/junos203/juniper-mibs-20.3R1.8.zip
 
 .DEFAULT: all
 
@@ -53,7 +54,8 @@ clean:
 		$(MIBDIR)/.cisco_v2 \
 		$(MIBDIR)/.net-snmp \
 		$(MIBDIR)/.paloalto_panos \
-		$(MIBDIR)/.synology
+		$(MIBDIR)/.synology \
+		$(MIBDIR)/.junos
 
 generator: *.go
 	go build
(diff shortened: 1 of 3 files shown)


brian-brazil:
This should include a generator.yml change, no point in having MIBs we don't use.

loganmc10:
Isn't the idea that users can write their own generator.yml to fit their needs? Or do you mean just to include a few things as an example/base for people?

brian-brazil:
That's the general idea of the generator. The MIBs referenced here are used for the example generator.yml - and if a MIB isn't used there there's no point in downloading or processing it. So the goal of a PR like this should be to add some useful Juniper-specific modules.

loganmc10:
I added an example for gathering the JunOS alarm values. I re-ran the generator and replaced the included snmp.yml file. Is this generally the right process? As you can see it removed a number of enum values when I did that

brian-brazil:
That's the right process, I'm guessing the MIBs you included include an older version of something so we'll need to be more judicious in what we use from Juniper.

The examples should also be reasonably usable as a starting point, so a bit more than just two metrics.

loganmc10:
Ok I was more picky about what gets pulled in from Junipers "StandardMibs" folder and that solved the problem. I added a few more metrics to the generator for JunOS. I think just having the MIBs available like this makes it a lot better for JunOS users. All they need to do is all the OIDs they want, without needing to worry about importing the MIBs the right way.

I guess what I'm saying is that I think there is value in having MIBs in this repository even if the example generator.yml is fairly sparse. Cisco users already had this benefit since the Cisco MIBs are already there.

generator/Makefile
```diff
@@ -247,3 +250,25 @@ $(MIBDIR)/LIEBERT_GP_PDU.MIB:
...
 	@tar --no-same-owner -C $(MIBDIR) -xvf $(TMP)
 	@rm -v $(TMP)
+
+$(MIBDIR)/.junos:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading JunOS"
+	@curl $(CURL_OPTS) -o $(TMP) $(JUNOS_URL)
+	@unzip -j -d $(MIBDIR) $(TMP) "JuniperMibs/*" \
+                                      "StandardMibs/mib-rtmib.txt" \
+                                      "StandardMibs/mib-rfc*.txt" \
```
brian-brazil, line 260: I'd expect that we'd already have these from Cisco, keep this minimal.

generator/generator.yml
```diff
@@ -503,3 +503,13 @@ modules:
...
         lookup: lgpPduAuxSensorSysAssignLabel
+
+# JunOS
+#
+# https://iphostmonitor.com/mib/JUNIPER-ALARM-MIB.html
+  junos:
+    walk:
+      - sysUpTime
+      - interfaces
+      - 1.3.6.1.4.1.2636.3.4.2.3.1 #Red Alarm state
```
brian-brazil, line 514: This is still only 2 metrics. These examples should be something that someone would plausibly use in production.

brian-brazil (commented):
> I guess what I'm saying is that I think there is value in having MIBs in this repository even if the example generator.yml is fairly sparse.

Users can already download them themselves, they've not much use just being a URL in a file.

loganmc10:
Ok thanks I'll just fork the repo

PR #592, opened 2020-12-17
Add Mikrotik module

carlosedp:
Add RFC1213 MIB to be able to generate configs from OIDs 1.3.6.1.2.1.4.

Added Mikrotik Router generator module.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -612,3 +612,30 @@ modules:
         lookup: lgpPduRcpEntrySysAssignLabel
       - source_indexes: [lgpPduAuxSensorIndex]
         lookup: lgpPduAuxSensorSysAssignLabel
+# Mikrotik Router
+#
+# http://download2.mikrotik.com/Mikrotik.mib
+  mikrotik:
+    walk:
+      - interfaces
+      - ifMIB
+      - laIndex
+      - sysUpTime
+      - sysDescr
+      - host
+      - mikrotik
+    lookups:
+      - source_indexes: [ifIndex]
+        lookup: ifName
+      - source_indexes: [mtxrInterfaceStatsIndex]
+        lookup: ifName
+      - source_indexes: [hrStorageIndex]
+        lookup: hrStorageDescr
+      - source_indexes: [laIndex]
+        lookup: laNames
+        drop_source_indexes: true
+    overrides:
+      ifName:
+        ignore: true # Lookup metric
+      ifType:
+        type: EnumAsInfo
--- a/snmp.yml
+++ b/snmp.yml
@@ -12163,6 +12163,3966 @@ liebert_pdu:
       labelname: lgpPduAuxSensorSysAssignLabel
       oid: 1.3.6.1.4.1.476.1.42.3.8.60.10.1.15
       type: DisplayString
+mikrotik:
+  walk:
+  - 1.3.6.1.2.1.2
+  - 1.3.6.1.2.1.25
+  - 1.3.6.1.2.1.31
+  - 1.3.6.1.4.1.14988
+  - 1.3.6.1.4.1.2021.10.1.1
+  - 1.3.6.1.4.1.2021.10.1.2
+  get:
+  - 1.3.6.1.2.1.1.1.0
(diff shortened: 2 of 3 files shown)


brian-brazil:
We only add mibs for the sake of the example generator config as otherwise they're dead weight, so can you add a module there for the device type you're interested in?

carlosedp:
Sure, added Mikrotik router module.

generator/generator.yml
```diff
@@ -612,3 +612,42 @@ modules:
...
+      - source_indexes: [laIndex]
+        lookup: laNames
+        drop_source_indexes: true
+    overrides:
+      ifAlias:
+        ignore: true # Lookup metric
+      ifDescr:
+        ignore: true # Lookup metric
+      ifName:
+        ignore: true # Lookup metric
```
brian-brazil, line 651: This is not looked up.

generator/generator.yml
```diff
@@ -612,3 +612,42 @@ modules:
...
+      - laIndex
+      - sysUpTime
+      - sysDescr
+      - host
+      - mikrotik
+    lookups:
+      - source_indexes: [ifIndex]
+        lookup: ifAlias
+      - source_indexes: [ifIndex]
+        lookup: ifDescr
```
brian-brazil, line 631: Do you really need both of these? Usually for a given device one is enough.

generator/generator.yml
```diff
@@ -612,3 +612,42 @@ modules:
...
+      - source_indexes: [ifIndex]
+        lookup: 1.3.6.1.2.1.31.1.1.1.1
+      - source_indexes: [mtxrInterfaceStatsIndex]
+        lookup: ifAlias
+      - source_indexes: [mtxrInterfaceStatsIndex]
+        lookup: ifDescr
+      - source_indexes: [mtxrInterfaceStatsIndex]
+        lookup: 1.3.6.1.2.1.31.1.1.1.1
```
brian-brazil, line 639: This doesn't seem right, what are you trying to do?

brian-brazil (commented):
You'll need to regenerate snmp.yml.

generator/generator.yml
carlosedp: True, cleaned up.

generator/generator.yml
carlosedp: True, replaced by ifNames.

generator/generator.yml
carlosedp: True, replaced by ifNames.

carlosedp:
I had to add unzip to the Dockerfile since it was failing when unzipping Kemp MIBs.

brian-brazil:
Hmm, a whole pile of other things are changing including losing many interface types. Is this MIB you're adding pulling in some older stuff?

carlosedp:
I believe RFC1213 was overriding some OIDs. Since I removed "ip" from my module, it's not required anymore.

snmp.yml
```diff
@@ -15736,28 +19696,36 @@ paloalto_fw:
     type: DisplayString
     help: Zone name for the interface - 1.3.6.1.4.1.25461.2.1.2.3.10.1.1
     indexes:
-    - labelname: ifIndex
```
brian-brazil, line 15739: Nothing here or below should be changing.

carlosedp:
That's weird, I've only added the Mikrotik module. Maybe it wasn't generated properly before?

I'm using Docker to generate and download MIBs:

```sh
docker build -t snmp-generator .

docker run -ti \
  -v "${PWD}:/opt/" --entrypoint make snmp-generator mibs

docker run -ti \
  -v "${PWD}:/opt/" snmp-generator generate
```

brian-brazil:
No, there's something weird going on. New indexes shouldn't be appearing on objects for paloalto_fw.

SuperQ:
Please rebase in order to pick up some fixes to the generator.

carlosedp:
Rebased. Apparently the weird generation for non-touched modules is gone.

brian-brazil:
Thanks!

PR #594, opened 2020-12-28
Fix issue where macOS would error on 'make mibs'

marri5317:
While trying to pull the MIBs using the ```make mibs``` command on macOS, it would fail with the following error:
```
sed: 1: "s/"([0-9]{12})[0-9]{2}Z ...": \1 not defined in the RE
make: *** [mibs/.kemp-lm] Error 1
```
This was fixed by properly escaping the regex in the Makefile. See also: 
https://stackoverflow.com/questions/24717676/bash-profile-sed-1-not-defined-in-the-re

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -166,7 +166,7 @@ $(MIBDIR)/.kemp-lm:
 	@curl $(CURL_OPTS) -o $(TMP) $(KEMP_LM_URL)
 	@unzip -j -d $(MIBDIR) $(TMP) *.txt
 	# Workaround invalid timestamps.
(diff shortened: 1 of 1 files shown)


brian-brazil:
This would break with a GNU sed.

Note that there's some odd issue with netsnmp on OSX that doesn't parse MIBs properly, so using a Linux box to generate is advised.

marri5317:
Should work now. Tested both on macOS and Ubuntu 20.04.1

brian-brazil:
The CI still isn't happy with that, looks like it doesn't have an effect on GNU sed.

PR #597, opened 2021-01-04
Fixes `make mibs` for BSD sed (e.g. macOS)

bnordbo:
There's a slight difference between the GNU and BSD seds; in the latter the backup file suffix for in-place replacement is mandatory. This fix sets a suffix and removes the backup files afterwards, thus making the replacement compatible with both GNU and BSD sed.


brian-brazil:
Can you add the DCO?

bnordbo:
@brian-brazil Sorry, I've added the DCO now. The CI failure does not seem to be related to this PR.

brian-brazil:
Can you rebase?

bnordbo:
@brian-brazil Done. All checks passes now.

brian-brazil:
Thanks!

PR #598, opened 2021-01-06
Sort children from netsnmp.

brian-brazil:
The data from netsnmp is non-deterministically ordered, which
was leading to the snmp.yml output also being non-deterministic.

@SuperQ


SuperQ (approved):
LGTM

PR #601, opened 2021-01-17
an optional config file and a new HTTP query string to override walk param

napw:


brian-brazil:
Thanks for your PR. This appears to be duplicating part of the the existing module functionality, which there's no need to do as you can already have multiple modules.

napw:
> Thanks for your PR. This appears to be duplicating part of the the existing module functionality, which there's no need to do as you can already have multiple modules.

multiple modules only are feasible when there are a small amount of devices or all devices having the same walk parameters. in my case, 97 network devices with all different community string and the size of final snmp.yml is almost 150MB, and most content of the file are repeats of metrics.

SuperQ:
I've proposed something similar before. I was planning to completely split the `auth` section from the `walk`/`get`/`metrics`.

Then the URL params would be something like this:
* `auth` an authentication module. (Defaults to `default`)
* `module` the walk+metrics. (Defaults to `if_mib`)
* `target` the target host. (Required)

This would provide an easier way to do per-device auth when you have a secure setup with SNMPv3.

And just to note for anyone seeing this thread. Community strings are not secrets, or passwords, or are in any way related to security. Only SNMPv3 has any security protection. You're doing nothing by having per-device v1/v2c community strings, it's window dressing. The community string is sent in plain text in every SNMP message over the wire.

brian-brazil:
> multiple modules only are feasible when there are a small amount of devices or all devices having the same walk parameters. in my case, 97 network devices with all different community string and the size of final snmp.yml is almost 150MB, and most content of the file are repeats of metrics.

I don't see anything there that'd make it infeasible, that's not a lot of RAM.

> Community strings are not secrets, or passwords, or are in any way related to security. 

In Prometheus terms, community strings are secrets and are treated as such. In just the same way that basic auth passwords are.

peterhoeg:
@SuperQ , did you manage to get anywhere with splitting out the auth bit?

SuperQ:
Closing in favor of #859

PR #608, opened 2021-02-08
Adding support for GUDE PDU devices

neredera:
Support for a rack PDUs with switching functionality, power measurement and environmental sensors.

I was only able to test it with an older device (Expert Power Control NET 8x 8210).


SuperQ:
It would be good to contact the vendor to get them to fix their MIB download link.

neredera:
> It would be good to contact the vendor to get them to fix their MIB download link.

I'm in contact with them.

There is an new (MIB)[http://demo.gude.info/epc/8210/GUDEADS-EPC8210-MIB.txt] that is not compatible with the one I used.
There is a firmware update for the device which corresponds to the new MIB according to them.  I will upgrade my device next time I have physical access (probably next week) and then update the PR.

SuperQ (commented):
I don't think we can accept this here if the vendor doesn't provide the correct MIB.

PR #624, opened 2021-03-01
Implement indices filters

sebastien-coavoux:
Hi,

This PR adds the feature of instance filtering in both the generator and the exporter. 
Filters in the generator are referred as "static" while filters in the exporter are "dynamic". 
The idea behind this, is to limit the number of oid returned by the monitored device , and in most case reduce the number of snmp get sent. 

Static filters can be seen as an ease of configuration, as it make instance filtering easier to do
Dynamic filters are simple for now in order to keep the exporter stateless between runs. Filter are evaluated when /snmp route is reached by prometheus by http. We could have a /filter route to make 2 separate flows but that would complexify the code. 

Filters are also not made to have multiple levels. Oid are targeted by a single filter, there is no "and / or" logic at all. 

However, Dynamic filter evaluation support regex. Regex matching was as complex as string comparison but with more options for the user. 

Let me know what do you think. I could squash commits / remove some if necessary

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -171,7 +171,7 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 	}
 	defer snmp.Conn.Close()
 
-	// evaluate rules
+	// Evaluate rules.
 	newGet := config.Get
 	newWalk := config.Walk
 	for _, filter := range config.Filters {
@@ -183,19 +183,21 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 		} else {
 			pdus, err = snmp.BulkWalkAll(filter.Oid)
 		}
+		// Do not try to filter anything if we had errors.
 		if err != nil {
+			level.Info(logger).Log("msg", "Error getting OID, won't do any filter on this oid", "oid", filter.Oid)
 			continue
 		}
 
 		allowedList = filterAllowedIndices(logger, filter, pdus, allowedList)
 
-		// Update config to get only index and not walk them
+		// Update config to get only index and not walk them.
 		newWalk = updateWalkConfig(newWalk, filter, logger)
 
-		// Only Keep indices not involved in filters
+		// Only Keep indices not involved in filters.
 		newCfg := updateGetConfig(newGet, filter, logger)
 
-		// We now add each index from filter to the get list
+		// We now add each index from filter to the get list.
 		newCfg = addAllowedIndices(filter, allowedList, logger, newCfg)
 
 		newGet = newCfg
@@ -300,7 +302,7 @@ func updateWalkConfig(walkConfig []string, filter config.DynamicFilter, logger l
 				break
 			}
 		}
-		// Oid not found in target,  we walk it
+		// Oid not found in target,  we walk it.
 		if !found {
 			newCfg = append(newCfg, elem)
 		}
@@ -313,14 +315,12 @@ func updateGetConfig(getConfig []string, filter config.DynamicFilter, logger log
 	for _, elem := range getConfig {
 		found := false
 		for _, targetOid := range filter.Targets {
-			if !strings.HasPrefix(elem, targetOid) {
-				continue
-			} else {
+			if strings.HasPrefix(elem, targetOid) {
 				found = true
 				break
 			}
 		}
-		// Oid not found in targets, we keep it
+		// Oid not found in targets, we keep it.
 		if !found {
 			level.Debug(logger).Log("msg", "Keeping get configuration", "oid", elem)
 			newCfg = append(newCfg, elem)
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -18,6 +18,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -170,7 +171,37 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 	}
 	defer snmp.Conn.Close()
 
-	getOids := config.Get
+	// evaluate rules
+	newGet := config.Get
+	newWalk := config.Walk
+	for _, filter := range config.Filters{
+		var pdus []gosnmp.SnmpPDU
+		allowedList := []string{}
+
+		if snmp.Version == gosnmp.Version1 {
+			pdus, err = snmp.WalkAll(filter.Oid)
+		} else {
+			pdus, err = snmp.BulkWalkAll(filter.Oid)
+		}
+		if err != nil {
+			continue
+		}
+
+		allowedList = filterAllowedInstances(logger, filter, pdus, allowedList)
+
+		// Update config to get only instance and not walk them
+		newWalk = updateWalkConfig(newWalk, filter, logger)
+
+		// Only Keep instance not involved in filters
+		newCfg := updateGetConfig(newGet, filter, logger)
+
+		// We now add each instance from filter to the get list
+		newCfg = addAllowedInstances(filter, allowedList, logger, newCfg)
+
+		newGet = newCfg
+	}
+
+	getOids := newGet
 	maxOids := int(config.WalkParams.MaxRepetitions)
 	// Max Repetition can be 0, maxOids cannot. SNMPv1 can only report one OID error per call.
 	if maxOids == 0 || snmp.Version == gosnmp.Version1 {
@@ -213,7 +244,7 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 		getOids = getOids[oids:]
 	}
 
-	for _, subtree := range config.Walk {
+	for _, subtree := range newWalk {
 		var pdus []gosnmp.SnmpPDU
 		level.Debug(logger).Log("msg", "Walking subtree", "oid", subtree)
 		walkStart := time.Now()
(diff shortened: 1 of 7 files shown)


SuperQ:
Nice, it will take some time to review this.

The first thing I am thinking about here is terminology. We use the term "instance" in Prometheus to refer to the target device.

I need to dig through various MIBs to see how the terminology is used there, there are some references to "INDEX", and some "instance".

From a high-level perspective, this is an "Index filter". It filters instances of results from the OID index(es). I think it would be more intuitive to align the naming in this way.

suckerfish:
Hi @sebastien-coavoux I'm having trouble getting the filtering to work. 

Can you look at this brief snip from the bottom of my generated snmp.yml and let me know what might not be working here? Per my understanding, this very short example should enable the exporter to only scrape interfaces whose descriptions start with 'DIA' and report the interface speed of those interfaces. Does that read right to you, as well? 

`filters:
  - oid: 1.3.6.1.2.1.31.1.1.1.18
    targets:
    - 1.3.6.1.2.1.2.2.1.5
    values:
    - ^DIA.*`

sebastien-coavoux:
Hey, any update on it? I gpg signed the commit but now the generator fails because it is not finding some OID (raritan). 

Pretty strange since I have touched the mib dir neither the generator yaml.

SuperQ:
The generator thing happens when upstream MIBs cause issues.

https://github.com/prometheus/snmp_exporter/pull/647 should fix that up.

sebastien-coavoux:
Hi,

I've rebased the MR on main so that the CI is green. I've also changed the vocab from instance to index/indices ( I had to choose a side for the [plural](https://grammarist.com/usage/indexes-indices/), feel free to argue on it :smile: ) 

Let me know if you would like something else.

xkilian:
Been running with this for a while, over 800 switches/routers/firewalls with various filters. (Admin up, interface names and interface descriptions) No problems to report.

SuperQ:
Sorry for the long delay in reviewing. Would you mind rebasing this?

sebastien-coavoux:
Done !

config/config.go
```diff
@@ -187,6 +188,21 @@ func (c WalkParams) ConfigureSNMP(g *gosnmp.GoSNMP) {
...
 }
 
+type Filters struct {
+	Static  []StaticFilter  `yaml:"static,omitempty"`
+	Dynamic []DynamicFilter `yaml:"dynamic,omitempty"`
+}
+
+type StaticFilter struct {
+	Targets []string `yaml:"targets,omitempty"`
+	Indices []string `yaml:"indices,omitempty"`
```
SuperQ, line 198: I see why this is a slice of string, but, maybe it would be worth doing this as a `[]uint` and doing bounds checking to make sure only a single byte worth of decimal integer is passed?

:thinking:

SuperQ:
This seems mostly sane, one question about the input config. Otherwise I think we can add this.

config/config.go
sebastien-coavoux: I started with an int at the beginning then switch to string to have the concatenation part esier with the oid. 

Would you prefer to switch it to uint or simply try once to `strconv.Atoi` the value, in UnmarshalYAML for example ?

SuperQ: I don't have a strong opinion here. But it would be good to have some kind of input validation.

sebastien-coavoux:
@SuperQ sorry for the delay, I was mostly off in summer. I've just added the index validation as part of Unmarshalling.

PrestonTaylor:
This change is very well thought out, it worked flawlessly for my setup and accounted for all my whacky corner cases. It cut my time for stat collection down from 30s per router to 1.5!!!!! That's huge savings. Obviously it also cut my storage requirements greatly as well. 

My only suggestion would be to make the example config in README.md a working one (filter targets should match oids listed in walk if I'm understanding correctly).

sebastien-coavoux:
@PrestonTaylor , is there anything not clear in https://github.com/prometheus/snmp_exporter/pull/624/files#diff-a715bdcf88cd771d7ae09bd01863bab314ff194ba28e984689e9040d87846db9R147 ?  Readme in generator provides an example, maybe you meant the Readme at the top level of the repo ?

PrestonTaylor:
@sebastien-coavoux My suggestion was in your readme you've linked to add the oids you have in your target section (1.3.6.1.2.1.2.2.1.4 and bsnDot11EssSsid) to the walk section on line 61 of your change and 54 of the original. This was required in my case for it to output the metrics, without it it would query them according to the debug but not output them. Very minor change that most people would figure out anyway I'm sure. Git doesn't let me add a code suggestion more than 3 lines from your edits or I would do so.

sebastien-coavoux:
@PrestonTaylor Ok I got it. I have added the 2 oid, the example provided is accurate with it. Thanks !

PrestonTaylor:
Any chance of this making it into a release? Been using it at scale without issues but would love to see it merged.

SuperQ:
Yes, It's on my todo list. Just a lot of change to review.

collector/collector.go
```diff
@@ -213,7 +244,7 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
 		getOids = getOids[oids:]
 	}
 
-	for _, subtree := range config.Walk {
+	for _, subtree := range newWalk {
 		var pdus []gosnmp.SnmpPDU
```
beorn7, line 248: I see, it has been done the other way here already.

Not sure about the snmp_exporter, but generally, we prefer the `pdus := []gosnmp.SnmpPDU{}` in the Prometheus codebase.

collector/collector.go
```diff
@@ -235,6 +266,79 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
...
+	newCfg := []string{}
+	for _, elem := range getConfig {
+		found := false
+		for _, targetOid := range filter.Targets {
+			if !strings.HasPrefix(elem, targetOid) {
+				continue
+			} else {
+				found = true
+				break
+			}
```
beorn7, line 321: Couldn't this just be written as
```Go
			if strings.HasPrefix(elem, targetOid) {
				found = true
				break
			}
```
?

config/config.go
beorn7: No strong opinion, either, just _some_ validation, as @SuperQ said.

beorn7 (changes requested):
Just minor Go style comments.

I have just looked at the shape of the Go code and didn't check correctness (assuming that was done by others).

Would adding tests be feasible?

collector/collector.go
```diff
@@ -170,7 +171,37 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
...
+		var pdus []gosnmp.SnmpPDU
+		allowedList := []string{}
+
+		if snmp.Version == gosnmp.Version1 {
+			pdus, err = snmp.WalkAll(filter.Oid)
+		} else {
+			pdus, err = snmp.BulkWalkAll(filter.Oid)
+		}
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
sebastien-coavoux: See updated diff

collector/collector.go
sebastien-coavoux: I am not confident to be able to switch to `[]gosnmp.SnmpPDU{}` within this PR without screwing up a bunch of thing. 

I see that `ScrapeResults` struct has this type so the idea would be to update it and all code using it. It seems like there is not much of it but maybe a followup PR would be safer.

config/config.go
sebastien-coavoux: Implemented in `UnmarshalYAML` [below](https://github.com/prometheus/snmp_exporter/pull/624/files#diff-d2187742eb30a92ec406b45dcd849d5db523031d682af72683892e83dd6c5f0fR57)

sebastien-coavoux:
Updated PR from the input. Also rebased on main

SuperQ:
Thanks @sebastien-coavoux, did you see #855?

sebastien-coavoux:
I haven't dive into it but it looks like that the panic is not coming from the code modified by the PR.
`Mar 20 18:56:07 myexporterserver snmp_exporter: panic: Unknown index type Bits` comes from https://github.com/prometheus/snmp_exporter/blob/main/collector/collector.go#L747

The minimal case seems not to have filter to reproduce so I feel like it might be possible to reproduce it on main branch

I could have a look tonight, I guess the provided MIB in there would help reproduce.

collector/collector.go
beorn7: Nothing would be screwed up, but I see now how the variable is used, which is indeed a flow with which the existing syntax has a (theoretical) performance benefit. So I retract my concern.

beorn7 (commented):
LGTM from the Go side.

sebastien-coavoux:
> Just minor Go style comments.
> 
> I have just looked at the shape of the Go code and didn't check correctness (assuming that was done by others).
> 
> Would adding tests be feasible?

Just noticed this, added some tests. Unless there is other comment I think I am done for that PR :+1:

SuperQ (approved):

collector/collector.go
```diff
@@ -300,7 +302,7 @@ func updateWalkConfig(walkConfig []string, filter config.DynamicFilter, logger l
 				break
 			}
 		}
-		// Oid not found in target,  we walk it
+		// Oid not found in target,  we walk it.
```
beorn7, line 305: Nit: Spurious blank (`,  we`).

collector/collector.go
```diff
@@ -235,6 +268,77 @@ func ScrapeTarget(ctx context.Context, target string, config *config.Module, log
...
+	for _, elem := range walkConfig {
+		found := false
+		for _, targetOid := range filter.Targets {
+			if elem == targetOid {
+				level.Debug(logger).Log("msg", "Deleting for walk configuration", "oid", targetOid)
+				found = true
+				break
+			}
+		}
+		// Oid not found in target,  we walk it.
```
SuperQ, line 305:
```suggestion
		// Oid not found in target, we walk it.
```

collector/collector.go
SuperQ: Eh, I'll clean that up later.

JDA88:
From my understanding with this change it is now possible to use dynamic filters to only get interfaces that have `ifAlias` that match a RegEx but I struggle to find the correct syntax, does anyone have an example?

sebastien-coavoux:
The readme should provide a good example of it. If not then maybe it is not clear enough so you could suggest a PR

From the snippet of readme we have 

```
      dynamic: # dynamic filters are handed by the snmp exporter. The generator will simply pass on the configuration in the snmp.yml.
               # The exporter will do a snmp walk of the oid and will restrict snmp walk made on the targets
               # to the index matching the value in the values list.
               # This would be typically used to specify a filter for interfaces with a certain name in ifAlias, ifSpeed or admin status.
               # For example, only get interfaces that a gig and faster, or get interfaces that are named Up or interfaces that are admin Up
        - oid: 1.3.6.1.2.1.2.2.1.7
          targets:
            - "1.3.6.1.2.1.2.2.1.4"
          values: ["1", "2"]
```

This means that the exporter will walk 1.3.6.1.2.1.2.2.1.7 and for each index of this oid, if the values match "1" or "2" it will remember this index and do a snmp get of this index on the target .

example :  `1.3.6.1.2.1.2.2.1.7.1` => `111` . The exporter will get `1.3.6.1.2.1.2.2.1.4.1`

JDA88:
I saw and understand the example but I fail to extrapolate It to my usercase.
Let say I want the: 
`ifAdminStatus` (`1.3.6.1.2.1.2.2.1.7`) if the `ifName` (`1.3.6.1.2.1.31.1.1.1.1`) match the regex `^SW-.*`
Will it translate to the following code?

```
      dynamic: 
        - oid: 1.3.6.1.2.1.31.1.1.1.1
          targets:
            - "1.3.6.1.2.1.2.2.1.7"
          values: ["^SW-.*"]
```

sebastien-coavoux:
Yes this would be the right snippet of config. 

If it is not working for you i would suggest to open an in issue. In the meantime things to check (if it is not working) : 
* Run exporter in Debug, you should see in the logs the rules / filter evaluated
* Snmpwalk `1.3.6.1.2.1.31.1.1.1.1` manually and look at the output. Check if the regexp apply to that output : https://play.golang.com/p/hU-XCVNgEd5

PR #616, opened 2021-02-21
Split collector into separate package

SuperQ:
Make it easier to embed the snmp_exporter into other systems.
* Update metric registration to use promauto.
* Update minimum version to Go 1.14. (required for exporter-toolkit)
* Update snmp.yml with latest mibs.

Re-implementation of https://github.com/prometheus/snmp_exporter/pull/456

Signed-off-by: Ben Kochie <superq@gmail.com>

PR #634, opened 2021-03-25
collector: pass Context

SaveTheRbtz:
This diff adds an ability to pass custom Context to the collector.  This is inline with some other exporter's like the [mysqld_exporter](https://github.com/prometheus/mysqld_exporter/blob/3381d9c187b54115e89764de20f5c43bad948ad0/collector/exporter.go#L87).

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -208,8 +208,8 @@ type collector struct {
 	logger log.Logger
 }
 
-func New(target string, module *config.Module, logger log.Logger) *collector {
-	return &collector{target: target, module: module, logger: logger}
+func New(ctx context.Context, target string, module *config.Module, logger log.Logger) *collector {
+	return &collector{ctx: ctx, target: target, module: module, logger: logger}
 }
 
 // Describe implements Prometheus.Collector.
--- a/main.go
+++ b/main.go
@@ -99,7 +99,7 @@ func handler(w http.ResponseWriter, r *http.Request, logger log.Logger) {
 
 	start := time.Now()
 	registry := prometheus.NewRegistry()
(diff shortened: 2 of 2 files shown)


SuperQ:
Ahh, yes, it would be useful use the `X-Prometheus-Scrape-Timeout-Seconds` header.

I'm not sure how useful it is to stub this out, without implementing any use of the context is. Are you planning to add more to this PR?

SaveTheRbtz:
We use collector directly as a library so we pass our own context therefore I marked the main.go context as TODO.  I you want I can add support for `X-Prometheus-Scrape-Timeout-Seconds` in this diff, or a subsequent one.

SuperQ:
Yes, with normal use, no context is necessary, as the scrape drops the HTTP connection on timeout, which cancels the walk. I was wrong, we don't need to explicitly handle the `X-Prometheus-Scrape-Timeout-Seconds` here. It's automatic.

SaveTheRbtz:
Ok. I've added context propagation from the original HTTP request to the underlying collector.  This should ensure cancelation in [following scenarios](https://golang.org/pkg/net/http/#Request.Context):

> For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns.

SuperQ:
Thanks, it looks like I screwed up the context [here](https://github.com/prometheus/snmp_exporter/pull/616/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R102). This puts it back to where it was.

SuperQ (approved):
LGTM.

Issue #635, opened 2021-03-29
Can't build a Docker image of the generator


bondif:
### Host operating system: output of `uname -a`
Linux monitor 5.4.0-66-generic # 74-Ubuntu SMP Wed Jan 27 22:54:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### What did you do that produced an error?
I was following the steps in README file for building the generator image, so I've run `docker build -t snmp-generator .`    
I didn't change the Dockerfile.

### What did you expect to see?
A new image got generated.

### What did you see instead?
```
...
go: downloading github.com/prometheus/snmp_exporter v0.20.0
go: downloading github.com/go-kit/kit v0.10.0
go: downloading github.com/prometheus/common v0.15.0
go: downloading gopkg.in/alecthomas/kingpin.v2 v2.2.6
go: downloading gopkg.in/yaml.v2 v2.4.0
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/gosnmp/gosnmp v1.29.0
go: downloading github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
go: downloading github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d
go: downloading github.com/go-logfmt/logfmt v0.5.0
/bin/sh: 1: cd: can't cd to /go/src/github.com/prometheus/snmp_exporter/generator
The command '/bin/sh -c apt-get update &&     apt-get install -y libsnmp-dev p7zip-full unzip &&     go get github.com/prometheus/snmp_exporter/generator &&     cd /go/src/github.com/prometheus/snmp_exporter/generator &&     go get -v . &&     go install' returned a non-zero code: 2
```

Any thoughts please?

SuperQ:
This might have to do with the changes to how Go 1.16 handles `go get`.

bondif:
Do you suggest any version please? I'm not a Go developer

PR #636, opened 2021-03-29
Update Dockerfile for Go 1.16

SuperQ:
Use go install to build the generator.

Fixes: https://github.com/prometheus/snmp_exporter/issues/635


RichiH (approved):
Why does this change snmp.yml?

Else, LGTM

SuperQ:
Because between the last update, some MIBs changed. Since we don't vendor the MIBs ourselves.

PR #638, opened 2021-04-04
generator: add a --config-file= option

bootc:
This allows easier use of multiple generator config files, for example
for different sets of authentication credentials or to break up the
giant snmp.yml file into smaller chunks that fit better in a Kubernetes
ConfigMap.

### DIFF
--- a/generator/main.go
+++ b/generator/main.go
@@ -37,7 +37,12 @@ func generateConfig(nodes *Node, nameToNode map[string]*Node, logger log.Logger)
 		return fmt.Errorf("unable to determine absolute path for output")
 	}
 
-	content, err := ioutil.ReadFile("generator.yml")
+	configPath, err := filepath.Abs(*configPath)
+	if err != nil {
+		return fmt.Errorf("unable to determine absolute path for configuration file")
+	}
+
+	content, err := ioutil.ReadFile(configPath)
 	if err != nil {
 		return fmt.Errorf("error reading yml config: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


RichiH:
This would be superseded by https://docs.google.com/document/d/1BK_Gc3ixoWyxr9F5qGC07HEcfDPtb6z96mfqoGyz52Y . If you feel strongly, we could add it as `--experimental-config-file` to make clear it will go away; hopefully we will see what's in the doc within a month, though.

bootc:
No I don't feel strongly about it, and it can wait to be solved properly. It just felt wrong to apply this locally without offering it up more widely.

xkilian:
The work being done on other PRs is in fact to split the processing of each family of devices associated with a generator.yml file and its associated snmp mibs as distinct independant blobs. 

We all wat to move away from a big generator.yml, with a massive directory of conflicting snmp mibs. 

By having a structure to handle distinct device families we can deliver a bunch of mini generator_family.yml, their own mibs directory containing only the required mibs require and have their output consolidated in a destination snmp.d directory. The mibs directory are not distributed with the Exporter, but would be a listing of minimum required mibs as is done today.

It will be the collectors job to join/process the different snmp_family.yml files together for running the collector.

xkilian:
This issue is still a valid concern all the more as we add scalability features to handle networking gear. We have used this for a couple years and have seen great value in sperating each device family in its own directory and specific generator yml file. Another specific way of dealing with it can be raised, but the monolithic generator file is not the answer.

generator/main.go
```diff
@@ -95,6 +100,7 @@ func generateConfig(nodes *Node, nameToNode map[string]*Node, logger log.Logger)
 
 var (
 	failOnParseErrors  = kingpin.Flag("fail-on-parse-errors", "Exit with a non-zero status if there are MIB parsing errors").Default("false").Bool()
+	configPath         = kingpin.Flag("config-file", "Path to configuration file").Default("generator.yml").Short('c').String()
```
SuperQ, line 103: We generally don't do short flags for things. We also prefix things so it's more clear what the flag is for.

```suggestion
	configPath         = kingpin.Flag("generator.config.file", "Path to configuration file").Default("generator.yml").String()
```

SuperQ (changes requested):

SuperQ (commented):
Any chance you want to finish this?

SuperQ:
Superseded by #1028

Issue #104, opened 2017-01-06
[Discussion] Add 'snmp_' prefix to metrics?


seanhoughton:
The first item in the exporter metric naming best practices list is

> A metric name ... should have a (single-word) application prefix relevant to the domain the metric belongs to.
- https://prometheus.io/docs/practices/naming/

None of the metric names in the current file adhere to this practice, it probably makes sense to just automatically prefix each metric with `snmp_`

brian-brazil:
In the case of SNMP, all the names have already been well-established over decades and follow another naming scheme. Adding snmp_ would likely cause more harm then good.

RichiH:
I agree, the naming of SNMP is too well-established.

Richard

Sent by mobile; excuse my brevity.

SuperQ:
Correct, the individual metrics have long-established names, but by Prometheus policy these should be grouped into their domain.  We shouldn't modify the metric name, simply prepend them.

`snmp_ifHCInOctets` for example makes it much easier to match when doing things like federation.

brian-brazil:
It's a guideline not policy, and SNMP is the canonical example of where the balance tips the other way.

You wouldn't be federating names like that, you'd be federating on the aggregation prefix.

seanhoughton:
I'll leave it up to you guys to decide, but the curse of "sure it doesn't follow convention and makes things more difficult, but that's the way we've always done it" has caused untold pain in my professional career. All of the other exporters provide a very clear prefix that makes browsing the available metrics for the exporter you're concerned with very easy. The snmp exporter is the only one in the bunch that scatters metrics all around because of the lack of a namespacing prefix. However, the nice thing about open source is that I can make a fork with the prefix added and go about my business :)

brian-brazil:
> All of the other exporters provide a very clear prefix that makes browsing the available metrics for the exporter you're concerned with very easy.

I mentioned this when this previously came up, but this is not true in general. It's normal for there to be quite a bit of overlap in the metrics between all of your applications, so this is not something you can depend on.

RichiH:
For the record, after some more deliberations, I see prefixing `snmp_` as a valid request. Having on shared namespace helps avoid confusion down the line.

Either way, this should either be implemented, or the issue closed.

SuperQ:
I'm thinking about adding an optional prefix config item to the generator.  That would solve it for everyone.

brian-brazil:
This should not be a configurable option, a given metric should only have one name.

[...]

PR #639, opened 2021-04-08
adding metrixPrefix param

julian3xl:
On the prometheus world we gather tons of metrics and we have to trust on the collectors to set proper naming to avoid clashes. In the case of snmp (a very old technology), we have the real risk of getting same naming from different technologies. Because the nature of this exporter, it's not the own machine that we're measuring who sends its metrics, it's an external scrapper, so prometheus metadata doesn't give you context about who and where is that machine. This is why mixing metrics from different sources and same name could be a mess.

Regarding this and because it's a real requested feature (https://github.com/prometheus/snmp_exporter/issues/104), this is my approach to achieve it.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -202,14 +202,15 @@ func buildMetricTree(metrics []*config.Metric) *MetricNode {
 }
 
 type collector struct {
-	ctx    context.Context
-	target string
-	module *config.Module
-	logger log.Logger
+	ctx           context.Context
+	target        string
+	module        *config.Module
+	logger        log.Logger
+	metricsPrefix string
 }
 
-func New(ctx context.Context, target string, module *config.Module, logger log.Logger) *collector {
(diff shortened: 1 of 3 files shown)


SuperQ:
As I said in the issue, I would like to solve the general issue with a relabel config.

julian3xl:
> As I said in the issue, I would like to solve the general issue with a relabel config.

do you mean relabelling with prometheus?

RichiH:
What if different teams run different exporters? Doing it at ingestion-time means everyone maintaining a Prometheus instance can do it consistently within their system.

julian3xl:
just for the record, some context:

prometheus-node-exporter, prefix node_
prometheus-nvidia-exporter, prefix nvidia_
nginx-ingress metrics, prefix nginx_ingress_
...and so on

SuperQ:
No, I mean relabeling in the exporter. Yes, it's possible to do this on the Prometheus side, but it would be useful to have things filtered before exposition.

PR #653, opened 2021-05-02
Add packet statistics

SuperQ:
Add per-SNMP packet statistics.
* Update Go modules to latest.
* Add exporter-wide metrics for sent packets and retries.
* Add per-scrape metrics for packets and retries.

The duplication betweeen the per-scrape and exporter-wide metrics is
intentional, so that packet histogram data can be gathered. This is to
keep the cardinality of the histogram data sane.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -32,11 +32,39 @@ import (
 	"github.com/prometheus/snmp_exporter/config"
 )
 
+const (
+	namespace = "snmp"
+)
+
 var (
+	buckets               = prometheus.ExponentialBuckets(0.0001, 2, 15)
 	snmpUnexpectedPduType = promauto.NewCounter(
 		prometheus.CounterOpts{
-			Name: "snmp_unexpected_pdu_type_total",
-			Help: "Unexpected Go types in a PDU.",
+			Namespace: namespace,
+			Name:      "unexpected_pdu_type_total",
+			Help:      "Unexpected Go types in a PDU.",
+		},
+	)
+	snmpDuration = promauto.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "packet_duration_seconds",
+			Help:      "A histogram of latencies for SNMP packets.",
+			Buckets:   buckets,
+		},
+	)
+	snmpPackets = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "packets_total",
+			Help:      "Number of SNMP packet sent, including retries.",
+		},
+	)
+	snmpRetries = promauto.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "packet_retries_total",
+			Help:      "Number of SNMP packet retries.",
 		},
 	)
 	// 64-bit float mantissa: https://en.wikipedia.org/wiki/Double-precision_floating-point_format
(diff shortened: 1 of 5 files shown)


collector/collector.go
```diff
@@ -230,12 +271,20 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
 		prometheus.NewDesc("snmp_scrape_walk_duration_seconds", "Time SNMP walk/bulkwalk took.", nil, nil),
 		prometheus.GaugeValue,
 		time.Since(start).Seconds())
+	ch <- prometheus.MustNewConstMetric(
+		prometheus.NewDesc("snmp_scrape_packets_sent", "Packets sent for the walk.", nil, nil),
```
RichiH, line 275: It's not clear from the help text if you're counting the retries.

RichiH (changes requested):

collector/collector.go
SuperQ: Hrm, good question, I would have to check the library code to see if that's included as well.

SuperQ: I reviewed the SNMP library, and retries are included in packets sent.

collector/collector.go
```diff
@@ -230,12 +279,20 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
 		prometheus.NewDesc("snmp_scrape_walk_duration_seconds", "Time SNMP walk/bulkwalk took.", nil, nil),
 		prometheus.GaugeValue,
 		time.Since(start).Seconds())
+	ch <- prometheus.MustNewConstMetric(
+		prometheus.NewDesc("snmp_scrape_packets_sent", "Packets sent for the walk, including retries.", nil, nil),
```
RichiH, line 283:
```suggestion
		prometheus.NewDesc("snmp_scrape_packets_sent", "Packets sent for get, bulkget, and walk; including retries.", nil, nil),
```

collector/collector.go
```diff
@@ -230,12 +279,20 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
...
+		prometheus.GaugeValue,
+		float64(results.packets))
+	ch <- prometheus.MustNewConstMetric(
+		prometheus.NewDesc("snmp_scrape_packets_retried", "Packets retried for the walk.", nil, nil),
```
RichiH, line 287:
```suggestion
		prometheus.NewDesc("snmp_scrape_packets_retried", "Packets retried for get, bulkget, and walk.", nil, nil),
```

RichiH (approved):
LGTM; please see suggestions.

RichiH (approved):

SuperQ:
The generator error is unrelated, I'm going to remove it from the required list for now, since we're having issues with the IANA MIB download again.

config/config.go
```diff
@@ -63,7 +63,7 @@ type Config map[string]*Module
 
 type WalkParams struct {
 	Version        int           `yaml:"version,omitempty"`
-	MaxRepetitions uint8         `yaml:"max_repetitions,omitempty"`
+	MaxRepetitions uint32        `yaml:"max_repetitions,omitempty"`
```
luizluca, line 66: FYI, before this commit, max_repetitions was stored as an uint8. If max_repetitions is set to anything greater than 127, it was encoded (ASN.1/DER) as "02 01 80", which, AFAIK, really means -128, breaking the request. 128 should be "02 02 00 80" as there is no unsigned integers for DER. I would double-check the SNMP library if it is correctly encoding unsigned values as it might break other scenarios.

I think that, with uint32, it is virtually impossible to actually use anything greater than 2^31. I didn't test this patch and I'll wait for the next release. Anyway, it would be nice to duble check this.

SuperQ: This change was related to an upstream gosnmp change.

https://github.com/gosnmp/gosnmp/pull/295.

PR #657, opened 2021-05-22
Fix default snmp.yaml

sebastien-coavoux:
Circle is failing because of the update in mibs that change the default snmp file

Signed-off-by: Sébastien Coavoux <sebastien.coavoux@azoplee.com>


SuperQ (approved):
LGTM, thanks

PR #679, opened 2021-08-18
Add net-snmp IP-MIB

ibigbug:
To walk https://oidref.com/1.3.6.1.2.1.4.20

This can be useful when walking IP address of interfaces


ibigbug:
ping

SuperQ:
We typically only add new MIBs when we need them for the generator.yml. Would you mind adding a config to it?

For example, the OID you mentioned is deprecated, but `ipv4InterfaceTable` might be one to add.

ibigbug:
added `ipv4InterfaceTable` to `IF_MIB`

generator/generator.yml
```diff
@@ -1,7 +1,7 @@
 modules:
   # Default IF-MIB interfaces table with ifIndex.
   if_mib:
-    walk: [sysUpTime, interfaces, ifXTable]
+    walk: [sysUpTime, interfaces, ifXTable, ipv4InterfaceTable]
```
SuperQ, line 4: Please add this as a separate module named `ip_mib`.

SuperQ (changes requested):

SuperQ:
You will need to run the generator to update the `snmp.yml` example in the repo.

ibigbug:
not sure `fixed_size: 6` is getting removed, same thing in https://github.com/prometheus/snmp_exporter/pull/681/files

snmp.yml
```diff
@@ -4636,7 +4636,6 @@ cisco_wlc:
     indexes:
     - labelname: bsnAPDot3MacAddress
       type: PhysAddress48
-      fixed_size: 6
```
SuperQ, line 4639: It looks like your generator or mibs are out of date. These should not be removed.

ibigbug: not sure - I've rebased the latest main branch of this repo and the mibs were just generated with make mibs.

any ideas?

SuperQ:
This still needs to be fixed up.

ibigbug:
ty @SuperQ the generator is working now

SuperQ (approved):

PR #685, opened 2021-10-01
Allow snmp_exporter to accept OIDs that aren't in an increasing order

mjsaarin:
Some SNMP agents return, when receiving walk/bulkwalk requests, OIDs in an order that isn't strictly increasing order. This patch adds an option that instructs GoSNMP not to discard such replies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/prometheus/snmp_exporter
 
 require (
 	github.com/go-kit/kit v0.10.0
-	github.com/gosnmp/gosnmp v1.31.0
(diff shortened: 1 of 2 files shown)


RichiH:
Thanks for this.

Please remember to sign the DCO.

For reference, which devices did you find that exhibit this?

I wonder if it wouldn't be better to make this a config option per module as opposed to a startup-time one.

mjsaarin:
I've observed this at least with Juniper QFX devices, PulseSecure appliances and some version of Cisco WLC. 

As I'm almost clueless with Github I don't know how to sign DCO. I tried some hints I found but all resulted in error messages I didn't understand.

PR #686, opened 2021-10-01
Added an option that instructs GoSNMP to accept OIDs in non-increasing order

mjsaarin:
When receiving walkwalk requests some devices, like Juniper QFX switches, return OIDs in an order that isn't strictly increasing. This modification adds a command line option that calls GoSNMP with c flag enabled.

Signed-off-by: Matti Saarinen <mjsaarin@cc.helsinki.fi>


SuperQ:
This seems like a reasonable idea, would you mind updating this branch to latest main?

It would also be nice to add this feature to the generator.

mjsaarin:
I hope the branch is now updated. Also, the generator can create configs with this option if it is given in generate.yml.  Sorry about the mess and extra commits with syntax issues.

I've built the collector and generator on Debian 11 (Bullseye) and on RHEL8. Both appear to work.

PR #702, opened 2021-11-11
Support equipment responding from different source address

### LINKED ISSUE #699: Support equipment responding from different source address
Some equipment responds using a different source address than the address SNMP request was sent to.

Such equipment can currently not be monitored using snmp_exporter (which fails with `request timeout (after 3 retries)`).

The required functionality appears to already be present in gosnmp (controlled by the `UseUnconnectedUDPSocket` option), but is not used/exposed by snmp_exporter right now.

Could support for this be added, and if so, how? (per-scrape request parameter/command line flag/always enabled)

tobbez:
Closes #699.


tobbez:
The CI failure seems to be unrelated to the changes in this PR.

SuperQ (approved):
LGTM, thanks.

I fixed the issue with the test, will merge this.

PR #704, opened 2021-11-14
Hacky workaround for net-snmp version detection

SuperQ:
Use defines that appear in different verions of net-snmp headers.

### DIFF
--- a/generator/net_snmp.go
+++ b/generator/net_snmp.go
@@ -18,10 +18,21 @@ package main
 #cgo CFLAGS: -I/usr/local/include
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/mib_api.h>
+#include <net-snmp/agent/agent_callbacks.h>
+#include <net-snmp/library/default_store.h>
 #include <unistd.h>
 // From parse.c
+// Hacky workarounds to detect which version of net-snmp this
+// based on various headers that appear in 5.8 and 5.9.
+#if !defined(NETSNMP_DS_LIB_ADD_FORWARDER_INFO)
+#if defined(SNMPD_CALLBACK_UNREGISTER_NOTIFICATIONS)
+#define MAXTC   16384
+#else
 #define MAXTC   4096
-  struct tc {
+#endif
+#endif
+
+struct tc {
   int             type;
   int             modid;
   char           *descriptor;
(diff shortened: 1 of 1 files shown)


SuperQ:
@tobbez I came up with this terrible workaround.

generator/net_snmp.go
```diff
@@ -18,10 +18,23 @@ package main
 #cgo CFLAGS: -I/usr/local/include
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/mib_api.h>
+#include <net-snmp/types.h>
+#include <net-snmp/library/container.h>
+#include <net-snmp/agent/snmp_agent.h>
+#include <net-snmp/agent/snmp_vars.h>
```
tobbez, line 24: These headers don't seem to be required (tested and works correctly without them on 5.7.3, 5.8, 5.9.1).

tobbez (commented):
Looks good overall (considering what we have to work with).

Maybe add a comment in each branch stating which version it's for? It's not obvious from the code itself, so that would save the next person reading this code the trouble of cross-referencing the net-snmp repository.

Could also remove some duplication by making only the last part of `tclist` conditional, i.e:

```c
#if !defined(NETSNMP_DS_LIB_ADD_FORWARDER_INFO)
#if defined(SNMPD_CALLBACK_UNREGISTER_NOTIFICATIONS)
#define MAXTC   16384
#else
#define MAXTC   4096
#endif
#endif

struct tc {
  int             type;
  int             modid;
  char           *descriptor;
  char           *hint;
  struct enum_list *enums;
  struct range_list *ranges;
  char           *description;
#if !defined(NETSNMP_DS_LIB_ADD_FORWARDER_INFO)
} tclist[MAXTC];
int tc_alloc = MAXTC;
#else
} *tclist;
int tc_alloc;
#endif
```

generator/net_snmp.go
SuperQ: I had compile errors from Go if I removed them.

tobbez: Weird. What OS?

Just tested on multiple distributions (Debian/Ubuntu/Fedora/Arch/Gentoo) and it works on all of them without including the four headers mentioned above.

SuperQ:
Ok, generator test now pass. Mixin is still broken, but I can take care of that separately.

SuperQ:
Ping @RichiH

RichiH (approved):
This looks horrible. I like it.

PR #705, opened 2021-11-15
Fix broken link

PierreF:
Current URL to Liebert/Vertiv MIB respond by a 404.

I've found another URL which seems to provide the same file: https://www.vertiv.com/en-us/support/software-download/monitoring/management-information-bases-mibs-for-liebert-products/

Since I still have the old file (from previous URL), I can say that the content of the two files (LIEBERT_GP_PDU.MIB LIEBERT_GP_REG.MIB) is the same with the exception to newline (CR-LF vs LF). But this don't seems to impact the resulting snmp.yml.

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -43,7 +43,7 @@ UBNT_DL_URL       := https://dl.ubnt-ut.com/snmp
 WIENER_URL        := https://file.wiener-d.com/software/net-snmp/WIENER-CRATE-MIB-5704.zip
 RARITAN_URL       := https://cdn.raritan.com/download/PX/v1.5.20/PDU-MIB.txt
 INFRAPOWER_URL    := https://www.austin-hughes.com/support/software/infrapower/IPD-MIB.7z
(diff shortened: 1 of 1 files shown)


PierreF:
I don't have the snmp.yml generation difference. It's the same on my box. Likely because I generate the snmp.yml on Ubuntu 20.04 while CI use Debian 11.

This issue already occur on parent commit: https://app.circleci.com/pipelines/github/prometheus/snmp_exporter/658/workflows/ab543a91-a064-4395-a8b6-0c365dc236f9/jobs/2993

Do you want me to generate the snmp.yml on Debian 11 and commit it ?

SuperQ:
I fixed the generator netsnmp library issue. If you could please rebase this against current HEAD, it should fix the CI issue.

PierreF:
The rebase is done and the snmp.yaml difference is gone. The error on mixin seems unrelated (and also happen on master).

SuperQ:
Yes, the mixin issue is unrelated.

PR #782, opened 2022-08-02
generator/tree.go: Add support for subsequent address family in combi…

isodude:
…ned type

Since some MIB carry information about Safi (subsequent address family),
it will cause the generator to bail out, even though there is a valid Subtype before.

E.g.
Afi
Safi
Address

This is especially valid for Juniper BGP MIB.

I tried this out and it fixed all errors, and the output was the same as before.

Link: https://learningnetwork.cisco.com/s/question/0D53i00000KssMOCAZ/bgp-afi-and-safi

Signed-off-by: Josef Johansson <josef86@gmail.com>


generator/tree.go
```diff
@@ -363,11 +366,14 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 				if subtype, ok := combinedTypes[index.Type]; ok {
 					if prevType == subtype {
 						metric.Indexes = metric.Indexes[:len(metric.Indexes)-1]
+					if prev2Type == subtype {
```
SuperQ, line 369: I think you wanted an else if here?

```suggestion
					} else if prev2Type == subtype {
```

SuperQ (commented):
I wonder if we should have a general recursive stack here.

SuperQ (changes requested):
Needs to compile.

SuperQ (approved):

Issue #785, opened 2022-08-03
retries 0 is not respected


xkilian:
### Host operating system: output of `uname -a`

Ubuntu 18.04LTS

### snmp_exporter version: output of `snmp_exporter -version`

snmp_exporter 0.19

### What device/snmpwalk OID are you using?

N/A

### If this is a new device, please link to the MIB(s).

N/A

### What did you do that produced an error?
configure a device collection using the option:

retries 0

generate the configuration

### What did you expect to see?

retries 0 appear in the generated configuration

### What did you see instead?

The line is not placed in the snmp.yml output. Causing it to default to retries: 3.

Clarify the documentation to indicate if retries 0 (no retries) is a supported configuration or not. If it is not, is there a reason for this. Some devices cannot be polled twice in a row without a delay, otherwise the snmp stacks will crash.

SuperQ:
That is very strange. I did some tests and found that the retries is correctly parsed from the generator.yml when you run `generator generate`. But the Marshal yaml output skips this value.

It seems to do the same thing for other values like `max_repetitions: 0`.

This seems to be a bug in the yaml library output. Or possibly the way the walk parameters are inlined in the yaml struct.

PR #786, opened 2022-08-03
Fix retries in generator

SuperQ:
Fix parsing of `0` in retries in the generator config by making the
struct value a pointer[0].

Fixes: https://github.com/prometheus/snmp_exporter/issues/785

[0]: https://www.sohamkamani.com/golang/omitempty/#the-difference-between-0--and-nil


roidelapluie (approved):

roidelapluie:
this actually seems like a bug in the yaml lib

SuperQ:
I agree, it's a very strange behavior of the library.

Sadly, we're stuck with yaml.v2 due the new library not having a strict unmarshal.

We might consider moving to a different yaml library.

roidelapluie:
The new library has strict unmarshal.

PR #791, opened 2022-08-21
Add Eaton EPDU G3 SNMP support to snmp_exporter

scottlaird:
This adds support for Eaton EPDU G3 PDUs, including per-outlet power monitoring.

@SuperQ


RichiH:
Thanks @scottlaird !

PR #797, opened 2022-09-05
Import CyberPower MIBs, add cyberpower config.

jlrgraham:
Adds in the CyberPower MIBs (https://www.cyberpowersystems.com/product/software/mib-files/mib-v210/) and sets up a sample `cyberpower` profile for monitoring CyberPower UPS and PDU devices.


generator/generator.yml
```diff
@@ -710,3 +710,24 @@ modules:
...
+
+# CyberPower
+#
+# https://www.cyberpowersystems.com/product/software/mib-files/mib-v29/
+  cps_ups:
+    version: 1
+    walk:
+      - sysUpTime
+      - interfaces
+      - ifXTable
```
SuperQ, line 722:
```suggestion
```

generator/generator.yml
```diff
@@ -710,3 +710,24 @@ modules:
...
+      - 1.3.6.1.4.1.3808.1.1.1 # ups
+      - 1.3.6.1.4.1.3808.1.1.4 # environmentSensor
+
+  cps_pdu:
+    version: 1
+    walk:
+      - sysUpTime
+      - interfaces
+      - ifXTable
```
SuperQ, line 731: Let's simplify this into one cyberpower module.
```suggestion
```

generator/Makefile
```diff
@@ -323,3 +325,9 @@ $(MIBDIR)/EATON-OIDS.txt:
 	@echo ">> Downloading EATON-OIDS.txt to $@"
 	@curl $(CURL_OPTS) -o $@ $(EATON_OIDS_URL)
 
+$(MIBDIR)/CyberPower_MIB_v2.9.MIB:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading CyberPower_MIB_v2.9.MIB to $(TMP)"
+	@curl $(CURL_OPTS) -o $(TMP) $(CYBERPOWER_URL)
+	@unzip -j -d $(MIBDIR) $(TMP) CyberPower_MIB_v2.9.MIB
```
SuperQ, line 332: Let's name this without the version so updating it doesn't leave old files behind.

```suggestion
$(MIBDIR)/CyberPower.MIB:
	$(eval TMP := $(shell mktemp))
	@echo ">> Downloading CyberPower_MIB_v2.9.MIB to $(TMP)"
	@curl $(CURL_OPTS) -o $(TMP) $(CYBERPOWER_URL)
	@unzip -j -d $(MIBDIR) $(TMP) CyberPower.MIB
```

generator/generator.yml
```diff
@@ -710,3 +710,24 @@ modules:
         type: EnumAsInfo
       "1.3.6.1.4.1.534.6.6.7.6.1.1.3":
         type: DisplayString
+
+# CyberPower
+#
+# https://www.cyberpowersystems.com/product/software/mib-files/mib-v29/
+  cps_ups:
```
SuperQ, line 717: This should probably be `cyberpower` to make it more clear.

```suggestion
  cyberpower:
```

SuperQ (changes requested):
Thanks, some changes are needed.

generator/generator.yml
```diff
@@ -710,3 +710,16 @@ modules:
...
+
+# CyberPower
+#
+# https://www.cyberpowersystems.com/product/software/mib-files/mib-v29/
+  cyberpower:
+    version: 1
+    walk:
+      - sysUpTime
+      - interfaces
+      - ifXTable
```
SuperQ, line 722: Like I said, this needs to be removed.

```suggestion
```

SuperQ (changes requested):

generator/generator.yml
jlrgraham: Could you elaborate on why?  These are things I want to receive in the stats data.  If these are now part of a default collection list, they're listed in almost every other device type in the file, should those be cleaned up?

I'd be nice to have context to understand requested changes.

jlrgraham: Hi @SuperQ, just wanted to follow up here and make sure I understand the request.

What is the intended way that a user could get both device MIB specific data (ie: PDU or UPS metrics) as well as network interface metrics on the same device?

I'm not a Go programmer, but it looks like the `module` parameter to the query can only take a single value as a string argument.  So would we need to create two scrape jobs in Prometheus?  One to get the UPS data and one to get the IF data?

SuperQ: We're trying to clean up the duplication in the example configuration. You can get this data using the `if_mib` module as the default.

But this is just an example config, it's not meant to be every possible combination of MIBs / walks.

jlrgraham: > We're trying to clean up the duplication in the example configuration. You can get this data using the `if_mib` module as the default.
> 
> But this is just an example config, it's not meant to be every possible combination of MIBs / walks.

Thank you for the explanation, the PR has been updated.

SuperQ (approved):
Thanks!

mdlayher:
Just wanted to chime in to say thanks for this! Works perfectly with my new CyberPower PDU!

PR #828, opened 2022-12-15
Fix generator lookups to match OIDs closer to the index OID.

stephan-windischmann-sky:
When running the configuration generator with lookups were multilpe different nodes have the same label the generator was always grabbing the OID of the last MIB that was parsed.

I modified the lookup mechanism to first attempt to match the lookup node so that the OID matches that of the metric in all but the last branch.
In case no match is found, it reverts to the old behavior.


SuperQ:
Interesting, can you point to any specific OIDs where this changes things?

stephan-windischmann-sky:
> Interesting, can you point to any specific OIDs where this changes things?

I have attached the MIB with which we noticed the issue.
[mibs-vss47.zip](https://github.com/prometheus/snmp_exporter/files/11934384/mibs-vss47.zip)

As an example, both SYMSOFT-SS7FILTER-0-MIB.txt and SYMSOFT-TRAFFICSIMULATOR-0-MIB.txt both have nodes with the label layerName. This is used to lookup a human readable name.
WIth the upstream code, all lookups went to the last layerName node that was parsed instead of to the layerName node belonging to the same level.

SuperQ:
Interesting, I was hoping this would fix the conflict in `ifDescr` lookups between IF-MIB and PaloAlto PAN-COMMON-MIB. But it doesn't seem to help

SuperQ (approved):

SuperQ:
It looks like this affects the raritan module. Does the output change look right to you? Would you mind updating the snmp.yml so the generator test passes?

SuperQ (changes requested):
This needs an update to the snmp.yml.

stephan-windischmann-sky:
The changes in snmp.yml look good to me. Thanks.

SuperQ:
Yikes, it looks like the git history got messed up.

stephan-windischmann-sky:
Fixed git messup. snmp.yml is updated.

SuperQ (approved):

SuperQ:
Nice, thanks.

PR #833, opened 2023-01-10
Import TP-Link EAP MIB

jeansebastienh:
Importing the MIB for TPLink EAP wifi access  point.

Adding some minimal snmp configuration.

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -599,3 +599,9 @@ modules:
       - 1.3.6.1.4.1.3808.1.1.1 # ups
       - 1.3.6.1.4.1.3808.1.1.4 # environmentSensor
       - 1.3.6.1.4.1.3808.1.1.6 # ePDU2
+#
+# TP LINK EAP
+# http://static.tp-link.com/EAP_Private_Mibs_1.0.zip
+  eap:
+    walk:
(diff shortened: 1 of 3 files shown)


generator/generator.yml
```diff
@@ -720,3 +721,11 @@ modules:
       - 1.3.6.1.4.1.3808.1.1.1 # ups
       - 1.3.6.1.4.1.3808.1.1.4 # environmentSensor
       - 1.3.6.1.4.1.3808.1.1.6 # ePDU2
+#
+# TP LINK EAP
+# http://static.tp-link.com/EAP_Private_Mibs_1.0.zip
+  eap:
+    walk:
+      - interfaces
+      - sysUpTime
```
SuperQ, line 730: Now that we support multiple modules per scrape, we're deduplicating these from the vendor-specific modules.

You also need to update the `snmp.yml` so the tests pass.

```suggestion
```

SuperQ (changes requested):
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

SuperQ:
Still needs an update to the `snmp.yml`.

SuperQ (approved):
Thanks!

PR #837, opened 2023-01-13
Fix EATON download

francois2metz:
fix #836 and #829

I had to use a custom user agent, as akamai seems to forbid curl.

Issue #829, opened 2022-12-22
make: *** [Makefile:322: mibs/EATON-EPDU-MIB.txt] Error 22


Ampsyy:
I have tried this on two separate VMs and get the same issue. 

### Host operating system: output of `uname -a`
Linux  5.4.0-135-generic #152-Ubuntu SMP Wed Nov 23 20:19:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
and 
Linux  5.15.0-1009-oracle #12-Ubuntu SMP Thu Jun 2 11:44:19 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### snmp_exporter version: output of `snmp_exporter -version`
Not running yet

### What device/snmpwalk OID are you using?
NA
### If this is a new device, please link to the MIB(s).
NA
### What did you do that produced an error?
git clone https://github.com/prometheus/snmp_exporter.git
cd snmp_exporter/generator/
make mibs

### What did you expect to see?
Complete without any errors

### What did you see instead?
>> Downloading apc-powernet-mib
>> Downloading Cisco AIRESPACE-REF-MIB
>> Downloading Cisco AIRESPACE-WIRELESS-MIB
>> Downloading ARISTA-ENTITY-SENSOR-MIB
>> Downloading ARISTA-SMI-MIB
>> Downloading ARISTA-SW-IP-FORWARDING-MIB
>> Downloading Cisco ENTITY-MIB
>> Downloading Cisco ENTITY-SENSOR-MIB
>> Downloading Cisco ENTITY-STATE-MIB
>> Downloading Cisco ENTITY-STATE-TC-MIB
>> Downloading IANA charset MIB
>> Downloading IANA ifType MIB
>> Downloading IANA printer MIB
>> Downloading Cisco ISDN-MIB
>> Downloading KEEPALIVED-MIB
>> Downloading VRRP-MIB
>> Downloading VRRPv3-MIB
>> Downloading MIKROTIK-MIB
>> Downloading PICO-IPSEC-FLOW-MONITOR-MIB.txt
>> Downloading PICO-SMI-ID-MIB.txt
>> Downloading PICO-SMI-MIB.txt
>> Downloading Printer MIB v2
>> Downloading servertech-sentry3-mib
>> Downloading servertech-sentry4-mib
>> Downloading UBNT-UniFi-MIB
>> Downloading UBNT-AirFiber-MIB
[...]

Issue #831, opened 2023-01-02
cannot find oid '1.3.6.1.4.1.14179.2.1.1.1.38' to walk"


eburghar:
Apparently some recent changes from `$(CISCO_URL)/AIRESPACE-WIRELESS-MIB.my` triggers the bug.

Replacing with  `http://www.circitor.fr/Mibs/Mib/A/AIRESPACE-WIRELESS-MIB.mib` fixes the error.

### Host operating system: output of `uname -a`
alpine 3.17

### snmp_exporter version: output of `snmp_exporter -version`
0.21.0

### What did you do that produced an error?
MIBDIRS=mib generator generate

### What did you expect to see?
```
ts=2023-01-02T19:48:31.659Z caller=main.go:91 level=info msg="Config written" file=/home/packager/snmp_exporter/main/snmp_exporter/src/snmp_exporter-0.21.0/generator/snmp.yml
```

### What did you see instead?
```
level=error msg="Error generating config netsnmp" err="cannot find oid '1.3.6.1.4.1.14179.2.1.1.1.38' to walk"
```

SuperQ:
Yes, it looks like Cisco broke the MIB with a bunch of syntax errors. I filed an upstream PR to fix them.

For now, I've rolled back to a commit before it was broken. https://github.com/prometheus/snmp_exporter/pull/843

PR #841, opened 2023-01-30
eaton-fix

### LINKED ISSUE #836: EATON mib location is no longer valid
In `generator.yml` it references the EATON MIB as: https://powerquality.eaton.com/Support/Software-Drivers/Downloads/ePDU/EATON-EPDU-MIB.zip

But that now returns a 404 page.

h-leth:
This is my fix for the Eaton issues

Fixes #836 and #829
And are the code that is posted in PR #837

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -14,7 +14,7 @@
 MIBDIR   := mibs
 MIB_PATH := 'mibs'
 
-CURL_OPTS ?= -L -s --retry 3 --retry-delay 3 --fail
+CURL_OPTS ?= -L -s --retry 3 --retry-delay 3 --retry-max-time 60 --fail
 
 DOCKER_IMAGE_NAME ?= snmp-generator
 DOCKER_IMAGE_TAG  ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -14,7 +14,7 @@
 MIBDIR   := mibs
 MIB_PATH := 'mibs'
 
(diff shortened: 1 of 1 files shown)


h-leth:
Solves  #831

Let me know if anything need to be changed

scottlaird:
I added the original Eaton support; this change looks good to me.  I was about to do something similar to fix the build problem.

scottlaird:
As mentioned in #837, Eaton (really Akamai) seems to block Curl's User-Agent.  Also, it looks like Eaton's upstream source for this moved to https://www.eaton.com/content/dam/eaton/products/backup-power-ups-surge-it-power-distribution/Firmware/eaton-pdu-g3-firmware-mib.zip, which would need an intermediate unzip step to get EATON-EPDU-MIB.zip.  It's almost certainly easier to just fetch this from a different source, which is what this PR already does.

SuperQ:
Since Eaton can't make their MIBs accessible, I think it's more appropriate that we simply remove them from the examples.

h-leth:
>Since Eaton can't make their MIBs accessible, I think it's more appropriate that we simply remove them from the examples.

Sure I can remove them.

But what to do about issue #831, want to fix it in this PR or should I open a new one?
Or should that one be removed too? It's something with the new version of [AIRESPACE-WIRELESS-MIB.my](https://github.com/cisco/cisco-mibs/blob/main/v2/AIRESPACE-WIRELESS-MIB.my) on https://github.com/cisco/cisco-mibs

SuperQ:
Fixed with https://github.com/prometheus/snmp_exporter/pull/843

Issue #832, opened 2023-01-10
snmp_exporter.service sample file


fred-lix:
Bonjour,

In the ExecStart line, the configuration file uses quotes ( --config.file='/home/prometheus/snmp_exporter/snmp.yml' )

In Centos7, the service does not start with the following error :
level=error msg="Error parsing config file" err="open '/etc/prometheus/snmp.yml': no such file or directory"

removing the quotes solves the problem

Thank you

sheraazb:
Faced the same issue. Looks like systemd does not like the quotes. Thanks for reporting @fred-lix .

PR #842, opened 2023-02-13
Remove unmaintained example

SuperQ:
We don't generally maintain example service configs in our repos. This is more appropriately handled by prometheus-community/ansible.

Fixes: https://github.com/prometheus/snmp_exporter/issues/832


SuperQ:
I really don't want to maintain these examples, we don't have any consistency for this stuff on any of our repos.

RichiH:
Semi-random idea: What if we put it into exporter-toolkit?

dswarbrick:
I agree that systemd service files, init scripts etc should really be provided by the distros which package the exporters, since they usually involve distro-specific paths, users / groups, etc.

SuperQ:
Closing this for now.

Issue #619, opened 2021-02-25
Support `auth` alongside target & module as distinct property


RichiH:
auth is orthogonal from the target, the data model and from what data to receive. As such, it will be split out into its own section.

For backwards compatibility, inline auth should be supported, but the new pattern encouraged.

Open questions:
* Should we allow auth to live in a secondary file to completely detach global OID etc configuration from user-specific config?
* Should we offer inline-line loading of files or explicitly have a few distinct files? Inline seems better.
* Should we offer a baseline UI to show which files are loaded?
* Should the exporter emit information like files loaded in `_info` or through a gauge?

As this touches a lot of questions in https://github.com/prometheus/snmp_exporter/issues/85 CC @glensc

RichiH:
https://docs.google.com/document/d/1McJJIiJfHgoecVrVNXx4ABJmI5M21e-6O9IgMNbVnvw

RichiH:
After more thought:
> Should we allow auth to live in a secondary file to completely detach global OID etc configuration from user-specific config?
https://github.com/prometheus/snmp_exporter/issues/85 will take care of this and people can structure it as they wish.

The new default config we provide should most likely break it out.

> Should we offer inline-line loading of files or explicitly have a few distinct files? Inline seems better.

I flipped. Let's make it explicit on CLI. Directory support takes away the need for inlinig, IMO. As such, we should offer fewer ways to do the same thing to avoid long-term confusion.

> Should we offer a baseline UI to show which files are loaded?

I think we must.

> Should the exporter emit information like files loaded in _info or through a gauge?

I think we should. Unclear if we should do
* one `_info`, lots of labels
* one `_info`, overload a single label
* one metric per file

New open question: Does that mean we're listing the directory name, what is found and activated inside, or both of those?

xkilian:
What you are describing should be handled in logging at exporter startup, and in the web UI. It is not meant for info metrics IMO.

pobk:
A secondary authentication file that can be linked to targets using the similar mechanism to module would be appreciated here. That way, we're not passing credentials in the clear, and admins have the choice of implementing common credentials, or device specific credentials.

The way I see this working is where we would normally use `?module=<mib>` we would add `credentials=<creds>`

pobk:
My 2 pence:

1. Credentials should be stored separately from the MIB definitions. Having a file which contains credentials separate from the snmp config would allow a greater flexibility around securing that file.
[...]

PR #859, opened 2023-03-26
Split config of auth and modules

SuperQ:
Allow configuration of auth/version parameters separately from the walk and metrics in the generator and exporter configuration.
* Simplify startup with `ReloadConfig()`
* Make sure to init metrics on config reload.

Fixes: https://github.com/prometheus/snmp_exporter/issues/619


SuperQ:
CC @sebastien-coavoux

sebastien-coavoux:
Hey thanks for the tag, the only thing that might not be straighforward would be what to do with the auth.
This is supposed to be a get parameter received by l'exporter from prometheus from the code 
```
            <label>Target:</label> <input type="text" name="target" placeholder="X.X.X.X" value="1.2.3.4"><br>
            <label>Auth:</label> <input type="text" name="auth" placeholder="auth" value="public"><br>
            <label>Module:</label> <input type="text" name="module" placeholder="module" value="if_mib"><br>
```

But that might not be clear when reading the parameter in the Readme

SuperQ:
Also CC @dswarbrick for review.

SuperQ:
@sebastien-coavoux Thanks, yea, I forgot to update the top-level README. I've updated that with some more information on how to use the the split auth/module.

I've also renamed the default auth to `public_v2` to make it more clear that it's not directly a community string, but rather a configuration identifier. Kinda like how we changed default walk module to `if_mib`.

README.md
```diff
@@ -4,7 +4,11 @@ This exporter is the recommended way to expose SNMP data in a format which
 Prometheus can ingest.
 
 To simply get started, it's recommended to use the `if_mib` module with
-switches, access points, or routers.
+switches, access points, or routers using the `public_v2` auth module,
+which should be a read-only access community on the target device.
+
+Note, community strings in SNMP are not considered secrets, as they are sent
+unencrypted in SNMP v1 and v2c. For secure acces, SNMP v3 is required.
```
dswarbrick, line 11: Typo
```suggestion
unencrypted in SNMP v1 and v2c. For secure access, SNMP v3 is required.
```

main.go
```diff
@@ -75,6 +75,16 @@ func handler(w http.ResponseWriter, r *http.Request, logger log.Logger) {
 		return
 	}
 
+	authName := query.Get("auth")
+	if len(query["auth"]) > 1 {
+		http.Error(w, "'auth' parameter must only be specified once", 400)
```
dswarbrick, line 80: Whilst we're touching this code, it might be a nice opportunity to replace the numeric HTTP status codes with package constants from net/http, e.g. `http.StatusBadRequest`.

README.md
```diff
@@ -68,9 +72,9 @@ Start `snmp_exporter` as a daemon or from CLI:
 ./snmp_exporter
 ```
 
-Visit http://localhost:9116/snmp?module=if_mib&target=1.2.3.4 where `1.2.3.4` is the IP or
-FQDN of the SNMP device to get metrics from and `if_mib` is the default module, defined
-in `snmp.yml`.
+Visit [http://localhost:9116/snmp?auth=public_v2&module=if_mib&target=1.2.3.4] where `1.2.3.4` is the IP or
```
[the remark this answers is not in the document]
SuperQ: Sure, I guess `192.0.0.8` is the right one for this.

dswarbrick (approved):
LGTM

dswarbrick:
It seems that this would also address what #601 attempted to solve.

marctc (approved):
LGTM! I've been working on adapting the changes to Grafana Agent's SNMP integration and overall I didn't find any problems. It will introduce breaking changes but for sake of simplicity. If you curious how is looking, here's [WIP branch](https://github.com/grafana/agent/compare/snmp_exporter_cfg?expand=1) (cc @RichiH).

README.md
```diff
@@ -68,9 +72,9 @@ Start `snmp_exporter` as a daemon or from CLI:
 ./snmp_exporter
 ```
 
-Visit http://localhost:9116/snmp?module=if_mib&target=1.2.3.4 where `1.2.3.4` is the IP or
-FQDN of the SNMP device to get metrics from and `if_mib` is the default module, defined
-in `snmp.yml`.
+Visit [http://localhost:9116/snmp?auth=public_v2&module=if_mib&target=192.0.0.8] where `192.0.0.8` is the IP or
+FQDN of the SNMP device to get metrics from. Also note the use of default auth (`public_v2`) and
+default module (`if_mib`). The auth and module must be defined in the `snmp.yml`.
```
matthiasr, line 77: This is not clear to me. Are `public_v2` and `if_mib` default, or must they be defined? Are they optional in the URL?

matthiasr (commented):
What would a typical rollout of this version look like? It seems like it requires both changes to the configuration, _and_ to the scrape parameters, at the same time?

How could we ease this transition? For example, could the exporter understand the legacy configuration for a while? Or could we still support modules referencing a certain auth, so that the scrape URL doesn't have to change atomically?

matthiasr (commented):
What would a typical rollout of this version look like? It seems like it requires both changes to the configuration, _and_ to the scrape parameters, at the same time?

How could we ease this transition? For example, could the exporter understand the legacy configuration for a while? Or could we still support modules referencing a certain auth, so that the scrape URL doesn't have to change atomically?

SuperQ:
@matthiasr Because we're adding top-level keys to the config file there's no easy transition. IMO it's just not something we can / should support. It's a breaking change, but for good reason.

RichiH:
Is there anything preventing us from parsing **either** the old format **or** the new?

I would strongly prefer to avoid a breaking change, but I would even more strongly prefer for this to move and the functionality to be merged in the near future.

SuperQ:
Of course it's technically possible to support both file formats. But it would require a lot of more complicated code and I just don't think anyone has the time to write and maintain the spaghetti necessary to do it.

I would rather cut v0.22.0, merge this, and soon after cut v0.23.0 with the breaking changes only.

roidelapluie:
I am on the side of breaking changes on this one.

roidelapluie:
I think we need an explicit "Migration Guide" as a blog post and in the readme. Maybe with a tool to do the migration.

roidelapluie:
What if we write a tool to do the migration, and we use it transparently for the users?

We would parse only the new format but would be able to "in-memory" change from the old to the new.

SuperQ:
Hmm, a `generator migrate` subcommand could be useful. But I don't have time to write such a thing.

RichiH:
How would it migrate the scrape config, though?

Or would we allow people to refer to a default auth per module and use that?

SuperQ:
Users will have to update. There's no way around this, it's a breaking change.

roidelapluie:
Ok let's go for the breaking change with good docs

SuperQ:
@roidelapluie I've added a changelog entry and migration guide.

auth-split-migration.md
```diff
@@ -0,0 +1,59 @@
...
+    community: public
+```
+
+Would now become:
+
+```yaml
+auths:
+  public_v2:
+    community: public
+    versoin: 2
```
sebastien-coavoux, line 31: `version`

auth-split-migration.md
```diff
@@ -0,0 +1,59 @@
+# Module and Auth Split Migration
```
matthiasr, line 1: One question I would like this to answer is how to sequence the change if you need a different auth than `public_v2`. Is it safe to specify the `auth` parameter with 0.22, then upgrade the exporter? Maybe this is obvious to experienced users of the exporter, but I'd rather make it explicit for the "someone else installed this 4 years ago and I need to keep it alive" target audience.

matthiasr (approved):

auth-split-migration.md
SuperQ: No, this is already explicitly stated on line 3 that the config format in 0.23.0 has changed.

In <= 0.22.0 there was no top level `modules:` key, just a bare list of modules.

SuperQ: I reworded line 3 slightly.

RichiH:
I can't find the reference off-hand, but I believe that we discussed having a top-level key for the config version. Even if no one is motivated to create config handling today, or ever, I think it would be good to have an instant and unambiguous way to discern config version.

SuperQ:
@RichiH A `version:` key would be something we could add, but it's not going to help much here since we're not going to support parsing the original version and the original config had no version key so it's not going to help the transition at all.

But it should be easy enough to add it.

SuperQ:
@RichiH Added a `version` key to the generator and exporter configs.

main.go
```diff
@@ -160,10 +183,10 @@ func main() {
 	prometheus.MustRegister(version.NewCollector("snmp_exporter"))
 
 	// Bail early if the config is bad.
-	var err error
-	sc.C, err = config.LoadFile(*configFile)
+	err := sc.ReloadConfig(*configFile)
 	if err != nil {
 		level.Error(logger).Log("msg", "Error parsing config file", "err", err)
+		level.Error(logger).Log("msg", "Possible old config file, see https://github.com/prometheus/snmp_exporter/blob/main/auth-split-migration.md")
```
RichiH, line 189: This should be more explicit and a clear yes/no?

RichiH:
Thanks! I know it doesn't help as of today, but it's still more resilient for the future. Nit: Maybe `config_version` instead of `version` as we have a `version` in `auth` as well?

I also pushed a minor commit to make errors more explicit into this branch.

RichiH:
NB: If we simply error out, we will make rolling upgrades impossible. Users will need to upgrade all their Prometheus configs at once.

SuperQ:
@RichiH Please use the suggestion feature of GitHub instead of pushing commits without asking. That error message change does not fit Go style.

ZoldanYao:
This function is very very useful. I hope I can use this function  asap.

SuperQ:
Any last thoughts? I plan to merge this soon.

main.go
SuperQ: Agreed, it's not an easy upgrade. But the end result is worth it and more future-proof.

config/config.go
```diff
@@ -60,17 +63,21 @@ var (
 	DefaultRegexpExtract = RegexpExtract{
 		Value: "$1",
 	}
+
+	ErrUnsupportedVersion = errors.New("unsupported config version")
```
RichiH, line 67:
```suggestion
	ErrUnsupportedVersion = errors.New("Unsupported config version, we only support config version 2. Please see https://github.com/prometheus/snmp_exporter/blob/main/README.md#debugging")
```

RichiH:
There's no in-review way to add the

````
#Debugging

## Unsupported config version

There was a change from SNMP exporter 0.22 to 0.23. The `auth` block is now split
out from the scrape config. You will need to change both the configuration of the
SNMP exporter and Prometheus' scrape config. For details, please see our
[config upgrading doc](https://github.com/prometheus/snmp_exporter/blob/main/auth-split-migration.md).
````

to the end of the README.

File: config/config.go (hunk 8 above)
Comment (SuperQ) [reply config/config.go:67]: This is a Go error, it shouldn't be this verbose.

**Comment by SuperQ** [comment]
@RichiH That change is not needed as we already give a URL on error.

File: config/config.go (hunk 8 above)
Comment (RichiH) [reply config/config.go:67]: Will the user see it on CLI?

Comment (SuperQ) [reply config/config.go:67]: Yes, but basically no. That's the whole reason for the "Possible old config file" error. The user won't directly see this error. It's for internal `errors.Is()` use.

**Comment by SuperQ** [comment]
There already is `"Possible old config file"`. As I explained previously, due to the current strict yaml parsing, we can't do much about this. We would have to rewrite the yaml strict parser with our own validator. Frankly, I don't have the time or energy to do this.

File: auth-split-migration.md [hunk 9]
```diff
@@ -0,0 +1,59 @@
...
+
+In [version 0.23.0](https://github.com/prometheus/snmp_exporter/releases/tag/v0.23.0) the configuration for the `snmp_epxorter` the configuration file format has changed. Configuration files for versions v0.22.0 and before will not work. The configuration was split from a flat list of modules to separate metric walking/mapping modules and authentication configuratoins.
```
candlerb, line 3: s/configuratoins/configurations/

PR #863, opened 2023-04-06
Remove interface specific OIDs from more specific configs in sample config

bastischubert:
As discussed with @SuperQ this PR removes the if_mib OIDs from the generated snmp.yml for all **more** specific modules **BUT** the mikrotik module (which still uses a slighly different interface walk - tested with routeros 6).

* overrides where inconsistent between different modules
* lookups where inconsistent between different modules

### DIFF
--- a/snmp.yml
+++ b/snmp.yml
@@ -2956,6 +2956,7 @@ cisco_wlc:
     indexes:
     - labelname: bsnAPDot3MacAddress
       type: PhysAddress48
+      fixed_size: 6
     - labelname: bsnAPIfSlotId
       type: gauge
     lookups:
@@ -2974,6 +2975,7 @@ cisco_wlc:
     indexes:
     - labelname: bsnAPDot3MacAddress
       type: PhysAddress48
+      fixed_size: 6
     - labelname: bsnAPIfSlotId
       type: gauge
     - labelname: bsnAPIfNoiseChannelNo
@@ -2993,6 +2995,7 @@ cisco_wlc:
     indexes:
     - labelname: bsnAPDot3MacAddress
       type: PhysAddress48
+      fixed_size: 6
     - labelname: bsnAPIfSlotId
       type: gauge
     lookups:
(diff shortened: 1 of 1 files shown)


generator/generator.yml
```diff
@@ -63,7 +58,6 @@ modules:
     version: 1
     walk:
       - sysUpTime
```
SuperQ, line 60: Let's also drop the `sysUpTime` pasta.

generator/generator.yml
```diff
@@ -635,7 +586,6 @@ modules:
 # http://download2.mikrotik.com/Mikrotik.mib
   mikrotik:
     walk:
-      - interfaces
       - ifMIB
```
SuperQ, line 589: I wonder if we should remove this as well, as it's covering both `interfaces` and `ifXTable`.

generator/generator.yml
bastischubert: ✅ removed

generator/generator.yml
bastischubert: `ifMIB` translates to `1.3.6.1.2.1.31` whereas `interfaces` is `1.3.6.1.2.1.2` and `ifXTable` `1.3.6.1.2.1.31.1.1` is a subset of `ifMIB` 
We'd be missing `ifStackTable`, `ifTestTable`, `ifRcvAddressTable`, `ifTableLastChange` and `ifStackLastChange` but i doubt that these are of any value

bastischubert: it took me some time to get hands on a mikrotik device to test the changes.
removed also the ifMIB as the few other OIDs are not really useful

SuperQ (changes requested):
Ping, the snmp.yml needs an update to match these changes.

SuperQ:
I think your generator binary is out of date, the snmp.yaml changes still don't match in CI.

SuperQ (approved):

PR #864, opened 2023-04-06
fix mib parse error

bastischubert:
Update Cyberpower MIB Version to 2.11 as there is an error in 2.10 which leads to parse error by net-snmp

```
generator git:(bastischubert/add-geist-mib) ✗ ./generator parse_errors
ts=2023-04-06T14:31:42.610Z caller=net_snmp.go:161 level=info msg="Loading MIBs" from=mibs
ts=2023-04-06T14:31:42.911Z caller=main.go:119 level=warn msg="NetSNMP reported parse error(s)" errors=2
No log handling enabled - using stderr logging
Did not find 'DisplayString' in module #-1 (mibs/CyberPower.MIB)
```


SuperQ:
Looks good, but you need to generate and update the snmp.yml file.

bastischubert:
unfortunately the error still exists even after the version bump. need to figure out why it worked once and now errors again 😕

SuperQ:
We actually worked around the MIB error in https://github.com/prometheus/snmp_exporter/pull/868

bastischubert:
fixed with #868

Issue #855, opened 2023-03-20
exporter crash upon using module


PrestonTaylor:
### Host operating system: output of `uname -a`
Linux someserver 3.10.0-1062.el7.x86_64 #1 SMP Wed Aug 7 18:08:02 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

### snmp_exporter version: output of `snmp_exporter -version`
I'm using #624 

### What device/snmpwalk OID are you using?
1.3.6.1.4.1.6527.3.1.2.81.1.11.3.1.200
 with 
      - source_indexes: [tmnxWlanGwUeQryId] (1.3.6.1.4.1.6527.3.1.2.81.1.11.3.1.1)
        lookup: tmnxWlanGwUeQryWhereState (1.3.6.1.4.1.6527.3.1.2.81.1.11.3.1.3)

### If this is a new device, please link to the MIB(s).
Not sure where to find publicly I can share via this https://drive.google.com/file/d/1MkizXzWaEokQTLpQSGFKQsBJMv-6UbY_/view?usp=share_link

### What did you do that produced an error?
error ocurrs only when I include the lookup

### What did you expect to see?
minimum: no crash, ideally: no error and valid data returned. using snmpwalk and manageengine MIBBrowser I can walk this table without issue

### What did you see instead?

Mar 20 18:56:04 myexporterserver snmp_exporter: ts=2023-03-20T22:56:04.339Z caller=main.go:98 level=debug module=mymodule target=someserver msg="Starting scrape"
Mar 20 18:56:04 myexporterserver snmp_exporter: ts=2023-03-20T22:56:04.358Z caller=collector.go:243 level=debug module=mymodule target=someserver msg="Walking subtree" oid=1.3.6.1.4.1.6527.3.1.2.81.1.11.3.1.200
Mar 20 18:56:05 myexporterserver snmp_exporter: ts=2023-03-20T22:56:05.413Z caller=collector.go:256 level=debug module=mymodule target=someserver msg="Walk of subtree completed" oid=1.3.6.1.4.1.6527.3.1.2.81.1.11.3.1.200 duration_seconds=1.055329302s
Mar 20 18:56:05 myexporterserver snmp_exporter: ts=2023-03-20T22:56:05.413Z caller=collector.go:243 level=debug module=mymodule target=someserver msg="Walking subtree" oid=1.3.6.1.4.1.6527.3.1.2.81.1.11.3.1.27
[...]

PR #870, opened 2023-04-18
Treat Bits as OctetString

sebastien-coavoux:
Related to #855 , looks like this quick fix is the best option until we have a use case to elaborate more


sebastien-coavoux:
poke @SuperQ in case you missed it

SuperQ (approved):
Thanks!

Issue #715, opened 2021-12-22
How can we allow MIBOPTS in generator


imranrazakhan:
### Host operating system: output of `uname -a`
`Linux  3.10.0-1127.10.1.el7.x86_64 #1 SMP Wed Jun 3 14:28:03 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

### snmp_exporter version: output of `snmp_exporter -version`
`v0.20.0`

### What device/snmpwalk OID are you using?
`MOXA EDR810`

### If this is a new device, please link to the MIB(s).
`https://cdn-cms.azureedge.net/getmedia/a8c78c0e-8065-402e-b492-89842ca1905b/moxa-edr-810-series-software-package-v5.10.mib`

### What did you do that produced an error?
I was not able to generate snmp.yml so for troubleshooting i ran following and i found that it was not able to parse my mib
```
./generator parse_errors
Bad operator (_): At line 91 in mibs/moxa-edr-810-series-software-package-v5.10.mib
```

I tried snmpwalk to verify my mib and found that if i set `MIBOPTS `with `u `option then it successfully parse my mib
```
snmpwalk --help
-P MIBOPTS            Toggle various defaults controlling MIB parsing:
                          u:  allow the use of underlines in MIB symbols
```

How can we pass this option to generator?

LupoOnGitHub:
Unfortunately, some/many vendor-specific MIB files use the underline in MIB symbols, so this would be a very helpful feature.

PR #873, opened 2023-05-02
Add MIBOPTS flag to the generator

SuperQ:
Allow passing net-snmp MIBOPTS for controlling MIB parsing.

Fixes: https://github.com/prometheus/snmp_exporter/issues/715

### DIFF
--- a/generator/main.go
+++ b/generator/main.go
@@ -94,6 +94,7 @@ func generateConfig(nodes *Node, nameToNode map[string]*Node, logger log.Logger)
 
 var (
 	failOnParseErrors  = kingpin.Flag("fail-on-parse-errors", "Exit with a non-zero status if there are MIB parsing errors").Default("false").Bool()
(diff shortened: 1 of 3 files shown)


generator/main.go
```diff
@@ -94,6 +94,7 @@ func generateConfig(nodes *Node, nameToNode map[string]*Node, logger log.Logger)
 
 var (
 	failOnParseErrors  = kingpin.Flag("fail-on-parse-errors", "Exit with a non-zero status if there are MIB parsing errors").Default("false").Bool()
+	snmpMIBOpts        = kingpin.Flag("snmp.mibopts", "Toggle various defaults controlling MIB parsing, see snmpwalk --help").String()
```
Nexucis, line 97: could be cool to check if the value is matching what is expected ? Or is it done in a later stage ?

SuperQ: The values are dependent on what version of the C library we compile against. It's basically up to the user to figure this out.

Nexucis (approved):

PR #883, opened 2023-05-29
Enable environment variable expansion during config file load

servak:
Hi. I wanted to hide community information in managing snmp.yml, so I made it available from environment variables.

Please let me know if there is anything I am missing.


SuperQ:
Environment variables are less safe than config files. We prefer to keep this configuration as-is.

PR #891, opened 2023-06-03
Update Prometheus libraries

SuperQ:
* Update Prometheus libraries.
* Update fixtures for new proto rendering.


SuperQ:
For the metric comparison, we should probably switch this over to the client_golang testutil helper we're working on. This should eliminate the need to do string comparison hacks.

matthiasr (approved):

PR #893, opened 2023-06-08
Add Airfiber LTU MIB support (MIBs and OIDs)

alejandroscf:
I've added support Ubiquiti AirFiber devices based in the LTU technology they use another MIB file. I've tested it in production in our network with more than twenty AirFiber 5XHD devices.
cc @SuperQ


SuperQ (changes requested):
You also need to update the snmp.yml

SuperQ:
Closing as stale.

PR #895, opened 2023-06-13
Add Netgear ReadyNAS and ReadyData appliances

reynico:


SuperQ:
This needs a rebase.

SuperQ (changes requested):
Rebase required.

generator/generator.yml
```diff
@@ -28,6 +28,13 @@ modules:
       ifType:
         type: EnumAsInfo
 
+  readynas:
+    walk:
+      - 1.3.6.1.2.1.1.5               # Name
+      - 1.3.6.1.2.1.25.1.1            # Uptime
+      - 1.3.6.1.4.1.4526              # Raid/Disks status
+      - 1.3.6.1.2.1.25.3.3.1.2        # CPU status
```
SuperQ, line 36: With the refactoring of the generator, we only put device-specific walks in modules. The other walks are accounted for in other modules. For example the `hrDevice` module.


Also, let's be a little more specific in our walk specifications. Not the whole enterprise.

```suggestion
      - 1.3.6.1.4.1.4526              # Raid/Disks status
```

reynico:
up to this point, and given the divergence between branches, I think it'd better start from scratch (-: @SuperQ what do you think?

PR #897, opened 2023-06-17
Improve linting

SuperQ:
* Enable more golangci-lint checks.
* Cleanup various linter issues.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,28 @@
-# Run only staticcheck for now. Additional linters will be enabled one-by-one.
 linters:
   enable:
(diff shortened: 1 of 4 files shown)


SuperQ:
The `govet` issue I'm not sure how to fix. It warns about an unkeyed struct literal field, but it seems like this is a common pattern for yaml unmarshal.

```
collector/collector_test.go:54:15: composites: github.com/prometheus/snmp_exporter/config.Regexp struct literal uses unkeyed fields (govet)
							Regex: config.Regexp{
							       ^
```

marctc (approved):

PR #904, opened 2023-07-01
Fix metrics path not using command-line argument value

hhromic:
Hello,
I casually discovered this bug while looking at the source code of the SNMP exporter.
The exporter currently hard codes `/metrics` instead of using the `--web.telemetry-path` command-line argument value.

The fix here aligns with the current approach in the Node Exporter:
https://github.com/prometheus/node_exporter/blob/c31ebb43590cb0055acd4bc00a432f6a2dafdd08/node_exporter.go#L188

### DIFF
--- a/main.go
+++ b/main.go
@@ -220,7 +220,7 @@ func main() {
 		}
 	}()
 
-	http.Handle("/metrics", promhttp.Handler()) // Normal metrics endpoint for SNMP exporter itself.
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Oh, good find. I must have missed that when I was adjusting the landing page code.

hhromic:
Happy to help! :)

PR #911, opened 2023-07-07
Added version to document

servak:
I was stuck with the following error, How about mentioning version in the documentation?
```
level=error msg="Error generating config netsnmp" err="unsupported config version"
```


SuperQ (approved):
Thanks, thought I had updated all the documentation locations. Missed a couple.

PR #918, opened 2023-07-17
Rename top level config version key

SuperQ:
Avoid confusion between SNMP versions and the config format version.


SuperQ:
Closing in favor of https://github.com/prometheus/snmp_exporter/pull/921

PR #923, opened 2023-07-19
Ignore version in config files

SuperQ:
Re-add the version key, but omitempty so it is simply ignored for now.


roidelapluie (commented):
LGTM

PR #908, opened 2023-07-05
Improved Lookup process for label information to utilize TypeMapping

servak:
We've made improvements to the Metric Lookup process. 
Previously, the byte representation of labels was output as-is under certain circumstances. With this update, it now references the processing of TypeMapping for output.

generator.yml
```
modules:
  default:
    walk:
    - lldpRemPortIdSubtype
    - lldpLocPortIdSubtype
    - lldpRemPortId
    - lldpLocPortId
    - lldpRemSysName
    lookups:
    - source_indexes:
      - lldpRemTimeMark
      - lldpRemLocalPortNum
      - lldpRemIndex
      lookup: "lldpRemPortId"
    - source_indexes: ['lldpRemLocalPortNum']
      lookup: "lldpLocPortId"
    overrides:
      lldpRemPortIdSubtype:
        ignore: true
      lldpLocPortIdSubtype:
        ignore: true
```

before
```
lldpLocPortId{lldpLocPortId="Management1",lldpLocPortNum="245"} 1
# HELP lldpRemPortId The string value used to identify the port component associated with the remote system. - 1.0.8802.1.1.2.1.4.1.1.7
# TYPE lldpRemPortId gauge
lldpRemPortId{lldpLocPortId="0x4D616E6167656D656E743100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",lldpRemIndex="2",lldpRemLocalPortNum="245",lldpRemPortId="0x4769312F31340000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",lldpRemTimeMark="0"} 1
# HELP lldpRemSysName The string value used to identify the system name of the remote system. - 1.0.8802.1.1.2.1.4.1.1.9
# TYPE lldpRemSysName gauge
lldpRemSysName{lldpLocPortId="0x4D616E6167656D656E743100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",lldpRemIndex="14",lldpRemLocalPortNum="245",lldpRemPortId="0x4769312F31340000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",lldpRemSysName="hostname",lldpRemTimeMark="0"} 1
```

[...]

PR #924, opened 2023-07-19
Release v0.23.0-rc.1

SuperQ:
BREAKING CHANGES:

This version of the exporter introduces a new configuration file format. This new format separates the walk and metric mappings from the connection and authentication settings. This allows for easier configuration of different auth params without having to duplicate the full walk and metric mapping.

See auth-split-migration.md for more details.

* [CHANGE] Split config of auth and modules #859
* [FEATURE] Add support for parsing SNMP transport from target #914
* [ENHANCEMENT] Improved Lookup process for label information #908
* [BUGFIX] Fix metrics path not using command-line argument value #904


roidelapluie (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.23.0-rc.0 / 2023-07-04
+## 0.23.0-rc.1 / 2023-07-04
```
roidelapluie, line 1: Jul 20 ?

SuperQ: Fixed

RichiH:
Would it make sense to generate the version key by default but only INFO or WARN when it's missing?

I still believe that it's a good idea long term, even if we can't rely on it existing as of today.

Issue #731, opened 2022-02-18
Can multiple modules be supported？


linqianye:
hello,
Can SNMP exporter support multiple modules to adapt to private MIBs of different vendors?
Now it's like this in prometheus.yml,

```yaml
  - job_name: 'test-switch'
    scrape_interval: 60s
    static_configs:
     - targets:
       - 172.22.1.57
       - 172.22.1.58
    metrics_path: /snmp
    params:
      module: [if_mib]
    relabel_configs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
        replacement: 127.0.0.1:9116
```
In the code above, can you realize the following functions?

```yaml
  - job_name: 'test-switch'
    scrape_interval: 60s
    static_configs:
     - targets:
       - 172.22.1.57
       - 172.22.1.58
    metrics_path: /snmp
    params:
      modules:
      - if_mib
      - cisco_bgp
      - juniper_ospf
```

mv-rodriguez:
Hi, the way I do it is using relabeling on prometheus.yml:
```
- job_name: snmp
  honor_labels: true
  metrics_path: /snmp
  params:
    module: [generic]
  file_sd_configs:
  - files:
    - 'snmp/snmp_targets.json'
  relabel_configs:
    - source_labels: [__module]
      target_label: __param_module
    - source_labels: [__address__]
      target_label: __param_target
    - source_labels: [__param_target]
      target_label: instance
    - target_label: __address__
      replacement: snmp_exporter:9116
```
In the _snmp_targets.json_ there is a disposable label "__module" that defines the specific snmp module the target should use.  I'm not sure if the "honor_labels" option is required though, but as the saying says _"if it is working..."_

aaronnad:
+1 for this, would make it much easier for being able to use a single job.
I personally would prefer the job to run in parallel, that way the interval is known for a user where they may have some modules that take longer than others to run, meaning the scrape_timeout would not need to be extended per job.

amuntean-godaddy:
+1

candlerb:
```
    params:
      modules:
      - if_mib
      - cisco_bgp
      - juniper_ospf
```

I would very much like this too. However, I want to be able to specify it per-target using rewriting rules, and currently a `__param_XXX` module cannot generate a list, i.e. repeated instances of the same query parameter.

Hence I would rather it were a comma-separated list:

```
    params:
      modules:
[...]

PR #945, opened 2023-08-14
Support fetching multiple modules in one scrape

servak:
I implemented the ability to scrape in multiple modules. 
Parallel processing depends on user requirements, so I made it possible to specify the number of parallels. #731 

```
> ./snmp_exporter -h
usage: snmp_exporter [<flags>]

Flags:
  -h, --[no-]help               Show context-sensitive help (also try --help-long and --help-man).
...
      --[no-]dry-run            Only verify configuration is valid and exit.
      --concurrency=1           Specify the number of modules to fetch concurrently <- Added
```

## test
generator.yml
```
modules:
  sysUpTime:
    walk:
    - sysUpTime
  sysDescr:
    walk:
    - sysDescr
```

output
```
curl 'http://localhost:9116/snmp?target=cumulus&module=sysUpTime&module=sysDescr'
# HELP snmp_packet_duration_seconds A histogram of latencies for SNMP packets.
# TYPE snmp_packet_duration_seconds histogram
snmp_packet_duration_seconds_bucket{le="0.0001"} 0
snmp_packet_duration_seconds_bucket{le="0.0002"} 0
snmp_packet_duration_seconds_bucket{le="0.0004"} 0
snmp_packet_duration_seconds_bucket{le="0.0008"} 0
snmp_packet_duration_seconds_bucket{le="0.0016"} 0
snmp_packet_duration_seconds_bucket{le="0.0032"} 0
snmp_packet_duration_seconds_bucket{le="0.0064"} 0
snmp_packet_duration_seconds_bucket{le="0.0128"} 0
snmp_packet_duration_seconds_bucket{le="0.0256"} 0
snmp_packet_duration_seconds_bucket{le="0.0512"} 0
snmp_packet_duration_seconds_bucket{le="0.1024"} 0
snmp_packet_duration_seconds_bucket{le="0.2048"} 0
snmp_packet_duration_seconds_bucket{le="0.4096"} 0
snmp_packet_duration_seconds_bucket{le="0.8192"} 0
snmp_packet_duration_seconds_bucket{le="1.6384"} 0
snmp_packet_duration_seconds_bucket{le="+Inf"} 0
snmp_packet_duration_seconds_sum 0
snmp_packet_duration_seconds_count 0
# HELP snmp_packet_retries_total Number of SNMP packet retries.
# TYPE snmp_packet_retries_total counter
snmp_packet_retries_total 0
# HELP snmp_packets_total Number of SNMP packet sent, including retries.
# TYPE snmp_packets_total counter
snmp_packets_total 0
# HELP snmp_scrape_duration_seconds Total SNMP time scrape took (walk and processing).
# TYPE snmp_scrape_duration_seconds gauge
snmp_scrape_duration_seconds{module="sysDescr"} 0.081744083
snmp_scrape_duration_seconds{module="sysUpTime"} 0.0794535
# HELP snmp_scrape_packets_retried Packets retried for get, bulkget, and walk.
# TYPE snmp_scrape_packets_retried gauge
snmp_scrape_packets_retried{module="sysDescr"} 0
snmp_scrape_packets_retried{module="sysUpTime"} 0
# HELP snmp_scrape_packets_sent Packets sent for get, bulkget, and walk; including retries.
# TYPE snmp_scrape_packets_sent gauge
snmp_scrape_packets_sent{module="sysDescr"} 1
snmp_scrape_packets_sent{module="sysUpTime"} 1
# HELP snmp_scrape_pdus_returned PDUs returned from get, bulkget, and walk.
# TYPE snmp_scrape_pdus_returned gauge
snmp_scrape_pdus_returned{module="sysDescr"} 1
snmp_scrape_pdus_returned{module="sysUpTime"} 1
# HELP snmp_scrape_walk_duration_seconds Time SNMP walk/bulkwalk took.
# TYPE snmp_scrape_walk_duration_seconds gauge
snmp_scrape_walk_duration_seconds{module="sysDescr"} 0.081711667
snmp_scrape_walk_duration_seconds{module="sysUpTime"} 0.079426875
# HELP snmp_unexpected_pdu_type_total Unexpected Go types in a PDU.
# TYPE snmp_unexpected_pdu_type_total counter
snmp_unexpected_pdu_type_total 0
# HELP sysDescr A textual description of the entity - 1.3.6.1.2.1.1.1
# TYPE sysDescr gauge
sysDescr{sysDescr="Cumulus-Linux **********"} 1
# HELP sysUpTime The time (in hundredths of a second) since the network management portion of the system was last re-initialized. - 1.3.6.1.2.1.1.3
# TYPE sysUpTime gauge
sysUpTime 1.387417375e+09
```

## Confirmation of Significant Changes
Since the change will add module information to the metrics output by snmp_exporter as standard, could you please confirm that this is the right direction for the change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,25 @@ Note that [URL encoding](https://en.wikipedia.org/wiki/URL_encoding) should be u
 to the `:` and `/` characters. Prometheus encodes query parameters automatically and manual encoding
 is not necessary within the Prometheus configuration file.
 
+## Multi-Module Handling
+The multi-module functionality allows you to specify multiple modules, enabling the retrieval of information from several modules in a single scrape.
+The concurrency can be specified using the snmp-exporter option --concurrency (the default is 1).
+
+Note: This implementation does not perform any de-duplication of walks between different modules.
+
+There are two ways to specify multiple modules. You can either separate them with a comma or define multiple params_module.
+The URLs would look like this:
+
+For comma separation:
+```
+http://localhost:9116/snmp?module=if_mib,arista_sw&target=192.0.0.8
+```
+
+For multiple params_module:
+```
+http://localhost:9116/snmp?module=if_mib&module=arista_sw&target=192.0.0.8
+```
+
 ## Configuration
 
 The default configuration file name is `snmp.yml` and should not be edited
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ is not necessary within the Prometheus configuration file.
 
 ## Multi-Module Handling
 The multi-module functionality allows you to specify multiple modules, enabling the retrieval of information from several modules in a single scrape.
-The concurrency can be specified using the snmp-exporter option --concurrency (the default is 1).
+The concurrency can be specified using the snmp-exporter option `--snmp.module-concurrency` (the default is 1).
 
 Note: This implementation does not perform any de-duplication of walks between different modules.
 
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -44,8 +44,8 @@ var (
 	wrapCounters           = kingpin.Flag("snmp.wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
 	srcAddress             = kingpin.Flag("snmp.source-address", "Source address to send snmp from in the format 'address:port' to use when connecting targets. If the port parameter is empty or '0', as in '127.0.0.1:' or '[::1]:0', a source port number is automatically (random) chosen.").Default("").String()
 
-	SnmpDuration = promauto.NewSummaryVec(
-		prometheus.SummaryOpts{
+	SnmpDuration = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
 			Name: "snmp_collection_duration_seconds",
 			Help: "Duration of collections by the SNMP exporter",
 		},
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -89,8 +89,12 @@ func listToOid(l []int) string {
 	return strings.Join(result, ".")
 }
 
-func InitModuleMetrics(auth, module string) {
-	snmpDuration.WithLabelValues(auth, module)
+func InitModuleMetrics(auths map[string]*config.Auth, modules map[string]*config.Module) {
+	for auth := range auths {
+		for module := range modules {
+			snmpDuration.WithLabelValues(auth, module)
+		}
+	}
 }
 
 type ScrapeResults struct {
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -527,12 +527,10 @@ func (c Collector) Collect(ch chan<- prometheus.Metric) {
 		}()
 	}
 
-	go func() {
-		for _, module := range c.modules {
-			workerChan <- module
-		}
-		close(workerChan)
-	}()
+	for _, module := range c.modules {
+		workerChan <- module
+	}
+	close(workerChan)
 	wg.Wait()
 }
 
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -44,7 +44,7 @@ var (
 	wrapCounters           = kingpin.Flag("snmp.wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
 	srcAddress             = kingpin.Flag("snmp.source-address", "Source address to send snmp from in the format 'address:port' to use when connecting targets. If the port parameter is empty or '0', as in '127.0.0.1:' or '[::1]:0', a source port number is automatically (random) chosen.").Default("").String()
 
-	SnmpDuration = promauto.NewHistogramVec(
+	snmpDuration = promauto.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name: "snmp_collection_duration_seconds",
 			Help: "Duration of collections by the SNMP exporter",
@@ -87,6 +87,10 @@ func listToOid(l []int) string {
 		result = append(result, strconv.Itoa(o))
 	}
 	return strings.Join(result, ".")
+}
+
+func InitModuleMetrics(auth, module string) {
+	snmpDuration.WithLabelValues(auth, module)
 }
 
 type ScrapeResults struct {
@@ -364,12 +368,24 @@ type internalMetrics struct {
 	snmpRetries           prometheus.Counter
 }
 
+type NamedModule struct {
+	*config.Module
+	name string
+}
+
+func NewNamedModule(name string, module *config.Module) *NamedModule {
+	return &NamedModule{
+		Module: module,
+		name:   name,
+	}
+}
+
 type Collector struct {
 	ctx         context.Context
 	target      string
 	auth        *config.Auth
 	authName    string
-	modules     map[string]*config.Module
+	modules     []*NamedModule
 	logger      log.Logger
 	metrics     internalMetrics
 	concurrency int
@@ -414,7 +430,7 @@ func newInternalMetrics(reg prometheus.Registerer) internalMetrics {
 	}
 }
 
-func New(ctx context.Context, target, authName string, auth *config.Auth, modules map[string]*config.Module, logger log.Logger, reg prometheus.Registerer, conc int) *Collector {
+func New(ctx context.Context, target, authName string, auth *config.Auth, modules []*NamedModule, logger log.Logger, reg prometheus.Registerer, conc int) *Collector {
 	internalMetrics := newInternalMetrics(reg)
 	return &Collector{ctx: ctx, target: target, authName: authName, auth: auth, modules: modules, logger: logger, metrics: internalMetrics, concurrency: conc}
 }
@@ -424,11 +440,11 @@ func (c Collector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- prometheus.NewDesc("dummy", "dummy", nil, nil)
 }
 
-func (c Collector) collect(ch chan<- prometheus.Metric, moduleName string, module *config.Module) {
-	logger := log.With(c.logger, "module", moduleName)
+func (c Collector) collect(ch chan<- prometheus.Metric, module *NamedModule) {
+	logger := log.With(c.logger, "module", module.name)
 	start := time.Now()
-	results, err := ScrapeTarget(c.ctx, c.target, c.auth, module, logger, c.metrics)
-	moduleLabel := prometheus.Labels{"module": moduleName}
+	results, err := ScrapeTarget(c.ctx, c.target, c.auth, module.Module, logger, c.metrics)
+	moduleLabel := prometheus.Labels{"module": module.name}
 	if err != nil {
 		level.Info(logger).Log("msg", "Error scraping target", "err", err)
 		ch <- prometheus.NewInvalidMetric(prometheus.NewDesc("snmp_error", "Error scraping target", nil, moduleLabel), err)
@@ -485,23 +501,34 @@ PduLoop:
 
 // Collect implements Prometheus.Collector.
 func (c Collector) Collect(ch chan<- prometheus.Metric) {
-	var wg sync.WaitGroup
-	sem := make(chan struct{}, c.concurrency)
-	for name, module := range c.modules {
-		sem <- struct{}{}
-		wg.Add(1)
-		go func(name string, module *config.Module) {
-			logger := log.With(c.logger, "module", name)
-			level.Debug(logger).Log("msg", "Starting scrape")
-			start := time.Now()
-			c.collect(ch, name, module)
-			duration := time.Since(start).Seconds()
-			level.Debug(logger).Log("msg", "Finished scrape", "duration_seconds", duration)
-			SnmpDuration.WithLabelValues(c.authName, name).Observe(duration)
-			<-sem
-			wg.Done()
-		}(name, module)
+	wg := sync.WaitGroup{}
+	workerCount := c.concurrency
+	if workerCount < 1 {
+		workerCount = 1
 	}
+	workerChan := make(chan *NamedModule)
+	for i := 0; i < workerCount; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for m := range workerChan {
+				logger := log.With(c.logger, "module", m.name)
+				level.Debug(logger).Log("msg", "Starting scrape")
+				start := time.Now()
+				c.collect(ch, m)
+				duration := time.Since(start).Seconds()
+				level.Debug(logger).Log("msg", "Finished scrape", "duration_seconds", duration)
+				snmpDuration.WithLabelValues(c.authName, m.name).Observe(duration)
+			}
+		}()
+	}
+
+	go func() {
+		for _, module := range c.modules {
+			workerChan <- module
+		}
+		close(workerChan)
+	}()
 	wg.Wait()
 }
 
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
@@ -42,6 +43,14 @@ var (
 	float64Mantissa uint64 = 9007199254740992
 	wrapCounters           = kingpin.Flag("snmp.wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
 	srcAddress             = kingpin.Flag("snmp.source-address", "Source address to send snmp from in the format 'address:port' to use when connecting targets. If the port parameter is empty or '0', as in '127.0.0.1:' or '[::1]:0', a source port number is automatically (random) chosen.").Default("").String()
+
+	SnmpDuration = promauto.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name: "snmp_collection_duration_seconds",
+			Help: "Duration of collections by the SNMP exporter",
+		},
+		[]string{"auth", "module"},
+	)
 )
 
 // Types preceded by an enum with their actual type.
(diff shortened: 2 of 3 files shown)


collector/collector.go
```diff
@@ -42,6 +43,14 @@ var (
 	float64Mantissa uint64 = 9007199254740992
 	wrapCounters           = kingpin.Flag("snmp.wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
 	srcAddress             = kingpin.Flag("snmp.source-address", "Source address to send snmp from in the format 'address:port' to use when connecting targets. If the port parameter is empty or '0', as in '127.0.0.1:' or '[::1]:0', a source port number is automatically (random) chosen.").Default("").String()
+
+	SnmpDuration = promauto.NewSummaryVec(
```
SuperQ, line 47: While we're doing this, maybe let's convert this to a histogram?

SuperQ (commented):
Thanks, we'll also need to update the README documentation.

We should make it clear in the docs that this implementation of multi-module handling does not do any de-duplication of walks between different modules.

servak:
Thanks for review.
Understood. I will try to add the documentation as well.
By the way, I was thinking of doing a comma delimiter as well since it will be done soon, what do you think?
https://github.com/prometheus/snmp_exporter/issues/731#issuecomment-1666815673

SuperQ:
I think supporting both multiple `module` URL params and comma separation would be just fine.

main.go
```diff
@@ -176,8 +167,11 @@ func main() {
 	kingpin.HelpFlag.Short('h')
 	kingpin.Parse()
 	logger := promlog.New(promlogConfig)
+	if *concurrency < 1 {
+		*concurrency = 1
+	}
 
-	level.Info(logger).Log("msg", "Starting snmp_exporter", "version", version.Info())
+	level.Info(logger).Log("msg", "Starting snmp_exporter", "version", version.Info(), "concurrency", concurrency)
```
SuperQ, line 174: Let's make this a separate log line. We're probably moving this particular log line into the `exporter-toolkit` package.

servak: I don't have an image of the implementation, will you be using this one?
https://pkg.go.dev/github.com/prometheus/exporter-toolkit@v0.10.0/web

collector/collector.go
```diff
@@ -466,11 +478,33 @@ PduLoop:
...
 		time.Since(start).Seconds())
 }
 
+// Collect implements Prometheus.Collector.
+func (c Collector) Collect(ch chan<- prometheus.Metric) {
+	var wg sync.WaitGroup
+	sem := make(chan struct{}, c.concurrency)
+	for name, module := range c.modules {
+		sem <- struct{}{}
+		wg.Add(1)
```
SuperQ, line 492: I'm not sure this concurrency approach is how I would do it.

The more common pattern is to startup a pool of worker routines and then send the module list down the channel to them.

Something [like this in node_exporter](https://github.com/prometheus/node_exporter/blob/37ce0bab8c7afe5133b33bb44cff06fe965cf109/collector/filesystem_linux.go#L58-L73).

What do you think?

servak: I thought it was a simple implementation for me, but I can change it if you want to match other implementations.

collector/collector.go
```diff
@@ -356,12 +365,14 @@ type internalMetrics struct {
...
-	target  string
-	auth    *config.Auth
-	module  *config.Module
-	logger  log.Logger
-	metrics internalMetrics
+	ctx         context.Context
+	target      string
+	auth        *config.Auth
+	authName    string
+	modules     map[string]*config.Module
```
servak, line 372: I just noticed that the order in which modules are acquired changes each time because of the use of a map. For some people this may not be the behavior they thought it would be.

collector/collector.go
SuperQ: Yes, I think the other method is easier to read/understand.

collector/collector.go
SuperQ: Oh, yea, maybe we need to make a wrapper type and make it a simple slice.

main.go
SuperQ: I'm working on a top-level `toolkit` package. It's [still a work in progress](https://github.com/prometheus/exporter-toolkit/pull/147).

servak:
Thank you for confirming some points.
I will make some changes to the implementation and test it, so please confirm again when I do so.

SuperQ:
Pro tip, `git config --global format.signoff true` for DCO signoffs.

candlerb:
Question: do we actually *want* the multiple module scrapes to be in parallel?  It ought to work of course, but it makes me a bit uncomfortable as I worry about tickling bugs in vendor implementations.

Taking this to an extreme, why doesn't snmp_exporter do all its snmpwalks and snmpgets in parallel for a single module (or does it in fact do this?)

SuperQ:
> Question: do we actually want the multiple module scrapes to be in parallel

Probably not most of the time. But the default concurrency of 1 is fine. If users know what they're doing, they can do more.

Really, we'll eventually want this to be a config option. I'm going to say this ahead of time, _not_ a scrape parameter. That's _very_ dangerous.

collector/collector.go
```diff
@@ -42,6 +43,14 @@ var (
 	float64Mantissa uint64 = 9007199254740992
 	wrapCounters           = kingpin.Flag("snmp.wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
 	srcAddress             = kingpin.Flag("snmp.source-address", "Source address to send snmp from in the format 'address:port' to use when connecting targets. If the port parameter is empty or '0', as in '127.0.0.1:' or '[::1]:0', a source port number is automatically (random) chosen.").Default("").String()
+
+	SnmpDuration = promauto.NewHistogramVec(
```
SuperQ, line 47: Minor nit, let's not make this a public variable.

```suggestion
	snmpDuration = promauto.NewHistogramVec(
```

SuperQ (approved):
Minor nit, otherwise, looks good.

servak:
Describe the log of the test.
You can see that the two run in parallel and the heavy if_mib starts first but the light other module completes first.

generator.yml
```
modules:
  sysUpTime:
    walk:
    - sysUpTime

  sysDescr:
    walk:
    - sysDescr

  # Default IF-MIB interfaces table with ifIndex.
  if_mib:
    walk: [interfaces, ifXTable]
    lookups:
      - source_indexes: [ifIndex]
        lookup: ifAlias
      - source_indexes: [ifIndex]
        # Uis OID to avoid conflict with PaloAlto PAN-COMMON-MIB.
        lookup: 1.3.6.1.2.1.2.2.1.2 # ifDescr
      - source_indexes: [ifIndex]
```

http request
```
> curl -s 'localhost:9116/snmp?target=192.168.64.3&module=if_mib,sysDescr,sysUpTime,sysDescr' | head
# HELP ifAdminStatus The desired state of the interface - 1.3.6.1.2.1.2.2.1.7
# TYPE ifAdminStatus gauge
ifAdminStatus{ifAlias="eth0",ifDescr="Intel Corporation 82540EM Gigabit Ethernet Controller",ifIndex="2",ifName="eth0"} 1
ifAdminStatus{ifAlias="lo",ifDescr="lo",ifIndex="1",ifName="lo"} 1
```

snmp-exporter log
```
> go run ./ --concurrency=2 --log.level=debug --config.file generator/snmp.yml
ts=2023-08-17T14:15:00.717Z caller=main.go:176 level=info msg="Starting snmp_exporter" version="(version=, branch=, revision=unknown)" concurrency=2
ts=2023-08-17T14:15:00.717Z caller=main.go:177 level=info build_context="(go=go1.20.5, platform=darwin/arm64, user=, date=, tags=unknown)"
ts=2023-08-17T14:15:00.723Z caller=tls_config.go:274 level=info msg="Listening on" address=[::]:9116
ts=2023-08-17T14:15:00.723Z caller=tls_config.go:277 level=info msg="TLS is disabled." http2=false address=[::]:9116
ts=2023-08-17T14:15:09.516Z caller=collector.go:512 level=debug auth=public_v2 target=192.168.64.3 module=if_mib msg="Starting scrape"
ts=2023-08-17T14:15:09.516Z caller=collector.go:512 level=debug auth=public_v2 target=192.168.64.3 module=sysDescr msg="Starting scrape"
ts=2023-08-17T14:15:09.516Z caller=collector.go:193 level=debug auth=public_v2 target=192.168.64.3 module=sysDescr msg="Getting OIDs" oids=1
ts=2023-08-17T14:15:09.516Z caller=collector.go:227 level=debug auth=public_v2 target=192.168.64.3 module=if_mib msg="Walking subtree" oid=1.3.6.1.2.1.2
ts=2023-08-17T14:15:09.634Z caller=collector.go:203 level=debug auth=public_v2 target=192.168.64.3 module=sysDescr msg="Get of OIDs completed" oids=1 duration_seconds=118.315333ms
ts=2023-08-17T14:15:09.635Z caller=collector.go:516 level=debug auth=public_v2 target=192.168.64.3 module=sysDescr msg="Finished scrape" duration_seconds=0.118678584
ts=2023-08-17T14:15:09.635Z caller=collector.go:512 level=debug auth=public_v2 target=192.168.64.3 module=sysUpTime msg="Starting scrape"
ts=2023-08-17T14:15:09.635Z caller=collector.go:193 level=debug auth=public_v2 target=192.168.64.3 module=sysUpTime msg="Getting OIDs" oids=1
ts=2023-08-17T14:15:09.639Z caller=collector.go:203 level=debug auth=public_v2 target=192.168.64.3 module=sysUpTime msg="Get of OIDs completed" oids=1 duration_seconds=4.183417ms
ts=2023-08-17T14:15:09.640Z caller=collector.go:516 level=debug auth=public_v2 target=192.168.64.3 module=sysUpTime msg="Finished scrape" duration_seconds=0.005313333
ts=2023-08-17T14:15:09.645Z caller=collector.go:241 level=debug auth=public_v2 target=192.168.64.3 module=if_mib msg="Walk of subtree completed" oid=1.3.6.1.2.1.2 duration_seconds=129.05475ms
ts=2023-08-17T14:15:09.645Z caller=collector.go:227 level=debug auth=public_v2 target=192.168.64.3 module=if_mib msg="Walking subtree" oid=1.3.6.1.2.1.31.1.1
ts=2023-08-17T14:15:09.664Z caller=collector.go:241 level=debug auth=public_v2 target=192.168.64.3 module=if_mib msg="Walk of subtree completed" oid=1.3.6.1.2.1.31.1.1 duration_seconds=19.145375ms
ts=2023-08-17T14:15:09.667Z caller=collector.go:516 level=debug auth=public_v2 target=192.168.64.3 module=if_mib msg="Finished scrape" duration_seconds=0.151105834
```

SuperQ:
@candlerb How do you like this compared to your version?

collector/collector.go
SuperQ: Oh, right. I prefer to not allow metrics to cross package boundaries.

How about we we have a function like `collector.InitModuleMetrics(modules []string)`?

collector/collector.go
```diff
@@ -355,13 +364,27 @@ type internalMetrics struct {
 	snmpRetries           prometheus.Counter
 }
 
+type NamedModule struct {
+	*config.Module
+	name string
+}
```
candlerb, line 370: Is there are point to keep Module and NamedModule separate, rather than add a "name" attribute to Module?

main.go
```diff
@@ -94,44 +88,43 @@ func handler(w http.ResponseWriter, r *http.Request, logger log.Logger) {
...
 	}
-	if !moduleOk {
-		http.Error(w, fmt.Sprintf("Unknown module '%s'", moduleName), http.StatusBadRequest)
-		snmpRequestErrors.Inc()
-		return
+	uniqueM := make(map[string]bool)
+	var modules []*collector.NamedModule
+	for _, qm := range queryModule {
+		for _, m := range strings.Split(qm, ",") {
+			module, moduleOk := sc.C.Modules[m]
```
candlerb, line 107: Perhaps skip if m=="", in order to handle cases like `foo,,bar` ?  And then at the very end, `if len(modules)==0 { modules=[]string{"if_mib"} }` (instead of lines 92-93)

candlerb:
It's a bigger change - for example it creates per-module scrape stats - but if that's OK with you then it's OK with me :-)

I can't see any circumstance in which I'd use concurrency>1 (as it can't be controlled on a target-by-target basis), but it doesn't add much complexity to the code, so it's fine as-is.

collector/collector.go
servak: Like this?
```
diff --git a/collector/collector.go b/collector/collector.go
index 12770d1..3f0b33b 100644
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -44,7 +44,7 @@ var (
        wrapCounters           = kingpin.Flag("snmp.wrap-large-counters", "Wrap 64-bit counters to avoid floating point rounding.").Default("true").Bool()
        srcAddress             = kingpin.Flag("snmp.source-address", "Source address to send snmp from in the format 'address:port' to use when connecting targets. If the port parameter is empty or '0', as in '127.0.0.1:' or '[::1]:0', a source port number is automatically (random) chosen.").Default("").String()

-       SnmpDuration = promauto.NewHistogramVec(
+       snmpDuration = promauto.NewHistogramVec(
                prometheus.HistogramOpts{
                        Name: "snmp_collection_duration_seconds",
                        Help: "Duration of collections by the SNMP exporter",
@@ -89,6 +89,10 @@ func listToOid(l []int) string {
        return strings.Join(result, ".")
 }

+func InitModuleMetrics(auth, module string) {
+       snmpDuration.WithLabelValues(auth, module)
+}
+
 type ScrapeResults struct {
        pdus    []gosnmp.SnmpPDU
        packets uint64
diff --git a/main.go b/main.go
index 3830afd..196f440 100644
--- a/main.go
+++ b/main.go
@@ -155,7 +155,7 @@ func (sc *SafeConfig) ReloadConfig(configFile string) (err error) {
        // Initialize metrics.
        for auth := range sc.C.Auths {
                for module := range sc.C.Modules {
-                       collector.SnmpDuration.WithLabelValues(auth, module)
+                       collector.InitModuleMetrics(auth, module)
                }
        }
        sc.Unlock()
```

collector/collector.go
servak: I thought a light wrapper would be better than playing around with the Unmarshal process.

candlerb: OK, that makes sense.

collector/collector.go
SuperQ: I was more thinking `collector.InitModuleMetrics(sc.C.Auths, sc.C.Modules)`.

SuperQ:
I would have been previously more objecting to the per-module stats. But with the new separated auths, the number of modules will likely be greatly reduced in real-world installs.

main.go
```diff
@@ -43,20 +43,14 @@ import (
 var (
 	configFile  = kingpin.Flag("config.file", "Path to configuration file.").Default("snmp.yml").String()
 	dryRun      = kingpin.Flag("dry-run", "Only verify configuration is valid and exit.").Default("false").Bool()
+	concurrency = kingpin.Flag("concurrency", "Specify the number of modules to fetch concurrently").Default("1").Int()
```
SuperQ, line 46: Another minor nit, we try to namespace our flag names.

```suggestion
	concurrency = kingpin.Flag("snmp.module-concurrency", "The number of modules to fetch concurrently per scrape").Default("1").Int()
```

servak: fixed

README.md
```diff
@@ -89,6 +89,25 @@ Note that [URL encoding](https://en.wikipedia.org/wiki/URL_encoding) should be u
 to the `:` and `/` characters. Prometheus encodes query parameters automatically and manual encoding
 is not necessary within the Prometheus configuration file.
 
+## Multi-Module Handling
+The multi-module functionality allows you to specify multiple modules, enabling the retrieval of information from several modules in a single scrape.
+The concurrency can be specified using the snmp-exporter option --concurrency (the default is 1).
```
SuperQ, line 94:
```suggestion
The concurrency can be specified using the snmp-exporter option `--snmp.module-concurrency` (the default is 1).
```

servak: Oh, I missed that.

SuperQ (approved):
Awesome

collector/collector.go
```diff
@@ -466,11 +498,44 @@ PduLoop:
...
+				start := time.Now()
+				c.collect(ch, m)
+				duration := time.Since(start).Seconds()
+				level.Debug(logger).Log("msg", "Finished scrape", "duration_seconds", duration)
+				snmpDuration.WithLabelValues(c.authName, m.name).Observe(duration)
+			}
+		}()
+	}
+
+	go func() {
```
bwplotka, line 530: Nit: I don't think we need this goroutine, it does not give us anything. We could use "Collect(" (this method) scope to distribute the work and close channel afterwards and then wait. Only readability/complexity nit though.

collector/collector.go
```diff
@@ -466,11 +498,44 @@ PduLoop:
...
 	}
 	ch <- prometheus.MustNewConstMetric(
-		prometheus.NewDesc("snmp_scrape_duration_seconds", "Total SNMP time scrape took (walk and processing).", nil, nil),
+		prometheus.NewDesc("snmp_scrape_duration_seconds", "Total SNMP time scrape took (walk and processing).", nil, moduleLabel),
 		prometheus.GaugeValue,
 		time.Since(start).Seconds())
 }
 
+// Collect implements Prometheus.Collector.
+func (c Collector) Collect(ch chan<- prometheus.Metric) {
```
bwplotka, line 507: I assume there is a unit test for this somewhere?

bwplotka (commented):
Not familiar with full codebase, but asked to review concurrency go - one nit, but overall looks good!

collector/collector.go
servak: Indeed, the last Go routine seems unnecessary. Could you please confirm if my understanding is correct?
```
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -527,12 +527,10 @@ func (c Collector) Collect(ch chan<- prometheus.Metric) {
                }()
        }

-       go func() {
-               for _, module := range c.modules {
-                       workerChan <- module
-               }
-               close(workerChan)
-       }()
+       for _, module := range c.modules {
+               workerChan <- module
+       }
+       close(workerChan)
        wg.Wait()

```

bwplotka: Correct. Do you mind writing quick unit test for it?

servak: Thanks. 
I think that unit test for Collector is difficult.
Because Collector does not have an SNMP Fetcher interface, but calls GoSNMP directly in ScrapeTargets.
Please let me know if I am not aware of this and if there is another way to make this easier.

SuperQ: I agree, writing a test for this is a bit difficult. I'm OK without an additional test here.

Issue #736, opened 2022-03-10
Enable Sentry automatic trace report


mirceaulinic:
I'm personally a big fan of [Sentry](https://sentry.io/) as it significantly simplifies my life when it comes to detecting the root cause of a bug, or when trying to figure what's the (use-specific) cause of a failure.

I would therefore like to file a feature request in order to add native support in this exporter. Enabling is quite straight forward (see https://docs.sentry.io/platforms/go/). Of course, the usage would be completely optional (i.e., typically only attempt to send traces when the `SENTRY_DSN` environment variable is configured). I'm happy to work on adding this myself, just wanted to discuss it here beforehand.

mirceaulinic:
Opened #946 to address this proposal.

SuperQ:
Sentry is a proprietary form of logging/tracing. We don't intend to support this.

PR #946, opened 2023-08-16
Enhance error reporting via automatic Sentry traces

mirceaulinic:
Since I've opened #736 I didn't receive much feedback, and I've been running an internal fork where I have this code that I've been running for aprox 1 year.
Opening this PR to hopefully hears some feedback whether this is something the maintainers are interested into.
Personally, having this, massively helped me understand what's the cause of the SNMP polling failures under very particular conditions (hard to reproduce in a testing environment otherwise), in an environment of thousands of devices, where manual spot-checks obviously wouldn't work.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
+	"github.com/getsentry/sentry-go"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gosnmp/gosnmp"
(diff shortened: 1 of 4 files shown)


SuperQ:
This would be more of a question for the prometheus-developers list, as we want to keep our exporter design patterns as consistent as possible between each exporter.

Personally, I don't see Sentry as being valuable. If you want to capture issues it should go into one of three categories:
* Metrics
* Logs
* Traces

Preferably in that order.

In addition, our codebase is in Go. Go is not an exception handling language, so the typical value that Sentry provides by capturing random uncaught exceptions just doesn't fit the way it does for Python, Ruby, Java.

mirceaulinic:
I think I disagree with that view, since at large scale neither of metrics, logs or traces can pin-point you to the exact root cause of the failures (across a group of devices / cluster, etc.). That said however, I won't insist, and respect your decision as maintainers. :)

SuperQ:
Sentry is a proprietary form of logging/tracing. We don't intend to support this.

Issue #950, opened 2023-08-16
snmp_packet stats are all zeros


candlerb:
### Host operating system: output of `uname -a`

```
Linux prometheus 5.4.0-153-generic #170-Ubuntu SMP Fri Jun 16 13:43:31 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

### snmp_exporter version: output of `snmp_exporter -version`

```
snmp_exporter, version 0.23.0 (branch: HEAD, revision: 20657d2f486a1ac237238827beed97fa67eb8611)
  build user:       root@f2310b8dd92c
  build date:       20230816-09:56:45
  go version:       go1.20.7
  platform:         linux/amd64
  tags:             netgo
```

### What did you do that produced an error?

Simple scrape (do it multiple times)

### What did you expect to see?

The histogram snmp_packet_duration_seconds_bucket to increment; the counter snmp_packets_total to increment

(Note: these appear to be new. I tried with 0.21 and these metrics don't exist at all)

### What did you see instead?

All metrics are zero, even after repeated scrapes:

```
# HELP snmp_packet_duration_seconds A histogram of latencies for SNMP packets.
# TYPE snmp_packet_duration_seconds histogram
snmp_packet_duration_seconds_bucket{le="0.0001"} 0
snmp_packet_duration_seconds_bucket{le="0.0002"} 0
snmp_packet_duration_seconds_bucket{le="0.0004"} 0
snmp_packet_duration_seconds_bucket{le="0.0008"} 0
snmp_packet_duration_seconds_bucket{le="0.0016"} 0
snmp_packet_duration_seconds_bucket{le="0.0032"} 0
snmp_packet_duration_seconds_bucket{le="0.0064"} 0
snmp_packet_duration_seconds_bucket{le="0.0128"} 0
snmp_packet_duration_seconds_bucket{le="0.0256"} 0
snmp_packet_duration_seconds_bucket{le="0.0512"} 0
snmp_packet_duration_seconds_bucket{le="0.1024"} 0
snmp_packet_duration_seconds_bucket{le="0.2048"} 0
snmp_packet_duration_seconds_bucket{le="0.4096"} 0
snmp_packet_duration_seconds_bucket{le="0.8192"} 0
snmp_packet_duration_seconds_bucket{le="1.6384"} 0
snmp_packet_duration_seconds_bucket{le="+Inf"} 0
snmp_packet_duration_seconds_sum 0
snmp_packet_duration_seconds_count 0
[...]

PR #971, opened 2023-08-28
Fix exporter metrics

SuperQ:
Move exporter metric definitions and registration out of the collector
package and into the main package. This fixes metrics always being empty
due to the epehemeral collector registry.

Fixes: https://github.com/prometheus/snmp_exporter/issues/950


candlerb:
Erm, I'm not sure this is complete.

`collector.New()` is called on every scrape, and it creates (and registers) a new instance of internalMetrics each time:

```
func New(... reg prometheus.Registerer ...) *Collector {
        internalMetrics := newInternalMetrics(reg)
        return &Collector{ctx: ctx, target: target, authName: authName, auth: auth, modules: modules, logger: logger, metrics: internalMetrics, concurrency: conc}
}
```

SuperQ:
Yea, I realized it after I pushed. Reworked, tested locally.

SuperQ:
:eyeglasses: @servak

I realized I made a mistake in the review of the multi-module support. The metrics for the SNMP data actually belong to the `main` package, since they're exporter-level stats. Move the definitions of these metrics back to the main package and pass them in as handles.

This is similar to what we did in the blackbox_exporter in order to make the collection packages reusable as a library.

https://github.com/prometheus/blackbox_exporter/pull/1060

servak:
@SuperQ Thanks for letting me know.

Memo:
scrape metrics -> /snmp
global metrics → /metrics

ExporterMetrics basically has global metrics, but only snmp_collection_duration_seconds is left in collector, is there any reason for that?
If snmp_collection_duration_seconds is moved to ExporterMetrics, the collector.InitModuleMetrics function can be removed.

SuperQ:
@servak Finished moving the collection duration metric to main.

servak:
@SuperQ LGTM

roidelapluie (approved):

roidelapluie:
Thanks!

PR #973, opened 2023-08-29
Remove broken MIB from generator

SuperQ:
The `WIENER-CRATE-MIB` can not be downloaded reliably. Remove this from the generator example config to avoid CI failures.


RichiH:
LGTM

roidelapluie (approved):

RichiH:
I am wondering if we should do `curl --insecure` after all... We're only hurting the users, not the company..

SuperQ:
Now that we have multi-file, we can move more of these to the prometheus-community/snmp repo. I plan to create generator fragments rather than a monolithic config. This will make it less fragile.

PR #915, opened 2023-07-13
Add Mikrotik indexes

willglynn:
This PR adds several Mikrotik indexes, causing the following changes:

* `mtxrGaugeIndex`

```diff
-mtxrGaugeValue{mtxrGaugeIndex="17"} 54
-mtxrGaugeValue{mtxrGaugeIndex="7101"} 45
-mtxrGaugeValue{mtxrGaugeIndex="7201"} 501
+mtxrGaugeValue{mtxrGaugeName="board-temperature1"} 45
+mtxrGaugeValue{mtxrGaugeName="cpu-temperature"} 54
+mtxrGaugeValue{mtxrGaugeName="jack-voltage"} 501
```

* `mtxrNeighborIndex`

```diff
-mtxrNeighborMacAddress{mtxrNeighborIndex="1",mtxrNeighborMacAddress="48:8F:5A:17:D6:14"} 1
-mtxrNeighborMacAddress{mtxrNeighborIndex="2",mtxrNeighborMacAddress="AC:CC:8E:F7:DE:6B"} 1
+mtxrNeighborMacAddress{mtxrNeighborIpAddress="172.16.4.10",mtxrNeighborMacAddress="48:8F:5A:17:D6:14"} 1
+mtxrNeighborMacAddress{mtxrNeighborIpAddress="172.16.4.30",mtxrNeighborMacAddress="AC:CC:8E:F7:DE:6B"} 1
```

* `mtxrOpticalIndex`

```diff
-mtxrOpticalSupplyVoltage{mtxrOpticalIndex="9"} 3289
-mtxrOpticalTxPower{mtxrOpticalIndex="9"} -6343
-mtxrOpticalWavelength{mtxrOpticalIndex="9"} 131000
+mtxrOpticalSupplyVoltage{mtxrOpticalIndex="9",mtxrOpticalName="sfpplus1"} 3289
+mtxrOpticalTxPower{mtxrOpticalIndex="9",mtxrOpticalName="sfpplus1"} -6305
+mtxrOpticalWavelength{mtxrOpticalIndex="9",mtxrOpticalName="sfpplus1"} 131000
```

* `mtxrPOEInterfaceIndex`

```diff
-mtxrPOEStatus{mtxrPOEInterfaceIndex="1"} 2
-mtxrPOEStatus{mtxrPOEInterfaceIndex="2"} 2
-mtxrPOEStatus{mtxrPOEInterfaceIndex="3"} 5
[...]

PR #986, opened 2023-09-01
Updated Mikrotik neighbor indexes make them unique

jvwag:
In response to PR #915, this change will update the keys of mtxrNeighborTable to unique label combinations. The current release will return http 500 replies on request with variations on following errors:

`* collected metric "mtxrNeighborInterfaceID" { label:{name:"mtxrNeighborIpAddress" value:"0.0.0.0"} gauge:{value:1}} was collected before with the same name and label values`

This PR will change the index from mtxrNeighborIpAddress to a combination of mtxrNeighborMacAddress and ifName.

The choice for mtxrNeighborIpAddress as unique index is not suitable for two reasons:
- Mikrotik allows for entries in the neighbor list without an IP adress, or only an IPv6 address.
- Same IP addresses can be assigned to neighbors, making them not unique.

The obvious solution would be to make the MAC address the unique key, but this is also not sufficient because a neighbour can be detected on multiple interfaces (which is useful information and should not be filtered out).

A combination of mac address and interface is unique enough and will benefit in having a more consistent set of data after device reboots.

Examples:

Change from IpAddress to ifName
```diff
-mtxrNeighborMacAddress{mtxrNeighborIpAddress="10.0.0.1",mtxrNeighborMacAddress="48:A9:82:00:00:01"} 1
-mtxrNeighborMacAddress{mtxrNeighborIpAddress="10.0.0.2",mtxrNeighborMacAddress="48:A9:82:00:00:02"} 1
+mtxrNeighborMacAddress{ifName="ether1",mtxrNeighborMacAddress="48:A9:82:00:00:01"} 1
+mtxrNeighborMacAddress{ifName="ether1",mtxrNeighborMacAddress="48:A9:82:00:00:02"} 1
```

Previously impossible:
```diff
-mtxrNeighborIpAddress{mtxrNeighborIpAddress="10.0.0.1",mtxrNeighborMacAddress="48:A9:82:00:00:01"} 1
-mtxrNeighborIpAddress{mtxrNeighborIpAddress="10.0.0.2",mtxrNeighborMacAddress="48:A9:82:00:00:02"} 1
+mtxrNeighborIpAddress{ifName="ether1",mtxrNeighborMacAddress="48:A9:82:00:00:01",mtxrNeighborIpAddress="0.0.0.0"} 1
+mtxrNeighborIpAddress{ifName="vlan14-test",mtxrNeighborMacAddress="48:A9:82:00:00:01",mtxrNeighborIpAddress="10.0.0.1"} 1
+mtxrNeighborIpAddress{ifName="vlan14-test",mtxrNeighborMacAddress="48:A9:82:00:00:02",mtxrNeighborIpAddress="10.0.0.2"} 1
```
Example Mikrotik Neighbor table in Winbox interface
image

@SuperQ @RichiH

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -569,7 +569,12 @@ modules:
         lookup: mtxrGaugeName
         drop_source_indexes: true
       - source_indexes: [mtxrNeighborIndex]
-        lookup: mtxrNeighborIpAddress
+        lookup: mtxrNeighborMacAddress
+        drop_source_indexes: true
+      - source_indexes: [mtxrNeighborIndex]
+        lookup: mtxrNeighborInterfaceID
+      - source_indexes: [mtxrNeighborInterfaceID]
+        lookup: ifName
         drop_source_indexes: true
       - source_indexes: [mtxrOpticalIndex]
         lookup: mtxrOpticalName
--- a/snmp.yml
+++ b/snmp.yml
@@ -14735,14 +14735,28 @@ modules:
       indexes:
       - labelname: mtxrNeighborIndex
         type: gauge
+      - labelname: mtxrNeighborInterfaceID
+        type: gauge
       lookups:
       - labels:
         - mtxrNeighborIndex
-        labelname: mtxrNeighborIpAddress
-        oid: 1.3.6.1.4.1.14988.1.1.11.1.1.2
-        type: InetAddressIPv4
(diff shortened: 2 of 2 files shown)


SuperQ (approved):
Makes sense, thanks!

Issue #996, opened 2023-09-08
Generator Makefile make mibs fails APC_URL 404 again


fue36:
APC_URL 404

In the generator Makefile, the APC_URL link to schneider-electric gives a 404.

https://download.schneider-electric.com/files?p_Doc_Ref=APC_POWERNETMIB_441_EN&p_enDocType=Firmware&p_File_Name=powernet441.mib

I couldn't locate another link for the mib from schneider electric.  However I found one closer to home. ;-)

https://github.com/prometheus/snmp_exporter/files/11223248/powernet446.mib.txt

```
$ git describe --tags
v0.24.1-5-g7fa8c4d
```

SuperQ:
Yes, it's now 4.4.6 or 4.4.9.

SuperQ:
Or, ugh, all the downloads are broken.

SirOch27:
hello

is there a workaound until the problem is solved ?

robbyleh:
Same issue here...
git describe --tags
v0.24.1-5-g7fa8c4d

make mibs
>> Downloading apc-powernet-mib
curl: (22) The requested URL returned error: 404
make: *** [Makefile:138: mibs/apc-powernet-mib] Error 22

SuperQ:
There is no workaround, APC has broken all MIB downloads. The only thing we could remove support for APC.

RichiH:
We could keep the old config around and splice it into the generated config.

Sent by mobile; please excuse my brevity.

gebn:
Frustratingly, it looks like only the newest download works: https://github.com/prometheus/snmp_exporter/pull/1003

PR #1003, opened 2023-09-16
Update PowerNet MIB to v4.5.1

gebn:
Fixes `make mibs`. It looks like APC remove old download links as soon as a new version is published, so this is a game of whack-a-mole until there's a mechanism for finding the latest version.

Fixes: https://github.com/prometheus/snmp_exporter/issues/996


SuperQ (approved):
Thanks, I thought 4.4.9 was the latest. The APC site is so broken, this didn't show up in the search for me.

Issue #734, opened 2022-02-23
Alternative to regex_extracts for scaled integer values


ghost:
I'm adapting some telecom power plant MIBs for monitoring with snmp_exporter. Since many SNMP implementations don't normally support floats, the manufacturers usually present these values scaled (milivolts instead of volts, tenths of amps instead of amps, etc).  
For example, to get DC output voltage in volts from milivolts, this would be the regex_extracts in generator.yml (that turns out similar in snmp.yml):
```
  overrides:
      u1DCOutputVoltage:
        regex_extracts:
          '':
            - regex: '(\d*?)(\d{3})'
              value: '$1.$2'
        type: DisplayString
```
I took me quite a while to reach this definition, after trying several regex strings, before I realized that I also needed to override the type to DisplayString, because the MIB defined the variable as INTEGER32.

By the way, this also seems to be a common use case for regex_extracts, based on the examples I found on the web.

As an easier self documenting alternative, I suggest the introduction of "factor" overrides, that simply multiplies the value,  for values that already are numbers.  It is also more intuitive than regex (at least for us, electrical engineers). The syntax could be:
```
  overrides:
      u1DCOutputVoltage:
        factor: 0.001
```
I guess that it would also have a positive impact in scrape processing performance.

Regards, Marco

readpe:
I am also just looking for a similar feature. I've run into a use case where I'd like to scale the same metrics differently in each module due to differences in manufacturer snmp values. I tried the following:

Example:

```go
type Metric struct {
        ...
	Scale          float64                    `yaml:"scale"`
	...
}
```

```go
func pduToSamples(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, oidToPdu map[string]gosnmp.SnmpPDU, logger log.Logger) []prometheus.Metric {
	...
	// Scale metric.
	if metric.Scale != 0 {
		value *= metric.Scale
	}
        ...
```

```yaml
  - name: rPDU2DeviceStatusPower
    oid: 1.3.6.1.4.1.318.1.1.26.4.3.1.5
    type: gauge
    scale: 0.01
    help: The power consumption of the Rack PDU load in -hundredths- of kilowatts -
      1.3.6.1.4.1.318.1.1.26.4.3.1.5
    indexes:
    - labelname: rPDU2DeviceStatusIndex
      type: gauge
```

Thank you

SuperQ:
This is something you handled at the PromQL side of things.

[...]

PR #1026, opened 2023-10-18
Add a scaling factor

SuperQ:
Allow the metric value to be pre-scaled based on a float64 multiplier. This allows for simple unit conversions like bits to bytes, centi-degrees, etc.

Fixes: https://github.com/prometheus/snmp_exporter/issues/734

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ dependencies-stamp
 
 generator/generator
 generator/mibs
+generator/snmp.yml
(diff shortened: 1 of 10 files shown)


SuperQ:
CC @bastischubert

bastischubert:
very nice, one thing that i noticed is that we do almost the same with modbus_exporter but named it "factor" instead of "scale"

matthiasr (approved):

matthiasr:
Conversely, in the statsd exporter we use `scale` 😂

RichiH:
I feel as if modbus and SNMP are closer in userbase than statsd.

Else LGTM

freeyssu:
Very nice feature that all want to get it.
When would this MR be released? Do you guys have a plan?

PR #1028, opened 2023-10-18
Enable passing input file, output file, and mibs dir on command line …

dnck:
# Description

@SuperQ @RichiH :wave: 

Hi there, I think this is a fairly trivial PR, but lmk what you think. 

What's it do?

In order to facilitate the organization of snmp.ymls in different directories, I wanted a way to specify 1) the path to the 
`generator.yml`, 2) the expected output `snmp.yml`, and 3) the directory of mibs. 

For this reason, I added a few flags to the cli `generator generate` command group. 

With the PR, you can do something like, 

```bash 
  ./generator generate \
    -g snmp-exporter/generator/myDevices/newManufacturer/generator.yml \
    -o snmp-exporter/generator/myDevices/newManufacturer/snmp.yml \ 
    -m snmp-exporter/generator/myDevices/shared-mibs \
    -m snmp-exporter/generator/myDevices/newManufacturer
```

which is helpful for me, and I hope for some others as well.


SuperQ (commented):
I'm generally in favor.

What do you think about implementing this with `Strings()` so that multiple files can be supported? Similar to what we now do in the exporter itself?

dnck:
> I'm generally in favor.
> 
> What do you think about implementing this with `Strings()` so that multiple files can be supported? Similar to what we now do in the exporter itself?

I think that makes sense for the `--mibs-dir` option, for sure. I can add that!

But, would it even make sense for the `--generator-path` and `--output-path` flags? Probably not. Unless I'm missing something?

SuperQ:
For Generator path, yea, I think multi-file would be nice. Output, no, a single output file is fine.

generator/net_snmp.go
```diff
@@ -157,24 +157,48 @@ var (
...
+	if *userMibsDir != "" {
+		// The user told us where to load the mibs
+		err = level.Info(logger).Log("msg", "Loading MIBs", "from", *userMibsDir)
+		if err != nil {
+			return "", err
+		}
+		C.netsnmp_set_mib_directory(C.CString(*userMibsDir))
+	} else {
+		// Help the user find their MIB directories.
+		err = level.Info(logger).Log("msg", "Loading MIBs", "from", C.GoString(C.netsnmp_get_mib_directory()))
```
SuperQ, line 173: We don't need to worry about `err` from the logger. The error as actually never tripped due to the way the library works. This is why we ignore this in the golanci-lint config in the project.

Also, do we need to duplicate the log lines? Shouldn't `C.netsnmp_get_mib_directory()` return the correct thing after `C.netsnmp_set_mib_directory()`?

dnck: > The error as actually never tripped due to the way the library works.

:facepalm: Thanks! I'll fix those places where I somewhat mindlessly checked for the err.

dnck: >  Shouldn't `C.netsnmp_get_mib_directory()` return the correct thing after `C.netsnmp_set_mib_directory()`?

Hmm, I'll check that. I would expect that it should.

dnck:
@SuperQ - Thanks for your patience. I think this is now ready for another review.

generator/tree.go
```diff
@@ -384,7 +393,10 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 					if prevType == subtype {
 						metric.Indexes = metric.Indexes[:len(metric.Indexes)-1]
 					} else {
-						level.Warn(logger).Log("msg", "Can't handle index type on node, missing preceding", "node", n.Label, "type", index.Type, "missing", subtype)
+						err := level.Warn(logger).Log("msg", "Can't handle index type on node, missing preceding", "node", n.Label, "type", index.Type, "missing", subtype)
```
SuperQ, line 396: Same with these, please remove all the extraneous err checks you've added.

SuperQ (changes requested):

generator/tree.go
dnck: I think I got them all now

SuperQ (approved):
Thanks!

PR #1053, opened 2023-11-24
Add Netgear ReadyNAS and ReadyData appliances support

reynico:
Follow-up of https://github.com/prometheus/snmp_exporter/pull/895


SuperQ:
I think you need to rebuild your generator binary

reynico:
I had to run the generator from a Linux machine. Seems like there's something off with the MacOS builds.

SuperQ (approved):

PR #1054, opened 2023-11-27
chore: allow `make docker-generate` to work with SELINUX

jvillal-amp:


SuperQ (changes requested):
Can you provide a little more description of why this is needed?

jvillal-amp:
> Can you provide a little more description of why this is needed?

Done @SuperQ 

Thanks.

SuperQ (approved):

hhromic:
This part in the linked documentation is a bit scary:

> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

I know is a bit late (this PR was already merged), but doing it like `etcd` does wouldn't be a safer approach?

<https://github.com/kubernetes/kubernetes/blob/ad9b60e2c9ddb21e8b00cabbe27e639638a0ea88/cluster/images/etcd/Makefile#L76-L81>

Note that they use `:z` instead of `:Z`.
Disclaimer: I am not a user of SELinux and I am not super familiar with these options and their implications.

SuperQ:
@hhromic Yea, I agree, I should have read the Docker documentation more closely. Would you mind opening a fix PR?

hhromic:
Sure, I will open a PR with that approach. 👍 
I hope @jvillal-amp can confirm if using `:z` instead of `:Z` solves the issue for them.

PR #1055, opened 2023-11-27
chore: do not force `:Z` flag in Docker volume mount

hhromic:
Follow up on #1054
This PR adopts the approach seen in `etcd` for compatibilty with SELinux:
<https://github.com/kubernetes/kubernetes/blob/ad9b60e2c9ddb21e8b00cabbe27e639638a0ea88/cluster/images/etcd/Makefile#L76-L81>

Instead of forcing `:Z`, detect if SElinux is enabled and add `:z` on demand.
According to the documentation and multiple examples I could find, using `:z` seems to be less dangerous than `:Z`.

@jvillal-amp what is your opinion on using `:z` vs `:Z`?
Can you confirm if this PR still allows you to use `make docker-generate` in your SELinux environmnet?

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -24,6 +24,12 @@ DOCKER_REPO       ?= prom
 
 SANITIZED_DOCKER_IMAGE_TAG := $(subst +,-,$(DOCKER_IMAGE_TAG))
 
+SELINUX_ENABLED := $(shell cat /sys/fs/selinux/enforce 2> /dev/null || echo 0)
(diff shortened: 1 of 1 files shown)


jvillal-amp:
> @jvillal-amp, how does this look to you?

Looks good to me. Thanks.

jvillal-amp:
As a note I tested this today and works fine for me on my Fedora system with SELINUX enabled. Thanks.

Issue #459, opened 2019-10-11
Environment / Credential injection for SNMP Generator


cameronkerrnz:
### Host operating system: output of `uname -a`

Host: Red Hat Enterprise Linux Server release 7.6 (Maipo)
Docker:  version 1.13.1
Container images: docker.io/prom/snmp-generator:master 8925d142d366 8 days ago

### snmp_exporter version: output of `snmp_exporter -version`

This relates to the generator; there is no version, and only 'master'. The docker image has image 8925d142d366, and is currently the latest available.

### Feature Request

generator.yml produces snmp.yml, and both contains credentials (in plain-text) that are passwords (in SNMPv3 at least they are properly considered passwords or shared keys).

As a deployer of Prometheus I want to use automate the deployment using tools such as Docker (in this particular case, docker-compose).

I need a way to safely inject credentials into the execution of the generator so I don't end up with credentials leaking into Git.

To illustrate, I'm working on the following:

```
modules:
  fortinet_fortianalyzer_production:
    auth:
      username: snmp_faz
      security_level: authPriv
      auth_protocol: MD5
      priv_protocol: AES
      password: XXX
      priv_password: XXX
    walk:
      - 1.3.6.1.4.1.12356.103.0.1.3  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmLogRate
      - 1.3.6.1.4.1.12356.103.2.1.1  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysCpuUsage
      - 1.3.6.1.4.1.12356.103.2.1.2  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysMemUsed
      - 1.3.6.1.4.1.12356.103.2.1.3  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysMemCapacity
      - 1.3.6.1.4.1.12356.103.2.1.4  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysDiskUsage
      - 1.3.6.1.4.1.12356.103.2.1.5  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmSysDiskCapacity
      - 1.3.6.1.4.1.12356.103.7.1.2  # FORTINET-FORTIMANAGER-FORTIANALYZER-MIB::fmRaidState
```

[...]

PR #1074, opened 2023-12-13
Support for passing username, password & priv_password as env vars

harshavmb:
Hi,

This PR aims to pass sensitive data such as `username`, `password` & `priv_password` as environment variables. 

The exporter binary looks for the environment variable presence, if present it fetches the value & pass the same as configuration to the binary. 

I know in the past [#459](https://github.com/prometheus/snmp_exporter/issues/459) was closed. Nevertheless, times have changed & new set of apps inject env vars in a safe manner. For instance Azure functions have a clever way of injecting keyvault references as env vars for the apps running in their environment. In this case, there is no need of resolving the template or writing unnecessary entrypoint scripts to circumvent them. 

Looking forward to have your feedback on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,14 +165,14 @@ This requires the `--config.expand-environment-variables` flag be set.
 
 ```YAML
 auths:
-  with_secret:
+  example_with_envs:
     community: mysecret
     security_level: SomethingReadOnly
-    username: ${env.ARISTA_USERNAME}
-    password: ${env.ARISTA_PASSWORD}
+    username: ${ARISTA_USERNAME}
+    password: ${ARISTA_PASSWORD}
     auth_protocol: SHA256
     priv_protocol: AES
-    priv_password: ${env.ARISTA_PRIV_PASSWORD}
+    priv_password: ${ARISTA_PRIV_PASSWORD}
 ```
 
 Similarly to [blackbox_exporter](https://github.com/prometheus/blackbox_exporter),


candlerb:
Now that auths are completely separated out, and snmp_exporter can read multiple config files, I see less of a need for this. For example, in Kubernetes you can create a secret with all the auths you need, and load it in addition to the main snmp.yml.

harshavmb:
Hi @candlerb,

Thanks for your feedback. As I mentioned in the description, we are targeting this deployment on Azure functions where in we can inject credentials directly as env vars. 

Many apps support credentials to be loaded via environment variables. Are there any constraints wrt., snmp_exporter with this approach I am not aware of?

SuperQ (changes requested):
In order to support this, I would like to see two changes.
* This should be behind a flag, disabled by default, for backwards compatibility. `--config.expand-environment-variables`.
* Use https://pkg.go.dev/os#Expand instead of a custom parser.

This would make it better match the Prometheus behavior (https://github.com/prometheus/prometheus/pull/8649).

SuperQ:
Are you still interested in implementing this? I think it would be very much appreciated.

harshavmb:
Hi @SuperQ,

I am on vacation & will be back next week. I shall rework this #PR as per your review. 

I hope it's okay for you. 

Thanks,
Harsha

harshavmb:
Hi @SuperQ ,

I've made added `--config.expand-environment-variables` config (disabled by default) to resolve env vars & removed custom parser. 

Could you please share your feedback?

SuperQ (changes requested):
Some minor documentation issues. Otherwise looks good.

harshavmb:
Hi @SuperQ ,

I made amendments to README.md docs as per the review. 

Thanks again for your support here.

SuperQ (approved):

RobertBergman:
This change broke SNMPv3 by removing the lines that need to go in config.go after 155 and 171 respectively:
  usm.AuthenticationPassPhrase = string(c.Password)

and 

 usm.PrivacyPassphrase = string(c.PrivPassword)

SuperQ:
@RobertBergman Oops, I'm surprised this didn't show up in our unit tests. Do you want to open a fix PR?

RobertBergman:
@SuperQ sorry, I'm not familiar with the process.

RobertBergman:
I figured it out.  PR submitted

README.md
```diff
@@ -123,6 +123,8 @@ using SNMP v2 GETBULK.
 The `--config.file` parameter can be used multiple times to load more than one file.
 It also supports [glob filename matching](https://pkg.go.dev/path/filepath#Glob), e.g. `snmp*.yml`.
 
+The `--config.expand-environment-variables` parameter allows passing environment variables into some fields of the configuration file. The `username`, `password` & `priv_password` fields in the auths section are supported. Defaults to disabled.
+
```
candlerb, line 127: I wonder if "community" should be allowed too, for consistency? (Although this may interact with community having a default value of "public")

RobertBergman: I see this and without diving deep think that SNMPv3 doesn't use community.

candlerb: You are correct. I mean for use with v1/v2c, which some people still do use, and requires a "secret" community (albeit sent over the wire in clear text)

PR #1084, opened 2024-01-02
Fix documented default value for community

PierreF:
The documentation (README) say that the default value for the SNMP community is `"public_v2"`. But I believe this is wrong:
* https://github.com/prometheus/snmp_exporter/blob/9c42d6c874d479314e612bca69558c81f8e26287/snmp.yml#L4
* https://github.com/prometheus/snmp_exporter/blob/9c42d6c874d479314e612bca69558c81f8e26287/config/config.go#L52

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -86,8 +86,8 @@ auths:
     version: 2  # SNMP version to use. Defaults to 2.
                 # 1 will use GETNEXT, 2 and 3 use GETBULK.
 
-    # Community string is used with SNMP v1 and v2. Defaults to "public_v2".
-    community: public_v2
+    # Community string is used with SNMP v1 and v2. Defaults to "public".
+    community: public
 
     # v3 has different and more complex settings.
     # Which are required depends on the security_level.


SuperQ (changes requested):
The default is actually implemented in `main.go`. The struct default is ignored since it's actually part of the URL parm parsing.

If you `curl` the exporter with no `auth` params, you can see it in the debug log.

```
$ curl 'http://localhost:9116/snmp?target=192.168.1.1'
```
Result:
```
ts=2024-01-11T11:33:33.332Z caller=collector.go:464 level=debug auth=public_v2 target=192.168.1.1 module=if_mib msg="Finished scrape" duration_seconds=5.002192868
```

If you don't mind, please update the `DefaultAuth` value in `config/config.go` to match the new default.

PierreF:
In main.go I see the default for the *auth* which default to "public_v2" which I'm fine with that.

In this PR I'm talking about the SNMP community send in the SNMP network query which currently had default to "public" but it's documented as "public_v2". And hopefully the default is "public" or I think it would break collection on most SNMP device which respond to "public" community.

So I'm not sure what you want:
* I don't want to change the default SNMP community from "public" to "public_v2" to match current documentation (due to risk breaking collection)
* I'm not sure changing the default auth from "public_v2" to "public" is a good idea, as it could cause breaking change on API

SuperQ:
Oh, right, I'm confusing which strings mean which things.

generator/README.md
```diff
@@ -86,7 +86,7 @@ auths:
     version: 2  # SNMP version to use. Defaults to 2.
                 # 1 will use GETNEXT, 2 and 3 use GETBULK.
 
-    # Community string is used with SNMP v1 and v2. Defaults to "public_v2".
+    # Community string is used with SNMP v1 and v2. Defaults to "public".
     community: public_v2
```
SuperQ, line 90: This example default is wrong as well. This should be just `public`.

PierreF: Fixed at the same time I've rebased this PR to fix conflict.

SuperQ (approved):
Thanks!

Issue #1080, opened 2023-12-26
servertech_sentry3 scraping broke between 0.20.0 and 0.21.0


robbat2:
Between 0.20.0 and 0.21.0, the servertech_sentry3 module stopped working, as is still broken as of v0.25.0.

### Host operating system: output of `uname -a`
`Linux (redacted) 6.1.38-gentoo-dist #1 SMP PREEMPT_DYNAMIC Thu Jul  6 11:45:23 -00 2023 x86_64 QEMU Virtual CPU version 2.5+ GenuineIntel GNU/Linux`

### snmp_exporter version: output of `snmp_exporter -version`

- broken: 0.25.0
- broken: 0.24.1
- broken: 0.21.0
- works: 0.20.0

### What device/snmpwalk OID are you using?
servertech_sentry3

### What did you do that produced an error?
`curl 'localhost:9116/snmp?target=10.0.1.18&auth=(redacted)&module=servertech_sentry3'`

### What did you expect to see?
Metrics

### What did you see instead?
```
# curl -sq 'localhost:9116/snmp?target=10.0.1.18&auth=(redacted)&module=servertech_sentry3' 
An error has occurred while serving metrics:

[...]

PR #1090, opened 2024-01-15
fix: workaround servertech_sentry3 zero-length integers

robbat2:
servertech_sentry3 devices can return bad ASN.1 data, an Integer with asnlength=0 under certain cases.

- Certain firmware versions only
- SNMP BULKGET
- Specific OIDs deep into the tree, at different repetition values

Work around the problem by setting max_repetitions=4, which doesn't trigger the device bug.

Closes: https://github.com/prometheus/snmp_exporter/issues/1080


SuperQ (approved):
Thanks, this seems like a reasonable workaround for now.

Issue #1093, opened 2024-01-15
curl: option --no-progress-meter: is unknown


weidongkl:
### Host operating system: output of `uname -a`
```
$ distro 
Name: Red Hat Enterprise Linux 8.8 (Ootpa)
Version: 8.8 (Ootpa)
Codename: Ootpa
$ uname -a
Linux localhost.localdomain 4.18.0-477.27.1.el8_8.x86_64 #1 SMP Thu Aug 31 10:29:22 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux
$ curl --version
curl 7.61.1 (x86_64-redhat-linux-gnu) libcurl/7.61.1 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.6 libidn2/2.2.0 libpsl/0.20.2 (+libidn2/2.2.0) libssh/0.9.6/openssl/zlib nghttp2/1.33.0
Release-Date: 2018-09-05
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz brotli TLS-SRP HTTP2 UnixSockets HTTPS-proxy PSL
```

### What did you do that produced an error?
```
$ cd generator
$ make generator mibs
make: 'generator' is up to date.
>> Downloading readydataos
curl: option --no-progress-meter: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
make: *** [Makefile:358: mibs/readydataos] Error 2

```

PR #1094, opened 2024-01-15
Fix curl failed

weidongkl:
Fixes: #1093 
```
curl version is 7.61.1  in rhel8.8,
and --no-progress-meter is added in 7.67.0.
so there is error curl: option --no-progress-meter: is unknown

For better compatibility, we shoule use the option -sS
```

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -14,7 +14,7 @@
 MIBDIR   := mibs
 MIB_PATH := 'mibs'
 
-CURL_OPTS ?= -L --no-progress-meter --retry 3 --retry-delay 3 --fail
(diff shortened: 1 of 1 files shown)


weidongkl:
curl man page says:
```
-s, --silent

         Silent or quiet mode. Don't show progress meter  or  error  mes‐
         sages.   Makes  Curl mute. It will still output the data you ask
         for, potentially even to the terminal/stdout unless you redirect
         it.

         Use  -S,  --show-error  in  addition  to  this option to disable
         progress meter but still show error messages.
```

SuperQ (approved):

Issue #1105, opened 2024-02-04
Override help string


SuperQ:
### Discussed in https://github.com/prometheus/snmp_exporter/discussions/1086

<div type='discussions-op-text'>

<sup>Originally posted by **llamafilm** January  3, 2024</sup>
It would be helpful to override the help text, for MIBs that are either lacking or inaccurate.

The help string is defined here: https://github.com/prometheus/snmp_exporter/blob/9c42d6c874d479314e612bca69558c81f8e26287/generator/tree.go#L353
I propose leaving that format as-is and allowing the user to override `description` in the generator.

This is especially useful in conjunction with the recent addition of scaling factor in #1026.  For example, this [Cyberpower UPS](https://github.com/prometheus/snmp_exporter/files/13823016/CPS-MIB.mib.txt) returns all sensor values as integers, so after scaling the voltage to the correct value, the help text is wrong.  This can be fixed in `snmp.yml` but I think we need a way to do it in the generator.

```yaml
  cyberpower_ups:
    walk:
      - 1.3.6.1.4.1.3808.1.1.1.3 # upsInput
    overrides:
      upsAdvanceInputLineVoltage:
        scale: 0.1
```

```
# HELP upsAdvanceInputLineVoltage The input line voltage in 1/10 VAC. - 1.3.6.1.4.1.3808.1.1.1.3.2.1
# TYPE upsAdvanceInputLineVoltage gauge
upsAdvanceInputLineVoltage 120.2
```
</div>

PR #1106, opened 2024-02-04
Add generator HELP override

SuperQ:
Allow override of the metric HELP text in the generator. Allows for altering generated MIB documentation.
* Update example help text for `st4TempSensorValue` to match scale override.

Fixes: https://github.com/prometheus/snmp_exporter/issues/1105


RichiH (approved):
LGTM

PR #1113, opened 2024-02-12
Skip using an interactive terminal in "make docker-generate".

jjdmol:
The "docker-generate" target does not actually seem to need user interaction. But requiring an interactive terminal in the "docker run" command does make it impossible/harder to embed it in an automated environment like CI/CD.

This patch simply removes the allocation of such a terminal. The "make docker-generate" seems to run just fine without it.


SuperQ (approved):
Thanks!

PR #1119, opened 2024-02-20
Add SNMPInflight metric

servak:
I have added an implementation to export the current number of requests.

```
> curl -s 'localhost:9116/metrics' | grep snmp_r
# HELP snmp_request_errors_total Errors in requests to the SNMP exporter
# TYPE snmp_request_errors_total counter
snmp_request_errors_total 0
# HELP snmp_request_in_flight Current number of requests being served.
# TYPE snmp_request_in_flight gauge
snmp_request_in_flight 0
```

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -275,6 +275,7 @@ type Metrics struct {
 	SNMPDuration           prometheus.Histogram
 	SNMPPackets            prometheus.Counter
 	SNMPRetries            prometheus.Counter
+	SNMPInflight           prometheus.Gauge
 }
 
 type NamedModule struct {
@@ -403,6 +404,8 @@ func (c Collector) collect(ch chan<- prometheus.Metric, logger log.Logger, clien
 
 // Collect implements Prometheus.Collector.
 func (c Collector) Collect(ch chan<- prometheus.Metric) {
+	c.metrics.SNMPInflight.Inc()
+	defer c.metrics.SNMPInflight.Dec()
 	wg := sync.WaitGroup{}
 	workerCount := c.concurrency
 	if workerCount < 1 {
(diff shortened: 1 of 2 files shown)


collector/collector.go
```diff
@@ -403,6 +404,8 @@ func (c Collector) collect(ch chan<- prometheus.Metric, logger log.Logger, clien
 
 // Collect implements Prometheus.Collector.
 func (c Collector) Collect(ch chan<- prometheus.Metric) {
+	c.metrics.SNMPInflight.Inc()
```
SuperQ, line 407: This doesn't account for multiple worker concurrency. I think this go in the `scaper` package. Or in the `collect()` internal function?

SuperQ (changes requested):

collector/collector.go
servak: It seems my understanding of "in-flight" might have been insufficient. 
I had assumed that the metric would measure the number of Collectors currently processing, but does it generally include the number of workers as well?

SuperQ: Yea, my understanding was more about the number of active SNMP requests. What do you think?

servak: OK! I will change it

SuperQ (approved):

main.go
```diff
@@ -267,6 +267,13 @@ func main() {
 				Help:      "Number of SNMP packet retries.",
 			},
 		),
+		SNMPInflight: promauto.NewGauge(
+			prometheus.GaugeOpts{
+				Namespace: namespace,
+				Name:      "request_in_flight",
+				Help:      "Current number of requests being served.",
```
SuperQ, line 274:
```suggestion
				Help:      "Current number of SNMP scrapes being requested.",
```

PR #1125, opened 2024-02-28
feat: adding dell support

buroa:
One caveat is that Dell URL does not like being downloaded without a user agent.


generator/Makefile
```diff
@@ -174,6 +177,14 @@ $(MIBDIR)/AIRESPACE-WIRELESS-MIB:
 	@echo ">> Downloading Cisco AIRESPACE-WIRELESS-MIB"
 	@curl $(CURL_OPTS) -o $(MIBDIR)/AIRESPACE-WIRELESS-MIB "$(CISCO_URL)/AIRESPACE-WIRELESS-MIB.my"
 
+$(MIBDIR)/.dell:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading dell to $(TMP)"
+	@curl $(CURL_OPTS) -H "User-Agent: snmp_exporter" -o $(TMP) $(DELL_URL)
+	@unzip -j -d $(MIBDIR) $(TMP)
```
SuperQ, line 184: I wonder if we should more selectively unzip this so we only extract the MIB files we really want.

buroa: I can do that. We really only need the iDRAC modules (I think).

buroa:
@SuperQ Ready for you :)

SuperQ (approved):

Issue #1132, opened 2024-03-12
Make generator error 22 when downloading PICO-IPSEC-FLOW-MONITOR-MIB.txt


suonlm:
### Host operating system: output of `uname -a`
Linux snmp-exporter 5.4.0-172-generic #190-Ubuntu SMP Fri Feb 2 23:24:22 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### What did you do that produced an error?
I run this command to build generator
` make generator mibs `
### What did you expect to see?
The build process must be completed
### What did you see instead?
>> Downloading PICO-IPSEC-FLOW-MONITOR-MIB.txt
curl: (22) The requested URL returned error: 403 
make: *** [Makefile:266: mibs/PICO-IPSEC-FLOW-MONITOR-MIB.txt] Error 22

SuperQ:
Looks like `jpn.nec.com` is blocking the default `curl` user agent string.

PR #1133, opened 2024-03-12
Update generator User-Agent

SuperQ:
Add a User-Agent flag variable to the generator Makefile.
* Use the User-Agent flag for NEC downloads.

Fixes: https://github.com/prometheus/snmp_exporter/issues/1132


jesusvazquez (approved):
LGTM

PR #1141, opened 2024-03-21
remove extra character from dell iDrac-SMIv1 MIB

fbegyn:
The dell idrac mib is currently failing on parsing with the following error. This is because after the `MIB-END` indicitor, there are some broken characters interrupting the parsing.

```
Bad operator (: At line 13878 in ./mibs//iDRAC-SMIv1.mib
```

This can easily be solved by just removing the last line of this MIB.


SuperQ (approved):
Thanks, I just noticed this with older versions of net-snmp.

PR #1142, opened 2024-03-21
Add Schneider_PM55XX MIB to downloads for SNMP generator

fbegyn:
Simple addition of the Schneider MIB for the PM55XX device series.


generator/Makefile
```diff
@@ -368,3 +370,12 @@ $(MIBDIR)/readynas:
...
 	@echo ">> Downloading readydataos"
 	@curl $(CURL_OPTS) -o $(MIBDIR)/readydataos "$(READYDATAOS_URL)"
+
+$(MIBDIR)/SchneiderPM55xx_V01_13.mib:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading SchneiderPM55xx_V01_13.mib to $(TMP)"
+	@curl $(CURL_OPTS) -o $(TMP) $(SCHNEIDER_PM55XX_URL)
+	@unzip -j -d $(MIBDIR) $(TMP) PM5560_PM5563_v2.1.0/SchneiderPM55xx_V01_13.mib
+	# Workaround invalid character in Schneider MIB
+	@sed -i 's/_//g' $(MIBDIR)/SchneiderPM55xx_V01_13.mib
```
bastischubert, line 380: Is this still necessary, given that we support mibopts (https://github.com/prometheus/snmp_exporter/pull/873) and you can simply allow those characters by adding mibopt "u" ?

fbegyn: Ah, totally slipped my mind that is an option. Switched it out in favor of your idea. The `--snmp.mibopts u` works perfectly.

SuperQ: Hmm, I'm not sure I'm happy with adding a MIB that requires custom mibopts. Perhaps we should adjust the generator default mibopts?

Or maybe get Schneider to fix their MIB. Cisco had some similar issues in the AIRESPACE-WIRELESS-MIB that I got them to fix. Teach them about `smilint` :grin:

SuperQ (changes requested):
If you want to add an additional MIB to the downloads, it would also be nice to add an example to the generator.yml. This way we get some testing that the MIB works.

Also, this needs a rebase.

fbegyn:
Rebased the code and tested with a few iterations locally. Still needs the `--snmp.mibopts u`. Any input on which way to go: have the hacky breakable `sed` in the Makefile or change the mibopts for the generator.

SuperQ:
The underscores issue seems to be somewhat common. I think we should add it to the flag default in `generator/main.go`.

What do you think @bastischubert?

bastischubert:
I also think we should have it in the generator as it takes too much effort for contributors do deal with unclean mibs.

PR #1148, opened 2024-03-30
Fix do not expand envvars for empty config fields

### LINKED ISSUE #1145: SNMPv2 Expansion of Environment Variables
### snmp_exporter version: output of `snmp_exporter -version`
current head version

**not relevant**
### If this is a new device, please link to the MIB(s).
**not relevant**

We are using a mix of SNMPv2 and SNMPv3 in the auth.yaml. With one device not configured with auth settings. 

```yaml
acistretch_v2:
  version: 2
  community: bla
```

This v2 module does not work as expected when using expansion of environment variables.
`ts=2024-03-27T13:28:01.053Z caller=main.go:203 level=error msg="Error parsing config file" err=" environment variable not found"`

The error seems to be here: https://github.com/prometheus/snmp_exporter/blob/733e5a1a31de2d2d74a3f48777db08a287c23f1d/config/config.go#L47-L65

Since the v2 configuration has no `username`, `password` or `privPassword` this expansion of env variables fails.

Introduced by https://github.com/prometheus/snmp_exporter/commit/59194f6ef22cf3804393421ce05948e31bfdbfb9

snmp-exporter does not fail parsing the env variables

### What did you see instead?

the above mentionend error message

IvoGoman:
Fixes an error loading the config, if any `Auth` section contains an empty `Username`, `Password` or `PrivPassword`
@SuperQ @RichiH

fixes #1145

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -47,20 +47,26 @@ func LoadFile(paths []string, expandEnvVars bool) (*Config, error) {
 	if expandEnvVars {
 		var err error
 		for i, auth := range cfg.Auths {
-			cfg.Auths[i].Username, err = substituteEnvVariables(auth.Username)
-			if err != nil {
-				return nil, err
+			if auth.Username != "" {
+				cfg.Auths[i].Username, err = substituteEnvVariables(auth.Username)
+				if err != nil {
+					return nil, err
+				}
 			}
-			password, err := substituteEnvVariables(string(auth.Password))
-			if err != nil {
-				return nil, err
+			if auth.Password != "" {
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
It would be good to add some test cases for this.

IvoGoman:
> It would be good to add some test cases for this.

Agreed, I missed the `config_test.go` in the root directory. 
Added a testcase for loading a SNMPv2 Auth Config without credentials specified. I believe the other cases are covered.

SuperQ (approved):

PR #1157, opened 2024-04-12
Add GoSNMP logger

SuperQ:
Add a debug logging to calls to GoSNMP.
* Requires `--log.level=debug`.
* Enable with `--snmp.debug-packets` or `snmp_debug_packets=true` URL param.


SuperQ:
CC @candlerb

SuperQ:
CC @bastischubert

PR #1162, opened 2024-04-22
Added 'oids-out-of-order' query parameter

jjdmol:
This PR adds a configuration option to have GoSNMP ignore the "OID not increasing" errors it encountes. This is done by setting `AppOpts["c"]` in GoSNMP (see https://github.com/gosnmp/gosnmp/blob/v1.37.0/gosnmp.go#L139) to `true`.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -290,6 +290,10 @@ func NewNamedModule(name string, module *config.Module) *NamedModule {
 	}
 }
 
+type ScraperOptions struct {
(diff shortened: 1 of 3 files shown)


SuperQ (changes requested):
This seems like it should be configured as a module flag, not a runtime query parameter.

jjdmol:
... where it already is an option as WalkParams.AllowNonIncreasingOIDs. I'm blind, sorry. Thanks for the pointer or I would have gotten further on that parallel path.

PR #845, opened 2023-02-14
Allow snmp_context to be passed as optional parameter to override the snmp.ContextName

dszarmach:
snmp?target=hostname&module=module will use existing c.Auth.ContextName from snmp.yml as g.ContextName if defined (same as previous behavior)

snmp?target=hostname&module=module&snmp_context=context_name allows for adding or overriding this value by passing it as a job param similar to target and module values.

this allows for simple automated configuration of jobs to scrape individual contexts without needing to define a new entry in snmp.yml for each

@SuperQ  @RichiH

PR #1163, opened 2024-04-23
Add a "snmp_context" parameter to the URL

### LINKED ISSUE #1040: Get snmp context from param
SNMP Context exists with snmpv3 and is often used to group various management informations together. (details can be found here: https://datatracker.ietf.org/doc/html/rfc5343#section-2)  

Currently we're able to use snmp contexts as part of the `auths` definition in snmp.yml which makes it quite hard to use them, especially whe you use one snmp context per vrf (for example). As the context is not security sensitive, we should allow to pass in the context via a url param
for example: `/snmp?target=10.11.12.13&auth=mystrongauth&module=snmp_inventory&snmp_context=vrf-mgmt`

ptodev:
This PR adds a `snmp_context` parameter to the URL, e.g.:
```
/snmp?target=10.11.12.13&auth=mystrongauth&module=snmp_inventory&snmp_context=vrf-mgmt
```

Fixes #1040

This is essentially the same PR as #845.

I have not tested this yet. What do you think would be a good way to test? For now I'll try to set something up with [snmpsim](https://github.com/etingof/snmpsim). Also, I'm not sure if I need to add any unit tests?

@SuperQ @RichiH

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,10 @@ is not necessary within the Prometheus configuration file.
 Metrics concerning the operation of the exporter itself are available at the
 endpoint <http://localhost:9116/metrics>.
 
+It is possible to supply an optional `snmp_context` parameter in the URL, like this:
+<http://localhost:9116/snmp?auth=my_secure_v3&module=ddwrt&target=192.0.0.8&snmp_context=vrf-mgmt>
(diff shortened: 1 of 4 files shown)


bastischubert:
tested the change with my arista, does whats expected and does not break scraping of snmpv1/v2

SuperQ (approved):
LGTM

PR #1167, opened 2024-05-01
Improve generator parse error handling

SuperQ:
Improve the handling of parse errors from net-snmp.
* Fix the line split when there are no errors.
* Default to failing on parse errors.
* Add extra help messages to help when there are errors.
* Improve `parse_errors` output.

### DIFF
--- a/generator/main.go
+++ b/generator/main.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/alecthomas/kingpin/v2"
@@ -27,6 +28,10 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/prometheus/snmp_exporter/config"
+)
+
+var (
+	cannotFindModuleRE = regexp.MustCompile(`Cannot find module \((.+)\): (.+)`)
 )
 
 // Generate a snmp_exporter config and write it out.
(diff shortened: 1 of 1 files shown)


SuperQ:
So in previous combinations of the generator and net-snmp we would get missing MIB file information in the STDERR output of net-snmp. This would be easy to parse and print back to the user. But in the version I'm testing here the information is missing and I'm not sure why.

SuperQ:
CC @bastischubert

bastischubert:
> But in the version I'm testing here the information is missing and I'm not sure why.
Can you share which versions you're using ;)

SuperQ:
It looks like the difference is between net-snmp 5.7 and 5.9. Previously we would show errors like this:
```
Cannot find module (SNMPv2-TC): At line 8 in mibs/HOST-RESOURCES-MIB
```

But that is not included anymore. If I use `snmpwalk -m`, it produces the error. So there is something about the way we call the C code that makes the errors not show up in newer net-snmp versions.

SuperQ:
For now, I think we should go ahead with this change. It's still useful as-is. We can continue to work on the generator error parsing later.

bastischubert:
adding `--snmp.mibopts=e` produces the desired behaviour with net-snmp >= 5.9.x

bastischubert:
We should add that param to the Makefile
```
diff --git a/generator/Makefile b/generator/Makefile
index 579610d..bb7530a 100644
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -70,7 +70,7 @@ generate: generator mibs
        MIBDIRS=$(MIB_PATH) ./generator --fail-on-parse-errors generate

 parse_errors: generator mibs
-       MIBDIRS=$(MIB_PATH) ./generator --fail-on-parse-errors parse_errors
+       MIBDIRS=$(MIB_PATH) ./generator --fail-on-parse-errors parse_errors --snmp.mibopts=e

 .PHONY: docker
 docker:```

**Comment by SuperQ** [comment]
Easier, I added `e` to the default options.

**Comment by SuperQ** [comment]
The error output is now much easier to read:

```
$ MIBDIRS=mibs ./generator generate
ts=2024-05-05T19:09:48.112Z caller=net_snmp.go:175 level=info msg="Loading MIBs" from=mibs
ts=2024-05-05T19:09:48.258Z caller=main.go:177 level=warn msg="NetSNMP reported parse error(s)" errors=2453
ts=2024-05-05T19:09:48.258Z caller=main.go:183 level=error msg="Missing MIB" mib=AIRESPACE-REF-MIB from="At line 66 in mibs/AIRESPACE-WIRELESS-MIB"
ts=2024-05-05T19:09:48.337Z caller=main.go:135 level=error msg="Failing on reported parse error(s)" help="Use 'generator parse_errors' command to see errors, --no-fail-on-parse-errors to ignore"
```

RichiH:
LGTM

PR #1183, opened 2024-05-10
Fix Dell iDrac mib sed for BSD/MacOS

SuperQ:
The previous implementation appearently breaks on the Mac `sed` implementation I noticed.
This still applies the fix, but uses an intermediate temporary file that atomically moved
in place after the operation.


SuperQ:
:eyeglasses: @bastischubert

bastischubert:
The iDRAC-SMIv1.mib is zero bytes after the sed (tested on macos/BSD and linux/GNU) 👎

SuperQ:
Fixed sed syntax.

SuperQ:
Closing in favor of https://github.com/prometheus/snmp_exporter/pull/1185

PR #1185, opened 2024-05-11
Remove old iDRAC mib

SuperQ:
It turns out we don't need iDRAC-SMIv1, so don't extract the broken MIB from the Dell bundle.


SuperQ:
:eyeglasses: @bastischubert

bastischubert:
LGTM

PR #1200, opened 2024-06-12
Deduplicate UCD-SNMP-MIB OIDs

SuperQ:
Consolidate the use of UCD-SNMP-MIB into separate modules, rather than include them in individual device modules.


SuperQ:
CC @bastischubert Saves about 400 lines of snmp.yml.

SuperQ:
I have been adding these changes as notes to the release docs.

PR #1217, opened 2024-07-17
Add schleifenbauer PDU

Wqrld:
This patch adds support for Power Distribution Units from https://www.schleifenbauer.eu

I was able to successfully read out my PDU's into grafana. I hope this may be useful for others too.

Note:
This query takes about one minute to run on my device, running into timeouts. Adding
```
    scrape_interval: 3m
    scrape_timeout: 2m
```
fixes this. I think this is an issue with many SNMP devices, so it might be an idea to add statements like these (with lower limits) to the readme example.


bastischubert:
Hi @Wqrld,

could you please provide the updated generator/generator.yml and add the needed mib files for this to work to the Makefile?

Wqrld:
The generator was also failing locally before adding my patch because of 404's, i had to rip those out of the makefile to generate this configuration.

Wqrld:
> Hi @Wqrld,
> 
> could you please provide the updated generator/generator.yml and add the needed mib files for this to work to the Makefile?

I do not have the rights to those original mib files so i'm not sure if i'm allowed to add them. They were taken from https://docs.schleifenbauer.eu/?dir=Interfaces/SNMP
If this is a hard requirement i'll send them an email for explicit permission

bastischubert:
Shouldn't be an issue to curl their public available mib files in the generator/Makefile

generator/Makefile
```diff
@@ -367,3 +371,10 @@ $(MIBDIR)/readynas:
 $(MIBDIR)/readydataos:
 	@echo ">> Downloading readydataos"
 	@curl $(CURL_OPTS) -o $(MIBDIR)/readydataos "$(READYDATAOS_URL)"
+
+$(MIBDIR)/SCHLEIFENBAUER-PRODUCTS-MIB.txt $(MIBDIR)/SCHLEIFENBAUER-DATABUS-MIB.txt $(MIBDIR)/SCHLEIFENBAUER-SMI.txt:
+	$(eval TMP := $(shell mktemp))
+	@echo ">> Downloading schleifenbauer to $(TMP)"
+	@curl $(CURL_OPTS) -o $(TMP) $(SCHLEIFENBAUER_URL)
+	@unzip -j -d $(MIBDIR) $(TMP) SCHLEIFENBAUER-PRODUCTS-MIB.txt SCHLEIFENBAUER-DATABUS-MIB.txt SCHLEIFENBAUER-SMI.txt
```
bastischubert, line 379: The SCHLEIFENBAUER-PRODUCTS-MIB.txt seems not to be included in this zipfile, only in the older https://docs.schleifenbauer.eu/Interfaces/SNMP/PDU_mib_244.zip. If the file is necessary we'd need to get it from the 2nd zipfile.

generator/generator.yml
```diff
@@ -628,3 +628,11 @@ modules:
...
     walk:
       - hrSWInstalled
+
+#
+# Schleifenbauer
+#
+# https://docs.schleifenbauer.eu/?dir=Interfaces/SNMP
+  schleifenbauer:
+    walk:
+      - 1.3.6.1.4.1.31034.12
```
SuperQ, line 638: This OID is overly broad. I think we want to be more targeted here with which walks to include. Some of the metrics included by this are read-write settings OIDs that have no meaning as metrics.

Let's start with the most useful part, the device table.

```suggestion
      - sdbMgmtCtrlDevicesTable
```

SuperQ (changes requested):

generator/generator.yml
Wqrld: I think almost everything in 1.3.6.1.4.1.31034.12.1.1.2 would be of use, which is basically the whole generated list. There are a few you could leave out, but for my usecase more is better.

PR #1231, opened 2024-08-07
Update generator default MIBOPTS

SuperQ:
Enable `u`, which is "allow the use of underlines in MIB symbols". This enables better compatibility with some MIBs.


SuperQ:
:eyeglasses: @fbegyn

bastischubert:
Looks good - 👍

bastischubert (approved):

PR #1247, opened 2024-09-21
Set UseUnconnectedUDPSocket option if one of the modules has if set

unarep:
Signed-off-by: Urmas Närep <urmas.narep@gmail.com>


bastischubert (approved):
LGTM

PR #1249, opened 2024-09-29
chore!: adopt log/slog, drop go-kit/log

tjhop:
This PR includes:
- Go version updates, so that `log/slog` can be used
- linter updates to remove configs for `go-kit/log` that are no longer needed, and enable `sloglint` linter
- Go dep updates for prometheus/{client_golang,common,exporter-toolkit} libs

The bulk of this PR was automated by the following script which is being used to aid in converting the various exporters/projects to use slog:

https://gist.github.com/tjhop/49f96fb7ebbe55b12deee0b0312d8434

Builds and passes tests.


tjhop:
cc: @SuperQ

collector/collector.go
```diff
@@ -365,7 +364,7 @@ func (c Collector) collect(ch chan<- prometheus.Metric, logger log.Logger, clien
 	results, err := ScrapeTarget(client, c.target, c.auth, module.Module, logger, c.metrics)
 	c.metrics.SNMPInflight.Dec()
 	if err != nil {
-		level.Info(logger).Log("msg", "Error scraping target", "err", err)
+		logger.Info("Error scraping target", "err", err)
```
tjhop, line 367: There are several instances like this that are logging errors at levels other than `error` -- is this intentional?

SuperQ: I will need to check the history, but these probably should be `Debug()` or `Warn()`.

SuperQ: I went through the history of the Info logging in `collector.go` and it seems like these were recent logging level mistakes.

For now let's preserve the existing logging levels and we can revisit them later in a separate PR.

SuperQ:
I've merged the prometheus/common update.

SuperQ (approved):

PR #1259, opened 2024-10-06
Support for passing community as an env var

cloudtechninja:
This follows the work of @harshavmb in #1074 and adds support for using environment variable substitution for `community` strings. Also includes a small fix for the generator makefile so that tests are passing.

@SuperQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ using SNMP v2 GETBULK.
 The `--config.file` parameter can be used multiple times to load more than one file.
 It also supports [glob filename matching](https://pkg.go.dev/path/filepath#Glob), e.g. `snmp*.yml`.
 
-The `--config.expand-environment-variables` parameter allows passing environment variables into some fields of the configuration file. The `username`, `password` & `priv_password` fields in the auths section are supported. Defaults to disabled.
+The `--config.expand-environment-variables` parameter allows passing environment variables into some fields of the configuration file. The `username`, `password`, `priv_password`, & `community` fields in the auths section are supported. Defaults to disabled.
 
 Duplicate `module` or `auth` entries are treated as invalid and can not be loaded.
 
@@ -162,7 +162,7 @@ scrape_configs:
       - targets: ['localhost:9116']
 ```
 
-You could pass `username`, `password` & `priv_password` via environment variables of your choice in below format. 
+You could pass `username`, `password`, `priv_password` & `community` via environment variables of your choice in below format.
 If the variables exist in the environment, they are resolved on the fly otherwise the string in the config file is passed as-is.
 
 This requires the `--config.expand-environment-variables` flag be set.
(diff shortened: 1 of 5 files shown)


generator/Makefile
```diff
@@ -300,7 +300,7 @@ $(MIBDIR)/.sophos_xg:
 	$(eval TMP := $(shell mktemp))
 	@echo ">> Downloading Sophos XG to $(TMP)"
 	@curl $(CURL_OPTS) -o $(TMP) $(SOPHOS_XG_URL)
-	@unzip -j -d $(MIBDIR) $(TMP) SOPHOS-XG-MIB20.txt
+	@unzip -j -d $(MIBDIR) $(TMP) sophos-xg-mib/SOPHOS-XG-MIB20.txt
```
bastischubert, line 303: this has already been fixed in #1266 - please remove this commit from the PR

bastischubert (changes requested):
Hey Eric, thanks for this PR. could you just check the two comments i made?

README.md
```diff
@@ -162,15 +162,15 @@ scrape_configs:
...
+You could pass `username`, `password`, `priv_password` & `community` via environment variables of your choice in below format.
 If the variables exist in the environment, they are resolved on the fly otherwise the string in the config file is passed as-is.
 
 This requires the `--config.expand-environment-variables` flag be set.
 
 ```YAML
 auths:
   example_with_envs:
-    community: mysecret
+    community: ${ARISTA_COMMUNITY}
```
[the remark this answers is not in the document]
cloudtechninja: I updated the examples so its more clear

generator/Makefile
cloudtechninja: Removed

SuperQ (changes requested):
It was an explicit choice in https://github.com/prometheus/snmp_exporter/pull/1074 to not support community strings, as they are not secrets.

We have no plans to support this.

cloudtechninja:
I think in practice they are treated as sensitive information. Functionally they are equivalent to v3 credentials and provide the same level of access, albeit without transport layer encryption.  If the position is that they are insecure and should be avoided I can add a note to this affect.  I will also update the testdata/*.yml files to use "notsecret" instead of "mysecret".

RFC 1157 implies that certain communities are sensitive, of course especially those with write permissions. There are also views/ACLs that restrict which OID trees a read only v2c or v3 community can access - implying potentially sensitive data.

From RFC1157:
```
+-----------------+
|Domain=Region#3  |
|CPU=router-1     |
|DCommunity=secret|
+-----------------+
```

All this aside, the intent for this PR is to better manage the auth.yml configuration to allow substitution using k8s environment configmaps such that auth.yml itself can be managed via GitOps.

@bastischubert Do you have any thoughts on this?

SuperQ:
Yes, updating the docs and testdata would be very helpful.

PR #1275, opened 2024-10-30
Add powercom MIBs

teran:
Powercom MIBs, tested against NetAgent DA807


teran:
Powecom MIBs are added and tested against DA807 card so looks working (i.e metrics are present and even named correctly during scrape), however It's my first dance with SNMP and adding MIBs to snmp exporter so please check it carefully before merging :)

Thanks!

teran:
Based on CI run there's amount of lgpPdu* metrics generated (probably from other MIBs?), could someone please advise how to ignore them and leave only upsSmart* ?

generator/generator.yml
```diff
@@ -752,3 +752,17 @@ modules:
...
+# https://www.upspowercom.com/Card-DA807.jsp
+#
+  powercom:
+    walk:
+      - 1.3.6.1.4.1.935.1.1.1.1
+      - 1.3.6.1.4.1.935.1.1.1.2
+      - 1.3.6.1.4.1.935.1.1.1.3
+      - 1.3.6.1.4.1.935.1.1.1.4
+      - 1.3.6.1.4.1.935.1.1.1.9
+      - 1.3.6.1.4.1.476.1.42.3
```
SuperQ, line 768: This is unrelated to the specified MIB. It's part of `LIEBERT-GP-ENVIRONMENTAL-MIB`.

```suggestion
      - 1.3.6.1.4.1.476.1.42.3
```
```suggestion
```

SuperQ (changes requested):
You also need to generate the snmp.yml and update it for CI to pass.

generator/generator.yml
teran: Based on OID table at  https://www.upspowercom.com/Card-DA807.jsp this OID is exposed by the device, so does it mean if it's present in any other MIB I don't need to place it for my set of MIBs ?

teran:
Now it should be fine, CI is passed, all relevant metrics are in place and I've checked it against real DA807 - metrics are also there.

generator/generator.yml
SuperQ: Yes, devices can output data for multiple MIBs. We prefer to make configuration blocks here that target subsections of data and then you can scrape devices with multiple targeted modules.

generator/generator.yml
```diff
@@ -771,3 +771,21 @@ modules:
...
+      - 1.3.6.1.4.1.935.1.1.1.1
+      - 1.3.6.1.4.1.935.1.1.1.2
+      - 1.3.6.1.4.1.935.1.1.1.3
+      - 1.3.6.1.4.1.935.1.1.1.4
+      - 1.3.6.1.4.1.935.1.1.1.5
+      - 1.3.6.1.4.1.935.1.1.1.6
+      - 1.3.6.1.4.1.935.1.1.1.7
+      - 1.3.6.1.4.1.935.1.1.1.8
+      - 1.3.6.1.4.1.935.1.1.1.9
+      - 1.3.6.1.4.1.935.2.1.2.1
```
SuperQ, line 791: Would you please document these walk OIDs with what identifier they belong to?

See the other examples in the generator file.

For example:
```suggestion
      - 1.3.6.1.4.1.935.1.1.1.1 # upsIdentp
      - 1.3.6.1.4.1.935.1.1.1.2
      - 1.3.6.1.4.1.935.1.1.1.3
      - 1.3.6.1.4.1.935.1.1.1.4
      - 1.3.6.1.4.1.935.1.1.1.5
      - 1.3.6.1.4.1.935.1.1.1.6
      - 1.3.6.1.4.1.935.1.1.1.7
      - 1.3.6.1.4.1.935.1.1.1.8
      - 1.3.6.1.4.1.935.1.1.1.9
      - 1.3.6.1.4.1.935.2.1.2.1
```

SuperQ (changes requested):
It seems like you're just including everything possible. This isn't usually a good idea as there are lots of walks that don't provide value. For example you have included `mconfigTrapsReceiversTable` which doesn't seem to be very useful.

teran:
@SuperQ I've updated the list of OIDs to contain the only ones meaning and not empty, however it's probably not the whole list and it's up to update while users will use the module since this card (DA807) will probably fit into amount of UPS' devices. Descriptions are also added.

CI is failed at the moment because upspowercom.com is down at the moment so it's up to retrigger the build.

Thank you for your guidance :-)

SuperQ (approved):
Thanks!

PR #1303, opened 2024-11-28
apc mib update

bastischubert:
update apc mib download (currently broken) and add instructions in case download fails


RichiH (approved):
LGTM

PR #1305, opened 2024-11-28
chore: fix yamllint errors during make of snmp_exporter due to missing newlines

bastischubert:
fix yamllint errors during make of snmp_exporter due to missing newlines


RichiH (approved):
LGTM

PR #1315, opened 2024-12-13
Add NTPTimeStamp type

phibos:
Parse the ntp timestamp and convert it to an unix timestamp.

Used in different MIBs

* Cisco: https://github.com/cisco/cisco-mibs/blob/26b38ac1b25a38510b6bb308465dd068e79df64b/v2/CISCO-NTP-MIB.my#L196
* Juniper: https://github.com/pgmillon/observium/blob/604a21c8abceeea9b24a264220082c4ca9b15dd8/mibs/junose/juniSystemClock.mi2#L133
* DELL: https://github.com/pgmillon/observium/blob/604a21c8abceeea9b24a264220082c4ca9b15dd8/mibs/dell/RADLAN-TIMESYNCHRONIZATION-MIB#L34
* https://github.com/pgmillon/observium/blob/604a21c8abceeea9b24a264220082c4ca9b15dd8/mibs/aos/TIMETRA-NTP-MIB#L98

At the moment I have tested it with some Cisco devices.

I have tried to add some tests, but I don't know if something is missing.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -562,6 +562,20 @@ func parseDateAndTimeWithPattern(metric *config.Metric, pdu *gosnmp.SnmpPDU, met
 	return float64(t.Unix()), nil
 }
 
+func parseNtpTimestamp(pdu *gosnmp.SnmpPDU) (float64, error) {
+	var data []byte = pdu.Value.([]byte)
+
+	// Prometheus uses the Unix time epoch (seconds since 1970).
+	// NTP seconds are counted since 1900 and must be corrected
+	// by removing 70 yrs of seconds (1970-1900) or 2208988800
+	// seconds.
+	secs := int64(binary.BigEndian.Uint32(data[:4])) - 2208988800
+	nanos := (int64(binary.BigEndian.Uint32(data[4:])) * 1e9) >> 32
+
+	t := time.Unix(secs, nanos)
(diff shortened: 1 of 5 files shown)


collector/collector.go
```diff
@@ -562,6 +562,17 @@ func parseDateAndTimeWithPattern(metric *config.Metric, pdu *gosnmp.SnmpPDU, met
 	return float64(t.Unix()), nil
 }
 
+func parseNtpTimestamp(pdu *gosnmp.SnmpPDU) (float64, error) {
+	var data []byte = pdu.Value.([]byte)
+
+	// 2208988800 is the Unix Epoch in NTP Seconds (1970 - 1900)
```
bastischubert, line 568: we should include a hint that the returned data is "Seconds and Fractions since 01.01.1900" as the ntp era starts on jan 1st 1900 whereas unix epoch starts 70y later (so the 2208988800 is basically the offset between those two eras)

phibos: I have added a comment with additional information, rebased against current main branch and pushed the changes.

Feel free to have a look.

bastischubert: 🙇

bastischubert (approved):
LGTM

SuperQ:
Nice!

Issue #648, opened 2021-04-21
Usage of WebUI behind reverse proxy (subtree / baseurl)


falkheiland:
I am using snmp_exporter in a docker container behind traefik. where, in prometheus and alertmanager, i can specifiy
```
--web.external-url=https://<baseurl>/<subtree>
```
this parameter does not exist in snmp_exporter. 
In the compose file i can have traefik labels like:

```
- "traefik.http.routers.snmp-exporter-secure.rule=Host(`<baseurl>`) && PathPrefix(`/<subtree>/`)"
- "traefik.http.middlewares.snmp-exporter-secure.chain.middlewares=secHeaders@file,default-auth@file,strip-prefix"
- "traefik.http.middlewares.strip-prefix.stripprefix.prefixes=/snmp-exporter"
```

which allow me to access /config + /snmp, but in the WebUI the links are still the same.

i would like to have the `web.external-url` parameter in place or another viable soulution for that scenario.

SuperQ:
This seems reasonable to me.

PR #1328, opened 2025-01-02
Add args to serve application behind a route prefix/external URL

kwilt:
Add feature to serve application behind a route prefix or external URL by supplying args --web.external-url/--web.route-prefix, in the same way as blackbox_exporter or other exporters

I essentially copied the code from blackbox_exporter to achieve this functionality. Currently, the [exporter-toolkit's web landing page](https://github.com/prometheus/exporter-toolkit/blob/master/web/landing_page.go) doesn't allow for serving the landing page on anything other than "/" (returns 404) so I had to replace the landing page section of code in this exporter with some HTML like blackbox_exporter is doing. 

_With that in mind_, I guess you could consider this a proof-of-concept pull request as it does change the existing implementation of the landing page - and because of this I don't expect this pull request to get merged as-is - however, it appears all functionality is retained.
 
I would consider these additional args to be "must-have" features to bring snmp_exporter closer to feature parity with the other exporters. I needed to serve this exporter behind a prefix like I'm doing with the other exporters, so I figured I would make an attempt at this update myself, as it does appear to be something other people have been wanting for quite some time. See: #648

fyi: this is my first pull request on GitHub, so hopefully I haven't done anything too awful here 😅


SuperQ:
I think we should make improvements to the exporter-toolkit to better support this first. This way we don't ave to rewrite existing functionality.

PR #1333, opened 2025-01-07
feat: allow module-qualified labels in generator.yml

astralblue:
Currently, some labels such as `ifName` and `ifDescr` are defined in multiple modules and must be specified using OID.  This patch avails such OIDs under both its bare label and module-qualified labels, e.g. both `ifName` and `IF-MIB::ifName`.

@bastischubert could you please take a look?

### DIFF
--- a/generator/generator.yml
+++ b/generator/generator.yml
@@ -13,11 +13,11 @@ modules:
       - source_indexes: [ifIndex]
         lookup: ifAlias
       - source_indexes: [ifIndex]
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
I love it.

Would this also work for `walk`?

```yaml
modules:
  if_mib:
    walk:
      - IF-MIB::interfaces
      - ...
```

bastischubert (approved):

PR #1341, opened 2025-01-15
Add Feature to Override Metric Name in Generator

marcovrv:
### Description  
This PR introduces a feature to the `snmp_exporter` generator, enabling users to override metric names directly in the `generator.yml` file. The new `name` parameter in the override section allows for custom metric names to be output in the `snmp.yml` file, streamlining monitoring of devices with similar metrics but differing MIB OID names.

#### Key Changes:  
1. **Code Changes**:  
   - **`generator/config.go`**: Added support for the optional `name` parameter in the override section.  
   - **`generator/tree.go`**: Ensures the `name` parameter, if specified and non-empty, replaces the default MIB-based metric name.  

2. **Documentation**:  
   - **`generator/README.md`**: Updated to describe the `name` parameter and its usage in the `generator.yml` override section.  

### Example Usage  
The following example demonstrates how to configure custom metric names for two types of power supplies in `generator.yml`:

```yml
# Emerson/Vertiv netsure Power Supply
  vertiv_netsure:
    walk:
    - psTemperature1
    - psInputLineAVoltage
    - systemVoltage
    - systemCurrent
    - psTotalBatteryCurrent
    overrides:
      psTemperature1:
        scale: 0.001
        name: pfuTemperature
        help: "System Temperature in Celsius"
      psInputLineAVoltage:
        scale: 0.001
        name: pfuInputVoltage
        help: "AC Input Voltage in Volts"
      systemVoltage:
        scale: 0.001
        name: pfuOutputVoltage
        help: "DC Output Voltage in Volts"
      systemCurrent:
        scale: 0.001
        name: pfuOutputCurrent
        help: "DC Output Current in Amperes"
      psTotalBatteryCurrent:
        scale: 0.001
        name: pfuBatteryCurrent
        help: "Battery Current in Amperes"

# Compas (Alpha Innovations) Power Supply
  compas:
    walk:
    - es1DcSys1DataBatMeasurementsTemp
    - es1DcSys1DataACInputsGlobalAverageVoltage
    - es1DcSys1DataConvertersOutVoltage
    - es1DcSys1DataConvertersOutCurrent
    - es1DcSys1DataBatMeasurementsInputCurrent
    overrides:
      es1DcSys1DataBatMeasurementsTemp:
        name: pfuTemperature
        help: "System Temperature in Celsius"
      es1DcSys1DataACInputsGlobalAverageVoltage:
        name: pfuInputVoltage
        help: "AC Input Voltage in Volts"
      es1DcSys1DataConvertersOutVoltage:
        name: pfuOutputVoltage
        help: "DC Output Voltage in Volts"
      es1DcSys1DataConvertersOutCurrent:
        name: pfuOutputCurrent
        help: "DC Output Current in Amperes"
      es1DcSys1DataBatMeasurementsInputCurrent:
        name: pfuBatteryCurrent
        help: "Battery Current in Amperes"
```

### Justification  
This feature was implemented to address the need to monitor -48 VDC power systems from multiple vendors. Although these systems share the same types of metrics (e.g., input/output electrical values, battery status), their MIB-based names differ. With this feature, metric names can be normalized at the generator step, eliminating the need for relabeling or recording rules in the main prometheus.yml.  

### Testing  
- Verified that the `name` parameter correctly overrides the default metric name when specified.  
- Tested with real-world configurations for multiple devices to ensure consistent and expected behavior.  

### Checklist  
- [x] The feature is backward-compatible and does not affect existing configurations.  
- [x] Documentation has been updated to reflect the new functionality.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -159,6 +159,7 @@ modules:
       metricName:
         ignore: true # Drops the metric from the output.
         help: "string" # Override the generated HELP text provided by the MIB Description.
+        name: "string" # Override the OID name provided in the MIB Description.
         regex_extracts:
           Temp: # A new metric will be created appending this to the metricName to become metricNameTemp.
             - regex: '(.*)' # Regex to extract a value from the returned SNMP walks's value.
--- a/generator/config.go
+++ b/generator/config.go
@@ -35,6 +35,7 @@ type MetricOverrides struct {
 	Scale           float64                           `yaml:"scale,omitempty"`
 	Type            string                            `yaml:"type,omitempty"`
 	Help            string                            `yaml:"help,omitempty"`
+	Name            string                            `yaml:"name,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -542,6 +542,9 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 				if params.Help != "" {
 					metric.Help = params.Help
 				}
+				if params.Name != "" {
+					metric.Name = params.Name
+				}
 			}
 		}
 	}


RichiH:
I'm somewhat torn as this can be a footgun, yet I also see the value in having the option.

I do wonder two things
1. Should there be an option to add labels in this stage as well? E.g. pfuInputVoltage{line="A"} in your example above.
2. Should there be a gauge for total renamed or so, per target? I kind of worry about renaming happening without the operator being able to verify

An ~obvious follow-up consideration is if there should be renaming in the default config. Likely no, unless there are really obvious mistakes?

marcovrv:
> 1. Should there be an option to add labels in this stage as well? E.g. pfuInputVoltage{line="A"} in your example above.

I don't believe it's necessary. In my experience, when there are multiple values of the same kind, vendors usually arrange them in tables with indexes, which the generator already understands.

> 2. Should there be a gauge for total renamed or so, per target? I kind of worry about renaming happening without the operator being able to verify

Do you mean in the logs or the exporter self-metrics? I haven’t explored that aspect in detail yet; I just replicated the same processing as the help text override. The default behavior should be to preserve the metric names as defined in the MIB, ensuring the operator sees the expected values unless there is a specific and explicit need to override them.
 
> An ~obvious follow-up consideration is if there should be renaming in the default config. Likely no, unless there are really obvious mistakes?

I agree; the default name should remain the one in the MIB (as it is now). However, this feature could be helpful for resolving duplicated names, should that occur in the default config.

The name override functionality is particularly valuable for us, as we manage DC systems from four vendors, each with its own interpretation of the enterprises subtree. I see this option as a practical tool for standardizing metric names across vendors.

In hindsight, it might be wise to restrict the name override functionality to the enterprises subtree. This approach would help maintain consistency and minimize the risk of unintended modifications in other parts of the MIB.

marcovrv:
> This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

Thanks. I followed the steps for this, but it seems is still on error., although in my git log it shows as signed off:

`commit 544c4f4934ec6d76b1a756c90e48286c0f399014 (HEAD -> feat/override-metric-name)
Author: PrometheusBot <prometheus-team@googlegroups.com>
Date:   Thu Jan 16 10:01:14 2025 +0100

    Update common Prometheus files (#1340)
    
    Signed-off-by: prombot <prometheus-team@googlegroups.com>
    Signed-off-by: Marco Rodríguez <marcovrv@gmail.com>

commit f07d54aa0829fbb9d9d14e9e789c5e1715e43670
Author: Marco Rodríguez <marcovrv@gmail.com>
Date:   Wed Jan 15 15:26:23 2025 -0600

    generator - added metric name override feature
    
    Signed-off-by: Marco Rodríguez <marcovrv@gmail.com>
`

SuperQ:
Yes, it looks like you pulled in a commit that is not your own. That is what is causing the error.

marcovrv:
> I've seen at least a few MIBs where the vendor didn't seem to understand tables and made a series of separate OIDs, so I think adding a label feature is actually an interesting idea. I'm not sure how name collisions would interact with the exporter at runtime tho.

I was referring to name collisions (in fact, name duplication) during the generator phase for the default generator.yml (across MIBs). However, I now recall that the recommended solution is to use module::name or OIDs instead of names, so it seems there's no need to apply name overrides for default MIBs.

As for the snmp_exporter itself, there don't seem to be any side effects caused by name collisions. All it does is expose the new metric names for the scraped OIDs.

SuperQ (approved):
I think this is a good first step. We can improve on things like labels later.

bastischubert (approved):
A usefull addition - lets go 👍

RichiH:
> Do you mean in the logs or the exporter self-metrics

I mean a `snmp_exporter_overriden_metrics` or so which the exporter would expose as part of the target's metrics if the amount of overriden metrics is non-zero.

That can be done later though, I'm fine merging early and improving later

PR #1370, opened 2025-02-13
Update generator.yml

manases13:


bastischubert:
@manases13 this PR uses oids that are already present in `system` and `if_mib` module. Also it's missing a proper description of the problem that is solved by this PR, the MIB file download and the changed `snmp.yml`

bastischubert (changes requested):
also the PR Misses a description - please either fix the already mentioned issues with the PR and also please add a description what problem this PR is solving. thx!

bastischubert:
closing PR due to no reaction.

PR #1377, opened 2025-02-21
use the existing mibs folder as default

veso0308:
To simplify the usage of the generator, use the containing mibs folder as default for "userMibsDir"


SuperQ (changes requested):
Sorry, this default is intentional in order to use the net-snmp default path.

PR #1411, opened 2025-04-16
chore(generator): fail early for bad walk OID

JohnVillalovos:
In `generator.yml` and in a `walk:` section.  If an OID starts with a
period, for example `.1.3.6.1.4.1.20916`, this will fail to be found,
as it should be: `1.3.6.1.4.1.20916`. But the error message lacks
detail when this happens.

Fail early if an OID starts with a "." and explain why.

### DIFF
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -302,6 +302,9 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 	// Remove redundant OIDs to be walked.
 	toWalk := []string{}
 	for _, oid := range cfg.Walk {
+		if strings.HasPrefix(oid, ".") {
+			return nil, fmt.Errorf("invalid OID %s, prefix of '.' should be removed", oid)
+		}
 		// Resolve name to OID if possible.
 		n, ok := nameToNode[oid]
 		if ok {


generator/tree.go
```diff
@@ -302,6 +302,9 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 	// Remove redundant OIDs to be walked.
 	toWalk := []string{}
 	for _, oid := range cfg.Walk {
+		if strings.HasPrefix(oid, ".") {
+			logger.Error("OID starts with a '.'", "OID", oid)
```
SuperQ, line 306: We prefer "fail fast" in order to highlight invalid configurations. Let's just error out the whole thing.

```suggestion
			return nil, fmt.Errorf("invalid OID %s, prefix of '.' should be removed", oid)
```

JohnVillalovos: Thanks for the feedback. I have updated it as you suggested. I also updated the commit message.

JohnVillalovos:
Example run:

```
./generator generate
time=2025-04-16T06:03:37.079Z level=INFO source=net_snmp.go:174 msg="Loading MIBs" from=$HOME/.snmp/mibs:/usr/share/snmp/mibs
time=2025-04-16T06:03:37.105Z level=INFO source=main.go:57 msg="Generating config for module" module=avtech
time=2025-04-16T06:03:37.108Z level=ERROR source=main.go:141 msg="Error generating config netsnmp" err="invalid OID .1.3.6.1.4.1.20916, prefix of '.' should be removed"
```

SuperQ:
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

JohnVillalovos:
> This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

Done.

JohnVillalovos:
@SuperQ There is a failing test but it seems unrelated to my change. As it is a failure in downloading a MIB. Please let me know if there are any other changes you would like. Thanks.

SuperQ:
See https://github.com/prometheus/snmp_exporter/issues/1412

SuperQ (approved):

PR #1417, opened 2025-04-19
bring mikrotik mib up to date (7.18.2) (`wifi(wave2/-qcom)` focussed)

hirnpfirsich:
Hello everyone ! :)

this pr bumps the mikrotik mib file to the current stable version of `routeros` (which is `7.18.2`) [source](https://mikrotik.com/download).
The download url now directly annotates which version of the mib file will be downloaded (which is nice).

My focus for this work was to be able to gather `capsman` information from my devices (which is the wifi controller in the `routeros` world - we had support for legacy wifi capsman (<= wifi 5) but not new world capsman (>= wifi 6)).  Therefore I also added a few `overrides` and `lookups` for this specific data.

Because the mib file for mikrotik hasn't been updated in 5 years there are a lot more things one could beautify via `lookups` etc..
Additionally maybe this update introduced breaking changes for some people.
I did not check the mib files so detailed for changes because I thought that this has to happen eventually.

Excited to hear your feedback :)

Greetings from Leipzig (germany) !


hirnpfirsich:
Holy shit your CI/CD is comparing changes between a known good `snmp.yml` and this pr - this so so smart !

Then I am going to mark this comment as obsolete:
> Additionally maybe this update introduced breaking changes for some people.
> I did not check the mib files so detailed for changes because I thought that this has to happen eventually.

My change did not remove any metrics / keys - it only added new ones.
[source](https://app.circleci.com/pipelines/github/prometheus/snmp_exporter/1807/workflows/c049d432-59cc-4bc5-85ec-bc595464877d/jobs/7969)

hirnpfirsich:
@bastischubert done :)

bastischubert (approved):
looks good

PR #1421, opened 2025-04-25
chore: clean up golangci-lint configuration

mmorel-35:
#### Description

chore: clean up golangci-lint configuration

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,7 +16,6 @@ linters:
           severity: warning
           disabled: true
   exclusions:
-    generated: lax
     presets:
       - comments
       - common-false-positives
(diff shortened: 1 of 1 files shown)


kakkoyun (commented):
The same comment:

https://github.com/prometheus/procfs/pull/720#pullrequestreview-2798524287

.golangci.yml
```diff
@@ -1,22 +1,12 @@
...
-      exclude-functions:
-        - (net/http.ResponseWriter).Write
-    revive:
-      rules:
-        - name: unused-parameter
-          severity: warning
-          disabled: true
   exclusions:
-    generated: lax
+    generated: strict
```
bastischubert, line 9: since strict is the default already, do we need it explicitly?

PR #1439, opened 2025-05-28
Added an error message for address resolution in the ConfigureTarget function

XingWei-Liu:
Added an error message for address resolution in the ConfigureTarget function


bastischubert (changes requested):
Why does the test now error and needed to be changed to `shouldErr` : `true` ?

bastischubert:
I cannot see how this improves snmp-exporter - the error message seems to have been moved to a function but i can't see an "added" message. Also some tests now all `shouldErr` set to `true` - why's that?

PR #1464, opened 2025-07-20
Add Generic UPS Mibs

varet80:
I have a ups which is not supported by the mibs included. Found on my SNMP Walk my UPS uses the UPS-MIBs

I am creating a PR to include these in the generated mibs.

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -59,6 +59,7 @@ UBNT_AIROS_URL    := https://dl.ubnt.com/firmwares/airos-ubnt-mib/ubnt-mib.zip
 UBNT_AIROS_OLD_URL:= https://raw.githubusercontent.com/pgmillon/observium/refs/heads/master/mibs/FROGFOOT-RESOURCES-MIB
 UBNT_AIRFIBER_URL := https://dl.ubnt.com/firmwares/airfiber5X/v4.1.0/UBNT-MIB.txt
 UBNT_DL_URL       := http://dl.ubnt-ut.com/snmp
+UPS_MIB_URL       := https://raw.githubusercontent.com/pgmillon/observium/refs/heads/master/mibs/UPS-MIB
 RARITAN_URL       := https://cdn.raritan.com/download/PX/v1.5.20/PDU-MIB.txt
 RARITAN2_URL      := https://cdn1.raritan.com/download/src-g2/4.0.20/PDU2_MIB_4.0.20_49038.txt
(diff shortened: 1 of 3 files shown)


bastischubert:
Hi @varet80 
thanks for your contribution.
In order to get it merged
- can you please sign the commit 
- add  the changes for the snmp.yml to the PR

generator/generator.yml
```diff
@@ -1211,3 +1211,19 @@ modules:
       - source_indexes:
         - bvValue
         lookup: bvDescription
+
+# Generic UPS Mibs from Observioum covering 
+# 1.3.6.1.2.1.33.1
+
+  ups_mib:
+    walk:
+      - 1.3.6.1.2.1.33.1
```
SuperQ, line 1220: I prefer to be a bit more specific than the whole MIB tree. Also use the text names.

```suggestion
      - "UPS-MIB::upsIdent"
      - "UPS-MIB::upsBattery"
      ...
```

generator/Makefile
```diff
@@ -59,6 +59,7 @@ UBNT_AIROS_URL    := https://dl.ubnt.com/firmwares/airos-ubnt-mib/ubnt-mib.zip
 UBNT_AIROS_OLD_URL:= https://raw.githubusercontent.com/pgmillon/observium/refs/heads/master/mibs/FROGFOOT-RESOURCES-MIB
 UBNT_AIRFIBER_URL := https://dl.ubnt.com/firmwares/airfiber5X/v4.1.0/UBNT-MIB.txt
 UBNT_DL_URL       := http://dl.ubnt-ut.com/snmp
+UPS_MIB_URL       := https://raw.githubusercontent.com/pgmillon/observium/refs/heads/master/mibs/UPS-MIB
```
SuperQ, line 62: Hmm, I wish there was a less indirect source for this MIB. I understand the authoritative definition is in [RFC 1628](https://datatracker.ietf.org/doc/html/rfc1628).

The closest thing I know about is the scripts in [net-snmp](https://github.com/net-snmp/net-snmp/tree/master/mibs) that can extract the data from RFCs.

generator/generator.yml
SuperQ: No worries! Thanks for the contribution.

generator/Makefile
varet80: I could not find it in net-snmp's mibs.
that was the only place I could find a mib. if there is an alternative, happy to adjust the url.

Issue #1465, opened 2025-07-24
APC URL no longer available


vvas1lev:
https://download.schneider-electric.com/files?p_Doc_Ref=APC_POWEhttps://download.schneider-electric.com/files?p_Doc_Ref=APC_POWERNETMIB_EN&p_enDocType=FirmwareRNETMIB_EN&p_enDocType=Firmware&p_File_Name=powernet458.mib

Schneider Electric Page not found

make generate mibs
>> Downloading apc-powernet-mib
>> if download fails please check https://www.se.com/at/de/search/?q=powernet+mib&submit+search+query=Search for the latest release
curl: (22) The requested URL returned error: 404
make: *** [Makefile:191: mibs/apc-powernet-mib] Error 22

bastischubert:
this has been fixed in #1467

PR #1467, opened 2025-07-25
changed apc download to prometheus-community/snmp

bastischubert:
change download for the (again) broken schneider website


bastischubert:
fixes #1465

PR #1482, opened 2025-08-13
Add snmp_engineID support

drobbnz:
Added the option for specifying an SNMPv3 EngineID to allow the querying of sub-systems using SNMPv3 proxying.


README.md
```diff
@@ -96,6 +96,10 @@ It is possible to supply an optional `snmp_context` parameter in the URL, like t
 <http://localhost:9116/snmp?auth=my_secure_v3&module=ddwrt&target=192.0.0.8&snmp_context=vrf-mgmt>
 The `snmp_context` parameter in the URL would override the `context_name` parameter in the `snmp.yml` file.
 
+It is also possible when using SNMPv3 to supply an optional `snmp_engineID` parameter in the URL, like this:
```
SuperQ, line 99: Let's stick with either snake or camel case. For URL params, I think we prefer all lower case.

```suggestion
It is also possible when using SNMPv3 to supply an optional `snmp_engineid` parameter in the URL, like this:
```

SuperQ (changes requested):

README.md
drobbnz: Thanks for catching that. Updating now.

SuperQ (approved):
Thanks!

PR #427, opened 2019-06-22
Add config for NEC IX into generator

kamijin-fanta:
NEC IX is a high-quality business router in Japan. It has a Cisco-like CLI.

<details>
  <summary>Click to expand example outputs</summary>

```
# HELP picoCelsius Indicates the temperature of the equipment inside, in degree (Celsius). - 1.3.6.1.4.1.119.2.3.84.2.1.1
# TYPE picoCelsius gauge
picoCelsius 47
# HELP picoExtIfAdminStatus The desired state of the interface. - 1.3.6.1.4.1.119.2.3.84.6.1.1.10
# TYPE picoExtIfAdminStatus gauge
picoExtIfAdminStatus{picoExtIfIndex="1",picoExtIfInstalledSlot="1"} 1
picoExtIfAdminStatus{picoExtIfIndex="2",picoExtIfInstalledSlot="1"} 1
picoExtIfAdminStatus{picoExtIfIndex="3",picoExtIfInstalledSlot="1"} 1
picoExtIfAdminStatus{picoExtIfIndex="4",picoExtIfInstalledSlot="1"} 1
# HELP picoExtIfDescr A textual string containing information about the interface. - 1.3.6.1.4.1.119.2.3.84.6.1.1.3
# TYPE picoExtIfDescr gauge
picoExtIfDescr{picoExtIfDescr="GigaEthernet1.0",picoExtIfIndex="1",picoExtIfInstalledSlot="1"} 1
picoExtIfDescr{picoExtIfDescr="GigaEthernet1.0",picoExtIfIndex="2",picoExtIfInstalledSlot="1"} 1
picoExtIfDescr{picoExtIfDescr="GigaEthernet1.0",picoExtIfIndex="3",picoExtIfInstalledSlot="1"} 1
picoExtIfDescr{picoExtIfDescr="GigaEthernet1.0",picoExtIfIndex="4",picoExtIfInstalledSlot="1"} 1
# HELP picoExtIfDuplex The current mode of this link. - 1.3.6.1.4.1.119.2.3.84.6.1.1.7
# TYPE picoExtIfDuplex gauge
picoExtIfDuplex{picoExtIfIndex="1",picoExtIfInstalledSlot="1"} 2
picoExtIfDuplex{picoExtIfIndex="2",picoExtIfInstalledSlot="1"} 2
picoExtIfDuplex{picoExtIfIndex="3",picoExtIfInstalledSlot="1"} 2
[...]

PR #1497, opened 2025-09-10
generator: Export NAPT metrics from NEC UNIVERGE IX series

yosida95:
This pull request extends the nec_ix module introduced in #427 to support the following four NAPT-related metrics:

* naptCacheEntries
* naptCachePeak
* naptCacheCreates
* naptCacheOverflows

Each metric is annotated with the following four labels:

* ifAlias
* ifDescr
* ifName
* naptCacheIfIndex

According to the MIB files, these metrics were introduced in June 2019, which is also when the original pull request was created.

I understand that keeping the number of metrics small is important, but monitoring the NAPT status is critical for routers.


yosida95:
@bastischubert Thank you for pointing it out. I've signed off the commit. Could you please review it again?

bastischubert (approved):
LGTM

PR #1504, opened 2025-09-25
Added "golang-bin" to RHEL dependicies for running "make generator mibs".

FreeWillyB:
On RHEL-based distros, the "make generator mibs" was failing after downloading the mentioned dependencies. In order to run this successfully, you will also need to ensure that "golang-bin" is also installed on the host.


bastischubert:
@FreeWillyB thanks for your contribution :partying_face: 

will merge once our checks are sane again - could take a couple of days

PR #1510, opened 2025-10-05
Update README.md

llamafilm:
When I tried following the build instructions in a fresh Ubuntu container, I hit some problems, which were solved by installing these dependencies.  Updating the docs to make it easier for others to follow.


bastischubert (approved):
LGTM

PR #1526, opened 2025-11-13
Allow community string as ENV Variable

nradonicich:
Although community strings are not really considered secret, we do have a use-case to provide them as environment variables. Updated is adding Community to be able to come from an environment variable, testing and the README adjustment.


bastischubert:
Do you mind explaining your use-case?

PR #1542, opened 2025-12-19
Handle uint32 type in pduValueAsString, minor gosnmp usage fixes

lukeod:
Fixes a bug where TimeTicks and Uinteger32 values corrupt label values when used in lookups.

`pduValueAsString` handles `uint` but not `uint32`. gosnmp returns TimeTicks and Uinteger32 as `uint32`, so they fall through to the default case which uses `fmt.Sprintf("%s", ...)`. This produces malformed labels like:

```
ifAdminStatus{ifDescr="lo",ifIndex="1",ifLastChange="%!s(uint32=0)"} 1
```

instead of:

```
ifAdminStatus{ifDescr="lo",ifIndex="1",ifLastChange="0"} 1
```

To reproduce, use a lookup on any TimeTicks field:

```yaml
modules:
  test:
    walk: [ifTable]
    lookups:
      - source_indexes: [ifIndex]
        lookup: ifLastChange
```

Also includes two minor correctness improvements:

- **Close() method** - Use `g.c.Close()` instead of `g.c.Conn.Close()` for proper mutex protection. Not a practical issue given the connection-per-scrape model, but better to use the API as intended.

- **EndOfMibView handling** - Added to Get response exception check alongside NoSuchObject and NoSuchInstance. Per RFC 3416 this shouldn't appear in Get responses, but non-compliant devices exist.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -143,7 +143,7 @@ func ScrapeTarget(snmp scraper.SNMPScraper, target string, auth *config.Auth, mo
 			return results, fmt.Errorf("error reported by target %s: Error Status %d", target, packet.Error)
 		}
 		for _, v := range packet.Variables {
-			if v.Type == gosnmp.NoSuchObject || v.Type == gosnmp.NoSuchInstance {
+			if v.Type == gosnmp.NoSuchObject || v.Type == gosnmp.NoSuchInstance || v.Type == gosnmp.EndOfMibView {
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
Usually I prefer separate PRs for fixes. But these are all simple enough.

bastischubert (approved):

Issue #1387, opened 2025-03-07
Exporter is treating "String64" syntax as "octetstring" instead of "displayString"


larsks:
I'm using `snmp_exporter` to gather metrics from some Dell iDRAC devices. I noticed that string values were all showing up as hex strings:

```
systemBIOSManufacturerName{instance="10.2.11.112", job="idrac_snmp", systemBIOSIndex="1", systemBIOSManufacturerName="0x44656C6C20496E632E", systemBIOSchassisIndex="1"}	
```

These OIDs are delcared as `String64` in the MIB:

```
systemBIOSManufacturerName                      OBJECT-TYPE
    SYNTAX      String64
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION
        "0300.0050.0001.0011 This attribute defines the name of the manufacturer
        of the system BIOS."
    ::= { systemBIOSTableEntry 11 }
```

`snmpwalk` displays them like this:

```
$ snmpwalk -M mibs -m ALL -v2c -c public 10.2.11.132 IDRAC-MIB-SMIv2::systemBIOSManufacturerName
IDRAC-MIB-SMIv2::systemBIOSManufacturerName.1.1 = STRING: "Dell Inc."
```

Is the problem that snmp_exporter doesn't know about the `String64` syntax? Evidence suggests that this is a Dell-ism. Is there a way to declare this an alias for `displayString`? It's going to be messy if I need to override all of these individually.

SuperQ:
Yes, it seems like the exporter doesn't understand String64 textual convention.

SuperQ:
It seems like the Dell MIB defines it this way:

```
String64                ::= TEXTUAL-CONVENTION
    STATUS current
    DESCRIPTION
        "General string type for 64 byte strings."
    SYNTAX              OCTET STRING (SIZE (0..64))
```

So, it does say it's a octet string, which isn't a DisplayString. So this appears to be a mistake on the Dell MIB side.

You can override it like this:

```yaml
  dell:
    walk:
      - 1.3.6.1.4.1.674.10892.5.2 # statusGroup
      - 1.3.6.1.4.1.674.10892.5.4 # systemDetailsGroup
      - 1.3.6.1.4.1.674.10892.5.5 # storageDetailsGroup
    overrides:
      systemBIOSManufacturerName:
        type: DisplayString
```

larsks:
> You can override it like this:

Right, but as I said in my earlier comment:

[...]

Issue #1383, opened 2025-03-04
Object gets wrong type compared to MIB


wwijkander:
Hey,
I'm getting metrics with the help of this MIB: [TIMETRA-SERV-MIB](https://github.com/librenms/librenms/blob/master/mibs/nokia/TIMETRA-SERV-MIB).

However, for some reason the object `svcDescription` which has the syntax `ServObjDesc` as it is generated in the snmp.yml gets the type OctetString, and not DisplayString. I feel like this should rather be the latter, based on line 247 of the MIB:

    ServObjDesc                      ::= TEXTUAL-CONVENTION
        STATUS      current
        DESCRIPTION
            "ASCII string used to describe various service objects."
        SYNTAX      DisplayString (SIZE (0..80))

Using an override as detailed in https://www.robustperception.io/why-is-my-snmp-string-showing-as-hexadecimal/ works, but it feels like this is not intended behavior?

SuperQ:
Yes, it seems like we're not recursing the aliased definition of DisplayString syntax.

Issue #607, opened 2021-02-05
FR: PhysAddress type for Fibre Channel/SCSI-WWNs


svalouch:
Fibre Channel and other storage-related communication protocols use longer physical addresses, namely 64 bits instead of 48 bits as used for MAC addresses. The existing `PhysAddress48` type truncates the WWN to 48 bits, rendering it unusable. There are 16 byte / 128 bit WWNs, but I haven't encountered any of those yet. [Wikipedia article: World Wide Name](https://en.wikipedia.org/wiki/World_Wide_Name).

Device is a Brocade 5100 or 6510 FC-switch using the Brocade SW-MIB. The `swFCPortWwn` is defined as follows:
```
  swFCPortWwn		OBJECT-TYPE
	SYNTAX		OCTET STRING (SIZE(8))
	MAX-ACCESS		read-only
	STATUS		current
	DESCRIPTION	"The World_wide_Name of the Fibre Channel port.
		The contents of an instance are in the IEEE extended format
		as specified in FC-PH; the 12-bit port identifier represents
		the port number within the switch."
  ::= { swFCPortEntry 34 }
```

`snmpwalk` returns the following for the OID (except, port number 1):
```
SW-MIB::swFCPortWwn.1 = Hex-STRING: 20 00 00 01 33 AA 15 D7
```
Configuration used for testing:
```yaml
---
brocade_fc:
  walk:
  - 1.3.6.1.4.1.1588.2.1.1.1.6.2.1.34
  metrics:
  - name: swFCPortWwn
    oid: 1.3.6.1.4.1.1588.2.1.1.1.6.2.1.34
    type: OctetString
    help: The World_wide_Name of the Fibre Channel port - 1.3.6.1.4.1.1588.2.1.1.1.6.2.1.34
    indexes:
    - labelname: swFCPortIndex
      type: gauge
```
This yields the following output when queried:
```
swFCPortWwn{swFCPortIndex="1",swFCPortWwn="0x2000000133AA15D7"} = 1
```
If I change the type to `PhysAddress48`, it looks as follows, which misses the last two elements:
```
swFCPortWwn{swFCPortIndex="1",swFCPortWwn="20:00:00:01:33:AA"} = 1
```
The output I want:
```
swFCPortWwn{swFCPortIndex="1",swFCPortWwn="20:00:00:01:33:AA:15:D7"} = 1
```

[...]

Issue #558, opened 2020-08-26
How can convert Hex string to String ? 


se7en102:
I have monitoring Aruba Temperature 
snmpwalk with OID can be see Temperature Min | Max | Current , but snmp_exporter only show Current temperature. Min & Max is Hexa number 0x323243
hpSystemAirMinTemp{hpSystemAirMinTemp="0x323243",hpSystemAirSensor="0"} 1
When convert it "0x323243" to string "22C" 
that is information i want to display in grafana.
So, how can i convert hex to string ini this case ? 
![temperature aruba OID](https://user-images.githubusercontent.com/56648095/91270750-592c9180-e7a3-11ea-8c21-e1ada5647079.jpg)
![snmp_exporter temperature aruba switch](https://user-images.githubusercontent.com/56648095/91270772-60539f80-e7a3-11ea-995d-914a253382dd.jpg)
![2020-08-26_8-18-37](https://user-images.githubusercontent.com/56648095/91270780-62b5f980-e7a3-11ea-88ad-b5a14ea15470.jpg)

brian-brazil:
It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

shitingting521:
so I have same problem ,how to sovle it

PR #1543, opened 2025-12-22
RFC 2579 DISPLAY-HINT Support for Unknown Textual Conventions

lukeod:
Note: This PR should be considered a draft / request for comments. I didn't get a chance to run through more than some very basic real-world testing - away for a week or so, and thought that it might be good to float the current state for feedback, or if anyone else wanted to do some testing with their environment.

There is a "soft" breaking change in that various TCs that currently hit the Octet String format - which may be handled via post-processing, or query-time formatting will suddenly be formatted "correctly". This might be a point of contention, and needs some thought. There are escape hatches available using overrides to maintain the old behavior, but it does require the user to set those.

**TL;DR:** Unknown textual conventions with valid DISPLAY-HINTs now render as formatted strings instead of raw hex. For example, a vendor TC with hint `255a` displays `"PowerEdge R750"` instead of `0x506F776572456467652052373530`. Override with `type: OctetString` to preserve hex output.

## Problem

When the generator encounters an unknown textual convention with an OCTET STRING base type, it falls back to raw hex output even when the MIB contains a valid DISPLAY-HINT that describes how to format the value.

Examples of affected scenarios:
- Vendor-specific TCs (Dell String64, Huawei LUN names, Brocade WWNs)
- Less common standard TCs not explicitly coded
- Any OCTET STRING-based TC with a DISPLAY-HINT that lacks a hardcoded handler

The current approach of having to handle less common TCs is a pretty common pain point for users and maintainers based on some looking through Github issues / discussions / prometheus-users mailing list. Hard coding TCs is fine for common types, but it ends up meaning that the barrier for entry is much higher for users with less common TCs. Currently the solution often requires a PR to add those TCs, which is too hard for many users to solve. There's also a bunch of maintenance and review overhead for the project as the TC support grows.

This provides a generic option for solving a class of issues, and gives users the ability to define custom formatting without having to resort to a bunch of regex-based overrides.

In principle some of the card coded TCs could be removed from the codebase with this approach, but that is currently outside the scope of the PR.

The current intent is to use this as a fallback in the case that there's not hard-coded parsers for a type.

## Scope Limitations

This PR addresses OCTET STRING formatting via DISPLAY-HINT. It does NOT address:

- **INTEGER DISPLAY-HINT scaling** - `d-N` format for implied decimal point (rare, no demand found).
- **Semantic types** - Types needing more than formatting (DateAndTime timezone handling already has dedicated support).

## Solution

A new `DisplayHint` type implements RFC 2579 DISPLAY-HINT parsing. 

### Design: Generator IR / Collector Evaluation

The generator compiles DISPLAY-HINT strings into an intermediate representation (`FormatSpec`) stored in snmp.yml. The collector evaluates this IR at scrape time without having to parse display-hints directly.

**Why IR instead of runtime parsing?**

An alternative would be storing the raw DISPLAY-HINT string in snmp.yml and parsing it at scrape time. The IR approach was chosen for:

1. **Performance** - Parsing happens once at generation, not per-PDU at scrape time. FormatSpec evaluation is a simple slice iteration with no string parsing, regex, or allocations beyond the output buffer.

2. **Validation** - Malformed hints fail at `make generate` with clear errors, not silently at scrape time. Users discover problems before deployment.

3. **Consistency** - Follows the existing snmp_exporter pattern where the generator compiles MIB knowledge into snmp.yml, keeping the collector MIB-agnostic. 

### Generator Phase
- Parses DISPLAY-HINT strings into `FormatOp` intermediate representation
- Stores pre-compiled format spec in snmp.yml
- Unknown TCs with valid hints get `type: DisplayHint` instead of `OctetString`

### Collector Phase
- Applies format spec via slice iteration (no runtime parsing)
- Handles all RFC 2579 format types: decimal (`d`), hex (`x`), octal (`o`), ASCII (`a`), UTF-8 (`t`)
- Implements implicit repetition rule (last spec repeats until data exhausted)
- Proper separator/terminator suppression per RFC

### generator.yml Example:

```yaml
  modules:
    mymodule:
      walk:
        - 1.3.6.1.4.1.12345
      overrides:
        # Force raw hex output (previous behavior)
        someBinaryOid:
          type: OctetString

        # Explicitly use MIB's DISPLAY-HINT
        someVendorTC:
          type: DisplayHint

        # Custom hint for broken/missing MIB hint
        badVendorString:
          display_hint: "64a"
```

### snmp.yml Example

```yaml
# DISPLAY-HINT "1x:" (MAC/PhysAddress) becomes:
format_spec:
  - {take: 1, fmt: x, sep: ":"}
# Single spec; collector applies RFC 2579 implicit repetition
# 6-byte input → "aa:bb:cc:dd:ee:ff"

# DISPLAY-HINT "255a" (ASCII string) becomes:
format_spec:
  - {take: 255, fmt: a}
# Renders bytes as ASCII text

# DISPLAY-HINT "4x-2x-2x-2x-6x" (UUID-style) becomes:
format_spec:
  - {take: 4, fmt: x, sep: "-"}
  - {take: 2, fmt: x, sep: "-"}
  - {take: 2, fmt: x, sep: "-"}
  - {take: 2, fmt: x, sep: "-"}
  - {take: 6, fmt: x}
```

## Sample / Example Issues This Would Address

| Issue | Problem | Vendor |
|-------|---------|--------|
| #1387 | String64 TC as hex (`0x44656C6C20496E632E` instead of "Dell Inc.") | Dell iDRAC |
| #607 | 8-byte WWN as hex (PhysAddress48 truncates to 6 bytes) | Brocade FC |
| #264 | modelName as hex (`0x44533231326A` instead of "DS212j") | Synology |
| #558 | Temperature as hex (`0x323243` instead of "22C") | Aruba |
| #930 | Model name as hex (`0x506F776572456467652052373530` instead of "PowerEdge R750") | Dell |
| #1383 | ServObjDesc TC not recognized as DisplayString derivative | Nokia |

Plus vendor TCs from Huawei (LUN names), HP iLO (timestamps), and others where DISPLAY-HINT exists but no hardcoded handler.

### Behavior Change on Regeneration

When users regenerate `snmp.yml`, unknown TCs with parseable DISPLAY-HINTs will change output format:

| Before | After | Hint |
|--------|-------|------|
| `0xaabbccddeeff` | `aa:bb:cc:dd:ee:ff` | `1x:` |
| `0x506F776572456467652052373530` | `PowerEdge R750` | `255a` |
| `0x44533231326A` | `DS212j` | `64a` |

**Affected downstream configurations:**
- `regex_extracts` patterns expecting hex format
- `metric_relabel_configs` matching hex strings
- Dashboards/alerts with hex string matches
- Automation parsing exporter output

### Escape Hatches

Users can preserve previous behavior:

```yaml
# In generator.yml
overrides:
  someOid:
    type: OctetString  # Forces raw hex output
```

Or specify a custom format:

```yaml
overrides:
  someOid:
    display_hint: "1x."  # Custom hint (dot-separated hex)
```

### Type Override Summary

| Override | Result |
|----------|--------|
| `type: OctetString` | Raw hex (previous behavior) |
| `type: DisplayHint` | Use MIB's DISPLAY-HINT |
| `display_hint: "..."` | Custom hint string (implies DisplayHint type) |

## Performance

Performance is pretty comparable to hard-coded variants:

| Scenario | FormatSpec | Hardcoded Handler | Delta |
|----------|------------|-------------------|-------|
| MAC (6 bytes) | 323 ns | 294 ns | +10% |
| IPv4 (4 bytes) | 77 ns | 89 ns | -13% |
| IPv6 (16 bytes) | 469 ns | 572 ns | -18% |

## Questions for Reviewers

1. Is the breaking change acceptable given the escape hatch exists?
2. is it disruptive enough that it should be opt-in (initially perhaps?) via cli flag or similar?
3. Are there vendor MIBs with unusual DISPLAY-HINTs that should be tested?
4. Any concerns about the snmp.yml schema addition (FormatOp struct design)?
5. Feedback on the FormatSpec IR approach vs alternatives?

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -661,14 +661,26 @@ func pduToSamples(indexOids []int, pdu *gosnmp.SnmpPDU, metric *config.Metric, o
 			}
 		}
 
+		// Get the string value, using FormatSpec for DisplayHint types.
+		var pduStr string
+		if metricType == "DisplayHint" && len(metric.FormatSpec) > 0 {
+			if bytes, ok := pdu.Value.([]byte); ok {
+				pduStr = strings.ToValidUTF8(applyFormatSpec(metric.FormatSpec, bytes), "�")
+			} else {
+				pduStr = pduValueAsString(pdu, metricType, metrics)
+			}
+		} else {
+			pduStr = pduValueAsString(pdu, metricType, metrics)
+		}
+
 		if len(metric.RegexpExtracts) > 0 {
-			return applyRegexExtracts(metric, pduValueAsString(pdu, metricType, metrics), labelnames, labelvalues, logger)
+			return applyRegexExtracts(metric, pduStr, labelnames, labelvalues, logger)
 		}
 		// For strings we put the value as a label with the same name as the metric.
 		// If the name is already an index, we do not need to set it again.
 		if _, ok := labels[metric.Name]; !ok {
 			labelnames = append(labelnames, metric.Name)
-			labelvalues = append(labelvalues, pduValueAsString(pdu, metricType, metrics))
+			labelvalues = append(labelvalues, pduStr)
 		}
 	}
 
@@ -841,11 +853,11 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ string, metrics Metrics) string {
 		for i, o := range v {
 			parts[i] = int(o)
 		}
-		if typ == "OctetString" || typ == "DisplayString" {
+		if typ == "OctetString" || typ == "DisplayString" || typ == "DisplayHint" {
 			// Prepend the length, as it is explicit in an index.
 			parts = append([]int{len(v)}, parts...)
 		}
-		str, _, _ := indexOidsAsString(parts, typ, 0, false, nil)
+		str, _, _ := indexOidsAsString(parts, typ, 0, false, nil, nil)
 		return strings.ToValidUTF8(str, "�")
 	case nil:
 		return ""
@@ -859,7 +871,7 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ string, metrics Metrics) string {
 // Convert oids to a string index value.
 //
 // Returns the string, the oids that were used and the oids left over.
-func indexOidsAsString(indexOids []int, typ string, fixedSize int, implied bool, enumValues map[int]string) (string, []int, []int) {
+func indexOidsAsString(indexOids []int, typ string, fixedSize int, implied bool, enumValues map[int]string, formatSpec []config.FormatOp) (string, []int, []int) {
 	if typeMapping, ok := combinedTypeMapping[typ]; ok {
 		subOid, valueOids := splitOid(indexOids, 2)
 		if typ == "InetAddressMissingSize" {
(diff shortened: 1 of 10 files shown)


lukeod:
I've done some additional work on an alternative - to just do direct display-string parsing at collection time, and think it might be a simpler approach with less performance drawbacks than i had anticipated. It makes the proposed FormatOp IR approach in this PR probably an unnecessary level of indirection / complexity. So i'll close this draft and look to submit a followup PR with the direct parser.

Issue #264, opened 2018-02-14
[Discussion] Do we want to opportunistically parse Octetstring to work around broken vendor implementations?


cukal:
### Host operating system: output of `uname -a`
Linux xxxxx 3.10.0-693.17.1.el7.x86_64 #1 SMP Thu Jan 25 20:13:58 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
### snmp_exporter version: output of `snmp_exporter -version`

snmp_exporter, version 0.8.0 (branch: HEAD, revision: 17cefcaf7010106861ce0faa2a9ec9e39b7c2b4b)
  build user:       root@57f813d2791a
  build date:       20171120-12:36:57
  go version:       go1.9.2
### What device/snmpwalk OID are you using?
Synology DS212j
### If this is a new device, please link to the MIB(s).
[https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip](https://global.download.synology.com/download/Document/MIBGuide/Synology_MIB_File.zip)
### What did you do that produced an error?

generator.yml:
```
modules:
  synology:
    walk:
      - 1.3.6.1.4.1.6574
```

The SYNOLOGY-SYSTEM-MIB contains: 
```
modelName OBJECT-TYPE
    SYNTAX	  OCTET STRING
    MAX-ACCESS  read-only
    STATUS	  current
    DESCRIPTION
    "The Model name of this NAS"
    ::= { dsmInfo 1 }
```

Execute generator

`./generator -generate`

output: 
```
INFO[0000] Loading MIBs from $HOME/.snmp/mibs:/usr/share/snmp/mibs  source="net_snmp.go:109"
WARN[0000] NetSNMP reported 1 parse errors               source="main.go:79"
INFO[0000] Generating config for module synology         source="main.go:36"
INFO[0000] Generated 139 metrics for module synology     source="main.go:39"
INFO[0000] Config written to /root/work/src/github.com/prometheus/snmp_exporter/generator/snmp.yml  source="main.go:63"
```

generated snmp.yml part for modelName:
```
  - name: modelName
    oid: 1.3.6.1.4.1.6574.1.5.1
    type: OctetString
    help: The Model name of this NAS - 1.3.6.1.4.1.6574.1.5.1
```

Created a snmp_exporter snmp.yml file and scraped the target. 

### What did you expect to see?
The same output as with a mib browser:

`modelName.0 | DS212j | OctetString | 192.168.10.10:161`

[...]

PR #1549, opened 2026-01-06
Add DISPLAY-HINT support for SNMP OctetString formatting (v2)

lukeod:
## Summary

This PR adds RFC 2579 DISPLAY-HINT support for formatting OCTET STRING values as readable strings. Unlike #1543 which introduced an intermediate representation compiled at generation time, this implementation uses direct runtime parsing with explicit opt-in per metric/lookup.

**TL;DR:** Add `display_hint: "@mib"` or `display_hint: "255a"` to overrides/lookups to format OctetString values using RFC 2579 DISPLAY-HINT rules. No automatic behavior changes, opt-in.

## Problem

When the exporter encounters an unknown textual convention with an OCTET STRING base type, it falls back to raw hex output even when the MIB contains a valid DISPLAY-HINT describing how to format the value.

Examples:
- Dell String64: `0x506F776572456467652052373530` instead of `PowerEdge R750`
- Synology modelName: `0x44533231326A` instead of `DS212j`
- Vendor MACs with non-standard TCs: `0xAABBCCDDEEFF` instead of `AA:BB:CC:DD:EE:FF`

Adding hardcoded handlers for every vendor TC doesn't scale. Users currently resort to complex `regex_extracts` workarounds or submit PRs for individual TCs. It therefore makes sense to implement a generic approach that implements TC parsing to cover those edge cases. There are quite a few open issues that would benefit from this.

## Solution

A new `display_hint` field on metric overrides and lookups enables RFC 2579 DISPLAY-HINT formatting:

```yaml
modules:
  vendor:
    walk:
      - 1.3.6.1.4.1.vendor
    overrides:
      # Use the DISPLAY-HINT from the MIB
      vendorModelName:
        display_hint: "@mib"

      # Provide a custom hint
      vendorSerial:
        display_hint: "255a"

      # Override an incorrect vendor hint
      vendorIpField:
        display_hint: "1d.1d.1d.1d"

    lookups:
      - source_indexes: [vendorIndex]
        lookup: vendorDescription
        display_hint: "@mib"
```

### Design

The implementation is a single ~200 line function (`applyDisplayHint`) that parses the hint string and applies it to data in one pass at scrape time. This differs from #1543's approach of compiling hints into an IR at generation time, which is more performant, but adds a fair bit of complexity. The assumption is that this is more for edge-case handling rather than being a significant majority of metrics, so it's probably not worth the tradeoff in the name of shaving off a few 10s of nanoseconds per metric.

I've added quite a lot of tests, and also did some pretty extensive fuzzing to ensure no panics with arbitrary input, or malformed/malicious display hints. Ensuring that no combination of inputs/display-hints resulted in infinite loops (not consuming bytes) was a primary consideration, and i'm confident that between the unit tests and fuzzing, that it shouldn't be possible.

It seemed like a big enough chunk of code to justify a new file, rather than tacking onto collector.go. 

Performance on my test machine showed that calling `applyDisplayHint` takes on the order of 60-500ns depending on the complexity. It's dominated by the amount of fmt.PrintF calls for hex formatting. We can implement a much faster hex formatter to reduce this significantly (copy the stdlib hex code, and change the lookup table to be uppercase), which would make it competitive with the existing hard-coded TC handlers, but it seemed like this was a big enough PR without including micro optimizations that would increase the scope.

I've opted to use uppercase notation for hex bytes. This seemed consistent with the rest of the project.

### RFC 2579 Compliance

Implements RFC 2579 Section 3.1:
- Format characters: `d` (decimal), `x` (hex), `o` (octal), `a` (ASCII), `t` (UTF-8)
- Implicit repetition rule (last spec repeats until data exhausted)
- Star prefix (`*`) repeat indicator for dynamic repeat counts
- Separator and terminator handling
- Trailing separator suppression

### Output Examples

| Hint | Input Bytes | Output |
|------|-------------|--------|
| `255a` | `[72,101,108,108,111]` | `Hello` |
| `1x:` | `[0x00,0x1a,0x2b,0x3c,0x4d,0x5e]` | `00:1A:2B:3C:4D:5E` |
| `1d.1d.1d.1d` | `[192,168,1,1]` | `192.168.1.1` |
| `2x:` | (16 bytes) | `2001:0DB8:0000:...` |
| `2d-1d-1d,1d:1d:1d.1d` | (8 bytes) | `2022-8-15,8:1:15.0` |

## Sample Issues This Addresses

| Issue | Problem | Vendor |
|-------|---------|--------|
| #1387 | String64 TC as hex | Dell iDRAC |
| #607 | 8-byte WWN as hex | Brocade FC |
| #264 | modelName as hex | Synology |
| #558 | Temperature as hex | Aruba |
| #930 | Model name as hex | Dell |
| #1383 | ServObjDesc TC not recognized | Nokia |

## Breaking Changes

**None.** This is fully opt-in. Existing configurations continue to work unchanged.

Users must explicitly add `display_hint` to overrides or lookups to enable formatting. Without it, OctetString values render as hex (current behavior).

## Non-Breaking Behavior

| Scenario | Behavior |
|----------|----------|
| No `display_hint` specified | Hex output (unchanged) |
| `display_hint: "@mib"` | Format using MIB's DISPLAY-HINT |
| `display_hint: "@mib"` but MIB has no hint | Warning logged at generation-time, hex output at collection |
| `display_hint: "255a"` | Format using user-provided hint |
| Malformed hint | Warning logged, hex fallback |

## Types & interactions

The `display_hint` field only applies to `OctetString` type.  

Regex_extracts:

When both are specified, `display_hint` applies first:

1. Raw bytes from SNMP
2. `display_hint` formats to string
3. `regex_extracts` matches against formatted string

This allows regex patterns to operate on human-readable values rather than hex.

## Scope Limitations

This PR addresses OCTET STRING formatting only. It does NOT address:

- **INTEGER DISPLAY-HINT scaling** - `d-N` format for implied decimal point (rare)
- **Index formatting** - Index types (PhysAddress48, DisplayString, etc.) already cover common cases. Wouldn't be too hard to add.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ executors:
   # should also be updated.
   golang:
     docker:
-    - image: cimg/go:1.25
+    - image: cimg/go:1.26
     parameters:
       working_dir:
         type: string
@@ -24,7 +24,7 @@ jobs:
     - prometheus/setup_environment
     - run: sudo apt-get update
     - run: sudo apt-get -y install build-essential libsnmp-dev
-    - run: make
+    - run: make build test
     - run: git diff --exit-code
     - prometheus/store_artifact:
         file: snmp_exporter
@@ -105,40 +105,14 @@ workflows:
         filters:
           tags:
             only: /.*/
-    - prometheus/build:
-        name: build
-        filters:
-          tags:
-            only: /.*/
     - generator:
         filters:
           tags:
             only: /.*/
-    - prometheus/publish_main:
-        context: org-context
-        requires:
-        - test
-        - build
-        - generator
-        filters:
-          branches:
-            only: main
-    - prometheus/publish_release:
-        context: org-context
-        requires:
-        - test
-        - build
-        - generator
-        filters:
-          tags:
-            only: /^v[0-9]+(\.[0-9]+){2}(-.+|[^-.]*)$/
-          branches:
-            ignore: /.*/
     - publish_generator_main:
         context: org-context
         requires:
         - test
-        - build
         - generator
         filters:
           branches:
@@ -147,7 +121,6 @@ workflows:
         context: org-context
         requires:
         - test
-        - build
         - generator
         filters:
           tags:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,120 @@
+---
+name: CI
+on:
+  pull_request:
+  push:
+    branches: [main, 'release-*']
+    tags: ['v*']
+
+jobs:
+  test_go:
+    name: Go tests
+    runs-on: ubuntu-latest
+    container:
+      # Whenever the Go version is updated here, .promu.yml
+      # should also be updated.
+      image: quay.io/prometheus/golang-builder:1.26-base
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: prometheus/promci@fc721ff8497a70a93a881cd552b71af7fb3a9d53 # v0.5.4
+      - uses: ./.github/promci/actions/setup_environment
+      - run: make
+      - run: git diff --exit-code
+
+  # generator:
+  #   name: Go tests
+  #   runs-on: ubuntu-latest
+  #   container:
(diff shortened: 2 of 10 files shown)


SuperQ (approved):
Finally got a chance to really look this over. LGTM, let's go for it.

PR #1554, opened 2026-01-29
Add branchTable to servertech_sentry3 module

bduffany:
The Sentry3 MIB includes branch current monitoring metrics in branchTable (1.3.6.1.4.1.1718.3.2.7) which are useful for monitoring per-branch load on older ServerTech PDUs, which is important because branches have their own current limits. This is already included in the newer servertech_sentry4 module as st4BranchMonitorTable.

Adds metrics like branchLoadValue, branchCapacity, and branchLoadStatus.


SuperQ (approved):
Thanks!

PR #1590, opened 2026-03-21
fix: use comma-ok type assertion and length guard in parseNtpTimestamp

Aprazor:
The bare type assertion pdu.Value.([]byte) panics if the PDU value
is not []byte. The subsequent data[:4] and data[4:] slicing panics
if data is shorter than 8 bytes. Add proper type check and length
validation with descriptive error messages.


SuperQ (changes requested):
This needs a test case.

SuperQ (approved):
Thanks!

PR #1606, opened 2026-04-08
Split generator CI

SuperQ:
Split the generator CI into a separate CI flow to avoid mixing CI caches.


SuperQ:
I'm not sure, but I think this should fix the CI failures on merge to main.

bastischubert (approved):

PR #1610, opened 2026-04-17
feat: allow display_hint config for indexes

jonaskuske:
In https://github.com/prometheus/snmp_exporter/pull/1549, support for DISPLAY-HINT was added for metrics/lookups. However, this does not cover labels/strings coming from Index OIDs.

This PR extends the `Index` type and `indexOidsAsString` function to also handle display hints. This allows e.g. cases where IP addresses are used as index in OctetString format.

I was not sure how to best integrate this into the generator and its config, so I've left this part out of the PR for now. Happy to receive some pointers on how this could be integrated into the generator step, otherwise this could be merged as-is and require manual config adjustment for now, and generator functionality could come later?

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -844,7 +844,7 @@ func pduValueAsString(pdu *gosnmp.SnmpPDU, typ, displayHint string, metrics Metr
 			// Prepend the length, as it is explicit in an index.
 			parts = append([]int{len(v)}, parts...)
 		}
-		str, _, _ := indexOidsAsString(parts, typ, 0, false, nil)
+		str, _, _ := indexOidsAsString(parts, typ, "", 0, false, nil)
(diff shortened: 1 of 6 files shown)


SuperQ (approved):

SuperQ:
Yea, I think this is fine for now. Plumbing this into the generator shouldn't be difficult.

Maybe @lukeod would be able to contribute this?

bastischubert (approved):

SuperQ:
This needs a rebase.

jonaskuske:
I've rebased and updated the PR against the current main. Simply committed the `snmp.yml` that the generator now outputs fyi, not familiar enough with the project to be able to review whether the changes there are fine or not.

Anyhow, seeing that the preparation work in #1646, #1648 and #1653 was merged, what's the status here? :)

PR #1624, opened 2026-06-08
feat: map enum lookup columns to string labels

### LINKED ISSUE #659: Label values not being resolved on metric export
### Host operating system: 
`Linux 4.15.0-55-generic #60-Ubuntu SMP Tue Jul 2 18:22:20 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`

### snmp_exporter version: 
`0.20.0`

See below.
### If this is a new device, please link to the MIB(s).
PROJECT-CLEARWATER-MIB. See here: https://github.com/Metaswitch/clearwater-snmp-handlers/blob/master/mib-generator/CLEARWATER-MIB-COMMON

Relevant snippet below:
```
...
sproutHomesteadLatencyEntry OBJECT-TYPE
    SYNTAX      SproutHomesteadLatencyEntry
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION "The average, variance, LWM and HWM of request
                 latency"
    INDEX       { sproutHomesteadLatencyScope }
    ::= { sproutHomesteadLatencyTable 1 }

SproutHomesteadLatencyEntry ::= SEQUENCE
{
  sproutHomesteadLatencyScope          INTEGER,
  sproutHomesteadLatencyAverage        Unsigned32,
  sproutHomesteadLatencyVariance       Unsigned32,
  sproutHomesteadLatencyHWM            Unsigned32,
  sproutHomesteadLatencyLWM            Unsigned32,
  sproutHomesteadLatencyCount          Unsigned32
}

sproutHomesteadLatencyScope OBJECT-TYPE
    SYNTAX      INTEGER {
                  scopePrevious5SecondPeriod(1),
                  scopeCurrent5MinutePeriod(2),
                  scopePrevious5MinutePeriod(3)
                }
    MAX-ACCESS  not-accessible
    STATUS      current
    DESCRIPTION "The period over which the statistics were collected."
    ::= { sproutHomesteadLatencyEntry 1 }

sproutHomesteadLatencyAverage OBJECT-TYPE
    SYNTAX      Unsigned32
    MAX-ACCESS  read-only
    STATUS      current
    DESCRIPTION "The average latency over the period."
    ::= { sproutHomesteadLatencyEntry 2 }
...
```

We generate a config snmp.yaml using the generator. We run the snmp_exporter using this generated snmp.yaml. Process is running, metrics are being exposed.

Metrics should have looked like the following:
```
# HELP sproutHomesteadLatencyAverage The average latency over the period. - 1.2.826.0.1.1578918.9.3.3.2.1.2
# TYPE sproutHomesteadLatencyAverage gauge
sproutHomesteadLatencyAverage{sproutHomesteadLatencyScope="scopeCurrent5MinutePeriod"} 0
sproutHomesteadLatencyAverage{sproutHomesteadLatencyScope="scopePrevious5MinutePeriod"} 0
sproutHomesteadLatencyAverage{sproutHomesteadLatencyScope="scopePrevious5SecondPeriod"} 0
```

### What did you see instead?
Metrics are exposed in the following format:
```
# HELP sproutHomesteadLatencyAverage The average latency over the period. - 1.2.826.0.1.1578918.9.3.3.2.1.2
# TYPE sproutHomesteadLatencyAverage gauge
sproutHomesteadLatencyAverage{sproutHomesteadLatencyScope="1"} 0
sproutHomesteadLatencyAverage{sproutHomesteadLatencyScope="2"} 0
sproutHomesteadLatencyAverage{sproutHomesteadLatencyScope="3"} 0
```

Note that the labels are being generated with the enum keys instead of values.

### Additional Details

We used the following config for generator.yaml:
```
modules:
  PROJECT-CLEARWATER-MIB:
    walk:
      - sprout

    version: 2

    auth:
      # Community string is used with SNMP v1 and v2. Defaults to "public".
      community: clearwater
```

Relevant portion in generated snmp.yaml is as follows:
```
...
  - name: sproutHomesteadLatencyScope
    oid: 1.2.826.0.1.1578918.9.3.3.2.1.1
    type: gauge
    help: The period over which the statistics were collected. - 1.2.826.0.1.1578918.9.3.3.2.1.1
    indexes:
    - labelname: sproutHomesteadLatencyScope
      type: gauge
      enum_values:
        1: scopePrevious5SecondPeriod
        2: scopeCurrent5MinutePeriod
        3: scopePrevious5MinutePeriod
    enum_values:
      1: scopePrevious5SecondPeriod
      2: scopeCurrent5MinutePeriod
      3: scopePrevious5MinutePeriod
  - name: sproutHomesteadLatencyAverage
    oid: 1.2.826.0.1.1578918.9.3.3.2.1.2
    type: gauge
    help: The average latency over the period. - 1.2.826.0.1.1578918.9.3.3.2.1.2
    indexes:
    - labelname: sproutHomesteadLatencyScope
      type: gauge
      enum_values:
        1: scopePrevious5SecondPeriod
        2: scopeCurrent5MinutePeriod
        3: scopePrevious5MinutePeriod
...
```

As an hacky solution, changing the `type: gauge` line to `type: EnumAsInfo` under the indexes section under the `sproutHomesteadLatencyAverage` metric in the above snmp.yaml produces the metrics in the desired form.

The generated snmp.yaml seems to be incorrect. Is this due to some missing config in the `generator.yaml` or is this an issue with the generator?

llamafilm:
## Summary

Fixes #659.

SNMP table columns with an `INTEGER` type that has MIB enum values (e.g. `cempMemPoolType`, `ifType`) were emitting integer strings like `"2"` as label values when used as lookups, instead of the human-readable enum string like `"ioMemory"`. The workaround was to manually edit the generated `snmp.yml`.

## Changes

**`config/config.go`** — Added `EnumValues map[int]string` to `Lookup`, parallel to the field that already exists on `Index`.

**`generator/tree.go`** — When building a lookup entry, copy `enum_values` from the MIB node onto the `Lookup` (and any chained index).

**`collector/collector.go`** — Two changes:
- Lookup resolution: when `lookup.Type == "EnumAsInfo"` and `lookup.EnumValues` is non-empty, map the raw integer through the enum table. Falls back to the decimal string for unknown values, consistent with existing `EnumAsInfo` behavior elsewhere.
- Guard in `enumAsInfo`: when a metric column has type `EnumAsInfo` and is also its own table index with type `EnumAsInfo`, the metric name would appear in `labelnames` twice, causing Prometheus to reject the metric. Added an early return when the metric name is already present as a label.

## Test plan

Three new cases in `TestIndexesToLabels`:
1. Enum lookup maps integer `2` → `"ioMemory"`
2. Integer not in enum map falls back to `"99"`
3. Enum lookup with `drop_source_indexes` drops the source label, leaving only the resolved string

All existing tests pass: `go test ./...`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -711,6 +711,14 @@ func enumAsInfo(metric *config.Metric, value int, labelnames, labelvalues []stri
 	if !ok {
 		state = strconv.Itoa(int(value))
 	}
+	// If the metric name is already a label (e.g. it is also a table index with
+	// type EnumAsInfo), the enum string is already captured there and we must not
+	// add it again or Prometheus will reject the duplicate label.
+	for _, ln := range labelnames {
+		if ln == metric.Name {
+			return []prometheus.Metric{}
+		}
+	}
 	labelnames = append(labelnames, metric.Name)
 	labelvalues = append(labelvalues, state)
 
@@ -1004,7 +1012,16 @@ func indexesToLabels(indexOids []int, metric *config.Metric, oidToPdu map[string
 					}
 				}
 			}
-			labels[lookup.Labelname] = pduValueAsString(&pdu, t, lookup.DisplayHint, metrics)
+			if t == "EnumAsInfo" && len(lookup.EnumValues) > 0 {
+				intVal := int(gosnmp.ToBigInt(pdu.Value).Int64())
+				if str, ok := lookup.EnumValues[intVal]; ok {
+					labels[lookup.Labelname] = str
+				} else {
+					labels[lookup.Labelname] = strconv.Itoa(intVal)
+				}
+			} else {
+				labels[lookup.Labelname] = pduValueAsString(&pdu, t, lookup.DisplayHint, metrics)
+			}
 			labelOids[lookup.Labelname] = []int{int(gosnmp.ToBigInt(pdu.Value).Int64())}
 		} else {
 			labels[lookup.Labelname] = ""
--- a/collector/collector_test.go
+++ b/collector/collector_test.go
@@ -1221,6 +1221,58 @@ func TestIndexesToLabels(t *testing.T) {
 			},
 			result: map[string]string{"lldpRemTimeMark": "1", "lldpRemLocalPortNum": "8", "lldpRemIndex": "1", "lldpLocPortId": "04:05:06:07:08:09"},
 		},
+		{
+			oid: []int{2},
+			metric: config.Metric{
+				Indexes: []*config.Index{{Labelname: "idx", Type: "gauge"}},
+				Lookups: []*config.Lookup{
+					{
+						Labels:     []string{"idx"},
+						Labelname:  "poolType",
+						Oid:        "1.2.3",
+						Type:       "EnumAsInfo",
+						EnumValues: map[int]string{1: "processorMemory", 2: "ioMemory", 3: "pciMemory"},
+					},
+				},
+			},
+			oidToPdu: map[string]gosnmp.SnmpPDU{"1.2.3.2": {Value: 2}},
+			result:   map[string]string{"idx": "2", "poolType": "ioMemory"},
+		},
+		{
+			oid: []int{2},
+			metric: config.Metric{
+				Indexes: []*config.Index{{Labelname: "idx", Type: "gauge"}},
+				Lookups: []*config.Lookup{
+					{
+						Labels:     []string{"idx"},
+						Labelname:  "poolType",
+						Oid:        "1.2.3",
+						Type:       "EnumAsInfo",
+						EnumValues: map[int]string{1: "processorMemory", 2: "ioMemory", 3: "pciMemory"},
+					},
+				},
+			},
+			oidToPdu: map[string]gosnmp.SnmpPDU{"1.2.3.2": {Value: 99}},
+			result:   map[string]string{"idx": "2", "poolType": "99"},
+		},
+		{
+			oid: []int{2},
+			metric: config.Metric{
+				Indexes: []*config.Index{{Labelname: "idx", Type: "gauge"}},
+				Lookups: []*config.Lookup{
+					{
+						Labels:     []string{"idx"},
+						Labelname:  "poolType",
+						Oid:        "1.2.3",
+						Type:       "EnumAsInfo",
+						EnumValues: map[int]string{1: "processorMemory", 2: "ioMemory", 3: "pciMemory"},
+					},
+					{Labelname: "idx"},
+				},
+			},
+			oidToPdu: map[string]gosnmp.SnmpPDU{"1.2.3.2": {Value: 2}},
+			result:   map[string]string{"poolType": "ioMemory"},
+		},
 	}
 	for _, c := range cases {
 		got := indexesToLabels(c.oid, &c.metric, c.oidToPdu, Metrics{})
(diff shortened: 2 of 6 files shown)


SuperQ:
> Downloading Eltex MES device mibs to /tmp/tmp.vM4M3E2QVA
> curl: (60) SSL certificate problem: certificate has expired

:sigh:

llamafilm:
This may be a breaking change for some people because it changes a lot of indexes from gauge to EnumAsInfo.  We could potentially avoid that by only making this lookup happen if the generator config explicitly includes `type: EnumAsInfo`.  But I did it this way because in the linked issue @SuperQ said it should be the default.

Before: `cempMemPoolUsed{entPhysicalIndex="1", cempMemPoolType="3"} 45678912`
After: `cempMemPoolUsed{entPhysicalIndex="1", cempMemPoolType="ioMemory"} 45678912`

SuperQ:
Hmm, yea, tricky with the breaking change. I still think the `EnumAsInfo` is the correct default as long as users can opt-out with `type: gauge` to go back to the numeric version.

SuperQ:
Should we also explicitly reject `type: EnumAsStateSet` for indexes? That combination doesn't make sense to me.

llamafilm:
Good point. I fixed it so that adding `type: gauge` restores the old behavior.

I agree it doesn't make sense to use EnumAsStateSet as an index.  If someone tried this, it would be silently ignored by the collector and labels show the integer.

SuperQ:
Silently ignored is not great. We want to reject invalid configs.

llamafilm:
I was a bit confused in my last comment.  I thought you were talking about rejecting EnumAsStateSet on lookups.  That is actually a legitimate user choice, although I would probably never do it myself, as I always add `ignore: true` to any metrics being applied as a lookup.  But if someone wanted to have that value as both a lookup and as a separate state set metric, that is possible.

I agree that EnumAsStateSet on an index is invalid and should throw an error.  However, that has nothing to do with this feature PR so I think that ought to be tracked as a separate issue.

SuperQ:
We fixed up the generator issues, would you mind doing a rebase?

llamafilm:
ok, done

llamafilm:
Do you want me to add something to the readme about this?  I'm not sure how that's handled, since I've never contributed to this project.

SuperQ:
Probably worth documenting in the generator docs. Not sure exactly what yet.

bastischubert:
i'm generally fine with merging this - also would propose to add a breaking change section to the changelog (tbd later this week) and an explanation to the generator docs as well.

bastischubert (approved):

JDA88:
Unless I have missed something using the `overrides` 
```
    overrides:
      entSensorType:
        type: gauge
```
fixes the generator issue but if the OID is used in the 'lookups' and we try to load the `snmp.yml` i get the following error messages:
```
  line xx1: field enum_values not found in type config.Lookup
  line xxx7: field enum_values not found in type config.Lookup
...
```
The 'lookups' section look like this and from my understanding it doesn't like having an `enum_values` field in the 'lookups'
```
      lookups:
      - labels:
        - entPhysicalIndex
        labelname: entSensorType
        oid: 1.3.6.1.4.1.9.9.91.1.1.1.1.1
        type: gauge
        enum_values:
          1: other
          2: unknown
          3: voltsAC
          4: voltsDC
          5: amperes
          6: watts
          7: hertz
          8: celsius
          9: percentRH
          10: rpm
          11: cmm
          12: truthvalue
          13: specialEnum
          14: dBm
```

Does that mean another generatof fix or a new release to snmp_exporter?

PR #1625, opened 2026-06-12
collector: fix multiple dynamic filters on the same OID overriding each other

### LINKED ISSUE #1621: Dynamic filters – only the last configured filter is applied
When more than one filter is defined, only the last filter defined in snmp.yml takes effect. This makes it impossible to build complex selections such as “only interfaces whose description is not empty and whose type is Ethernet”.

koushik717:
Fixes #1621

When you define more than one `dynamic` filter and they target the same OID, only the last one actually has an effect. I dug into `ScrapeTarget` and the issue is that each filter's loop iteration calls `updateGetConfig` which strips out the indices that the previous filter just added, then `addAllowedIndices` adds back only its own matches. So if filter A adds indices [2,3] and filter B then runs on the same target, it removes [2,3] first and adds whatever it matched on its own, basically wiping out filter A's work.

This makes it impossible to do something like "interfaces where description isn't empty AND ifType is ethernet" since the two filters would target the same set of OIDs.

What I changed: instead of mutating the get config inside the filter loop, I first collect each filter's allowed indices per target OID, and if a target OID was already touched by a previous filter I intersect the two index lists (AND semantics). Once all filters have run, I apply the final intersected list to the get config, once per target OID.

Added a test case (`dynamic_filter_intersection`) with two filters on the same target OID to cover this. Ran `go test ./collector/...` and everything passes including the existing filter tests, also did `go build ./...`.

Happy to adjust the approach if there's a preferred direction for this.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -93,6 +93,14 @@ func ScrapeTarget(snmp scraper.SNMPScraper, target string, auth *config.Auth, mo
 	// Evaluate rules.
 	newGet := module.Get
 	newWalk := module.Walk
+
+	// allowedIndicesByTarget accumulates, for each target OID, the
+	// intersection of the indices allowed by every filter targeting it.
+	// This makes multiple filters that target the same OID act as an AND,
+	// rather than the later filter discarding the earlier one's results.
+	allowedIndicesByTarget := map[string][]string{}
+	var filteredTargets []string
+
 	for _, filter := range module.Filters {
 		allowedList := []string{}
 		pdus, err := snmp.WalkAll(filter.Oid)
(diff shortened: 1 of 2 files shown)


koushik717:
@SuperQ @bastischubert @RichiH could one of you take a look when you get a chance? Happy to make changes if the intersection approach isn't what you'd want here.

koushik717:
@SuperQ @bastischubert @RichiH the Generator / Build generator failure looks like a pre-existing issue on main (failing on the last 5+ runs there too) seems to be the make mibs step hitting a network issue. not related to this change, all the Go tests, lint and build checks are passing.

bastischubert (approved):

SuperQ (approved):
Thanks

PR #1629, opened 2026-06-15
fix: reject empty module query parameter

immanuwell:
Hi @bastischubert, tiny fix.

`/snmp?target=127.0.0.1&module=` is accepted today.
That leaves the module list empty, then the scrape returns 200 and no SNMP metrics. Kinda a silent bad request.

This change rejects an explicit empty `module` with `400`.
Omitting `module` still keeps the `if_mib` default.

Repro before:
1. Run `./snmp_exporter`.
2. Call `curl -i 'http://127.0.0.1:9116/snmp?target=127.0.0.1&module='`.
3. It returns `200`, but no module gets scraped.

After this change:
1. Same request returns `400`.
2. Body says `'module' parameter must contain at least one module name`.

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -14,11 +14,18 @@
 package main
 
 import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/prometheus/common/promslog"
 	"go.yaml.in/yaml/v2"
+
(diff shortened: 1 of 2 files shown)


bastischubert:
@immanuwell could you please fix the two concerns of the go linter so we can go on and merge?

immanuwell:
@bastischubert done, fixed

bastischubert (approved):
looks good

PR #1631, opened 2026-06-18
docs: fix hex casing in display_hint example comment

s3onghyun:
The '1x:' example in display_hint.go showed lowercase hex (00:1a:2b...), but writeHex emits uppercase (and the test asserts 00:1A:2B:3C:4D:5E). Corrected the comment to uppercase.


bastischubert (approved):
LGTM

PR #1637, opened 2026-06-30
fix: reject invalid dynamic filter regexes during config loading

immanuwell:
Bad dynamic filter regexes are accepted during config load, then can panic later when the scrape path hits `regexp.MustCompile`.

This makes the exporter reject that config earlier, and keeps the collector off the panic path too. Pretty small fix, but useful.

Repro:
1. Use a dynamic filter with `values: ["("]`
2. Start `snmp_exporter`. Config load is fine on current `main`
3. Hit a scrape that evaluates that filter
4. The exporter panics with `missing closing )`

With this patch, config loading fails early with `invalid dynamic filter value "("`.

cc @SuperQ

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -19,7 +19,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log/slog"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -101,7 +100,8 @@ func ScrapeTarget(snmp scraper.SNMPScraper, target string, auth *config.Auth, mo
 	allowedIndicesByTarget := map[string][]string{}
 	var filteredTargets []string
 
-	for _, filter := range module.Filters {
+	for i := range module.Filters {
+		filter := &module.Filters[i]
 		allowedList := []string{}
 		pdus, err := snmp.WalkAll(filter.Oid)
 		// Do not try to filter anything if we had errors.
@@ -113,7 +113,7 @@ func ScrapeTarget(snmp scraper.SNMPScraper, target string, auth *config.Auth, mo
 		allowedList = filterAllowedIndices(logger, filter, pdus, allowedList, metrics)
 
 		// Update config to get only index and not walk them.
-		newWalk = updateWalkConfig(newWalk, filter, logger)
+		newWalk = updateWalkConfig(newWalk, *filter, logger)
 
 		for _, targetOid := range filter.Targets {
 			if existing, ok := allowedIndicesByTarget[targetOid]; ok {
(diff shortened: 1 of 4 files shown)


collector/collector.go
```diff
@@ -197,13 +197,22 @@ func intersectIndices(a, b []string) []string {
 
 func filterAllowedIndices(logger *slog.Logger, filter config.DynamicFilter, pdus []gosnmp.SnmpPDU, allowedList []string, metrics Metrics) []string {
 	logger.Debug("Evaluating rule for oid", "oid", filter.Oid)
+	regexps := make([]*regexp.Regexp, 0, len(filter.Values))
+	for _, value := range filter.Values {
+		re, err := regexp.Compile(value)
+		if err != nil {
+			logger.Error("Skipping invalid dynamic filter value", "value", value, "err", err)
```
SuperQ: Hmm, I'm not a fan of skipping like this at runtime.

What if we changed `type DynamicFilter` to have an unexported field that is populated with the result of `regexp.Compile() in `func (c *DynamicFilter) UnmarshalYAML()`.

This way we get two advantages.
* We do all the regexp compiles at config load time.
* We don't need a second compile at every scrape.

immanuwell: got it, fixed.
the regexes are now compiled once during DynamicFilter YAML unmarshal and stored on the filter, and the collector uses the precompiled regexes at scrape time.
I also kept the config-load rejection test

immanuwell:
@SuperQ addressed your suggestions, PTAL

config/config.go
```diff
@@ -230,9 +230,35 @@ type StaticFilter struct {
...
+		}
+		regexps = append(regexps, re)
+	}
+	c.regexps = regexps
+	return nil
+}
+
+func (c *DynamicFilter) Regexps() []*regexp.Regexp {
+	return c.regexps
+}
```
SuperQ, line 254: Minor nit, can we call this `Filters` for consistency with what it is?


```suggestion
func (c *DynamicFilter) Filters() []*regexp.Regexp {
	return c.filters
}
```

config/config.go
```diff
@@ -230,9 +230,35 @@ type StaticFilter struct {
 	Indices []string `yaml:"indices,omitempty"`
 }
 type DynamicFilter struct {
-	Oid     string   `yaml:"oid"`
-	Targets []string `yaml:"targets,omitempty"`
-	Values  []string `yaml:"values,omitempty"`
+	Oid     string           `yaml:"oid"`
+	Targets []string         `yaml:"targets,omitempty"`
+	Values  []string         `yaml:"values,omitempty"`
+	regexps []*regexp.Regexp `yaml:"-"`
```
SuperQ, line 236:
```suggestion
	filters []*regexp.Regexp `yaml:"-"`
```

SuperQ (commented):
Nice, minor nit about naming.

PR #1645, opened 2026-07-01
ci(generator): publish snmp-generator image directly

bastischubert:
publish_main builds the main snmp_exporter binary; use make -C generator docker instead to publish prom/snmp-generator.


bastischubert:
@SuperQ this should be the last part after promci(-images) to make the release workflow work again

SuperQ (approved):
:shipit:

PR #1651, opened 2026-07-04
fix: return partial metrics when one module fails (ContinueOnError)

amuntean-godaddy:
## Summary

When scraping multiple modules in a single request, if any one module times out or errors, `promhttp` previously discarded all collected metrics from the other modules and returned HTTP 500 with only the error message.

This change sets `ErrorHandling: promhttp.ContinueOnError` on the handler, allowing the response to include both the `snmp_error` metric for the failed module and all successfully collected metrics from healthy modules.

## Changes

- `main.go`: set `ErrorHandling: promhttp.ContinueOnError` on the per-scrape `promhttp.HandlerFor` call
- `scraper/mock.go`: add `SetWalkError(oid, err)` helper for per-OID error injection in tests
- `collector/collector_test.go`: add `TestContinueOnErrorHTTPHandler` — an httptest integration test that verifies the HTTP response body contains good-module metrics even when another module errors; the test fails without the `main.go` change

## Testing

```
go test ./... -race
```

All packages pass.

### DIFF
--- a/collector/collector_test.go
+++ b/collector/collector_test.go
@@ -14,15 +14,24 @@
 package collector
 
 import (
+	"context"
 	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
 	"reflect"
 	"regexp"
 	"strings"
 	"testing"
+	"time"
 
 	kingpin "github.com/alecthomas/kingpin/v2"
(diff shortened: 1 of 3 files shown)


SuperQ:
Failing on error is intentional. Having partial failure does not follow Prometheus best practices.

amuntean-godaddy:
@SuperQ Thanks for point this out, closing this.

PR #1652, opened 2026-07-04
feat: parallel OID walks with per-module walk_concurrency config

amuntean-godaddy:
## Summary

Sequential walking of N OID subtrees in a module takes sum(N) seconds. On devices with large interface tables (e.g. a 48-port switch with hundreds of sub-interfaces, 19-OID `if-mib` at 45ms RTT), the walk can exceed the configured scrape timeout.

This PR adds parallel OID walks within a single module, controlled by a new `walk_concurrency` field in the module config. The default is `1` — identical to the current sequential behaviour — so existing configs are completely unaffected.

### New config field: `walk_concurrency`

```yaml
modules:
  if_mib:
    walk_concurrency: 4   # walk 4 OID subtrees in parallel; default 1 (sequential)
    max_repetitions: 10
    timeout: 5s
    walk:
      - ifAdminStatus
      - ifOperStatus
      # ...
```

Propagates through `generator.yml` → `snmp.yml` like `max_repetitions` and `timeout`.

### Implementation

**Fixed worker pool** (`collector/collector.go`): `ScrapeTarget` now accepts `ctx context.Context` as its first parameter. When `walk_concurrency > 1`, a fixed pool of `min(walk_concurrency, len(walk))` workers is created, each with its own cloned SNMP connection. Workers pull OID subtrees from a pre-filled buffered channel. `errgroup.WithContext(ctx)` cancels all workers on the first error and `g.Wait()` guarantees `ScrapeTarget` never returns with live goroutines.

**`Clone()` on `SNMPScraper`** (`scraper/`): Each worker gets its own connection via `Clone()`. The implementation does a field-by-field copy of `GoSNMP` and explicitly:
- Zeroes callback fields (`OnSent`, `OnRecv`, `OnRetry`, `OnFinish`, `PreSend`) — they capture per-scrape counters by reference
- Strips the port from `LocalAddr` via `hostOnly()` — prevents `bind: address already in use` with `--snmp.source-address=host:port`
- Deep-copies `SecurityParameters` via `.Copy()` — avoids shared AES/DES IV state across connections
- Zeroes `Conn` — connections are not shared

**Metrics**: `packets`/`retries` counters in `collect()` use `sync/atomic` types for correctness. Walk-phase packets are not counted when `walk_concurrency > 1` (clone callbacks are intentionally nil); the metric help strings document this.

## Testing

```
go test ./... -race
```

All packages pass. New tests:
- `TestScrapeTargetParallelWalk` — `walk_concurrency=3`, 3 subtrees × 2 PDUs, verifies all 6 PDUs present
- `TestScrapeTargetParallelWalkConnectError` — clone connect failure propagates cleanly
- `TestHostOnly`, `TestCloneStripsLocalAddrPort`, `TestMockCloneCopiesConnectError` — `Clone()` correctness


lukeod:
Not a maintainer, but took a look at the PR as it's a feature i've considered doing a PoC for but never got around to it.

I think there's some significant issues with the PR as it stands:

- Nothing in the test suite ever sets `walk_concurrency > 1`, so the new code path is never executed by any test. So the `go test ./... -race` result doesn't mean much.
- `testMetrics()` was added but has no callers. The mock's `Clone()` also doesn't copy `ConnectError`, so the clone-connect-failure branch can't be tested as-is.
- There's a race: `ScrapeTarget` needs to wait for all its walk goroutines before returning, but on the first walk error it returns immediately (`wg.Wait()` only happens in the channel-closer goroutine) - the leftover goroutines then `Clone()` the shared client while the worker reconfigures it via `SetOptions` for the next module.
- `Clone()` copies `LocalAddr` verbatim, so with a fixed source port (e.g. `--snmp.source-address=10.0.0.5:16100`) every clone's `Connect()` fails with `bind: address already in use`
- Clones drop `OnSent`/`OnRecv`/`OnRetry`, which are the only source of `snmp_packets_total`, `snmp_retries_total`, `snmp_duration_seconds` and `snmp_scrape_packets_sent/_retried`. With concurrency on, walk traffic (i.e. nearly all of it) goes uncounted.
- Connections are made per-subtree, not per-worker: N walk entries = N connect/close cycles per scrape regardless of the concurrency setting, and nothing cancels in-flight clones after an early error.

Suggestion: a fixed pool of `concurrency` workers, each doing `Clone()`+`Connect()` once and pulling subtrees off a channel (same pattern as the existing worker pool in `Collect()`), wrapped in `errgroup.WithContext` so the first error cancels the siblings and the call waits for them before returning. That bounds connections to `concurrency` and fixes the race and orphan-walk issues in one go; `golang.org/x/sync` is already an indirect dep.

amuntean-godaddy:
Thank you @lukeod @SuperQ 
I am going to re-work on the issues you have pointed out and run the real test in my environment.
I am not very proficient in go, however I really need the concurrency support and using AI here :) .

amuntean-godaddy:
@SuperQ could you please review the current changes ?

SuperQ:
Looks like some CI issues.

amuntean-godaddy:
Hi @SuperQ @bastischubert @RichiH — addressed all the lint issues raised by CI:

- **`gci`**: fixed import grouping in `collector/collector.go`, `collector/collector_test.go`, and `scraper/gosnmp.go`
- **`errorlint`**: replaced `err != wantErr` with `errors.Is(err, wantErr)` in `scraper/gosnmp_test.go`
- **`staticcheck`**: removed the deprecated `NonRepeaters` field from `Clone()` in `scraper/gosnmp.go`

The **Build generator** failure is an infrastructure flake (503 fetching Dell MIBs from an external URL — unrelated to this PR).

The branch is ready for re-review. Thanks for your time!

SuperQ:
@lukeod Want to give this another pass?

lukeod:
The PR is much improved. The structural concerns from my earlier comment are resolved.

Taking a step back, i think the `Clone()` should be upstreamed to gosnmp. Getting it right requires pretty detailed understanding of gosnmp internals (the field-by-field copy because the struct embeds a mutex, deep-copying `SecurityParameters` so clones keep the discovered engine state, the `LocalAddr` port handling). It's also sensitive to gosnmp changes: any new config field gets silently dropped by clones until someone notices.

I've put together a `(*GoSNMP).Copy()` for gosnmp that i am happy to raise as a PR if @SuperQ concurs that it's the right direction: https://github.com/lukeod/gosnmp/tree/add-copy-method.

In the meantime, two issues with the current implementation:

1. `hostOnly()` breaks clone `Connect()` whenever `--snmp.source-address` is set. It strips the port entirely, but gosnmp passes `LocalAddr` verbatim to `net.ResolveUDPAddr`, which requires host:port form, so every clone Connect fails with "missing port in address" for any source address containing a colon, including the auto-port forms host: and host:0 that work sequentially. Would be eliminated if that add-copy-method gosnmp change was merged though.

2. Walk traffic is invisible to the packet/retry/duration metrics. `Clone()` drops the `OnSent`/`OnRecv`/`OnRetry` hooks, so all walk-phase traffic is uninstrumented at `walk_concurrency > 1`.

Smaller points:

- For SNMPv3 walk-only modules, each worker redoes engine discovery: the parent hasn't sent anything before `Clone()` is called, so the copied engine ID is still empty and every clone's first request triggers its own discovery probe. Not a correctness problem, just N discovery exchanges per scrape instead of 1. Preventing it would mean priming the parent before cloning (e.g. walking the first subtree on the parent connection, then fanning out the rest), since clones inherit whatever engine state the parent has at `Clone()` time; possibly not worth the complexity.
- No test asserts each subtree is walked exactly once; sharing the call recording between the mock and its clones would enable it. Would be nice to add.
- `i := i` in the worker loop is unnecessary

amuntean-godaddy:
Thanks @lukeod

amuntean-godaddy:
@lukeod I have made the change:

1. **Clone `Connect()` with `--snmp.source-address`:** `hostOnly()` is replaced by `cloneLocalAddr()`, which normalizes the port to `0` (`net.JoinHostPort(host, "0")`) instead of stripping it — keeping the `host:port` form `net.ResolveUDPAddr` requires while still giving each clone a fresh source port. `TestCloneConnectsWithSourceAddress` reserves a bound UDP port as the parent's source address and asserts a clone can still connect. (3a88e72)

2. **Walk traffic instrumentation:** `SetOptions` now records the option fns it applies and `Clone()` replays them on the copy. Re-invoking the metrics option closure gives each clone its own `OnSent`/`OnRecv`/`OnRetry` closures (fresh `sent` state, so no shared-closure race) feeding the same atomic counters, so `snmp_scrape_packets_sent`/`snmp_scrape_packets_retried`/`snmp_packet_duration_seconds` cover walk traffic at `walk_concurrency > 1`. This also replays `ConfigureSNMP`, so clones pick up auth the same way the parent did. The metric help strings are restored to their original wording. (1451c65)

3. **Exactly-once walk test:** the mock scraper now shares one mutex-protected call record between parent and clones, and `TestScrapeTargetParallelWalk` asserts each subtree is walked exactly once. (177c825)

4. `i := i` removed. (58c7906)

amuntean-godaddy:
@lukeod @SuperQ can you please review this PR again ?

lukeod:
I think the big outstanding item is a decision on where the clone code should live. It seems like it should be upstreamed to gosnmp to me, but one for @SuperQ to make a call on - may be some considerations with gosnmp v2 timing/scope.

The `config.Regexp{regexp.MustCompile(...)}` changes in collector_test.go should be descoped from the PR, not related to the feature.

amuntean-godaddy:
@SuperQ What's your call on this PR ?

PR #1653, opened 2026-07-06
Reject index types the collector cannot render

lukeod:
The generator emits index types the collector cannot render as labels (`DateAndTime`, `Bits`, etc). `indexOidsAsString` panics on them in the collect goroutine, which exits the whole exporter process - one row in an affected table is enough. This is reachable from standard MIBs: `ALARM-MIB::alarmActiveTable` (RFC 3877) is indexed by `alarmActiveDateAndTime`, so generating a module for it produces a config that crashes the exporter on the first scrape of a populated table.

The proposed approach:

- Drop tables whose index type is not renderable, with a warning, matching the existing unhandled-type behavior.
- Fail generation when a chained lookup promotes a non-renderable type to an index.
- The allowlist is `config.RenderableIndexTypes`; a collector test asserts every entry renders without panic. It lives in `config`, which the generator and collector both already import. `Integer`/`Integer32` are listed for hand-written configs (the generator never emits them as index types)

No module in the shipped `snmp.yml` carries an affected index type, so regenerating it drops nothing. The Bits test expectation is updated: the previous expected output encoded a config that panics the exporter. A `type: OctetString` override keeps a dropped table (renders as hex). Some rejected types could reasonably be rendered in future (e.g. DateAndTime as a timestamp label); adding the collector case and listing the type re-admits those tables. Hardening the collector against such configs already in the field (hand-written or from older generators) is another possible follow-up.

Part of the generator display_hint work for #1610, alongside #1646 and #1648.


lukeod:
Earlier CICD failures were caused by some sort of outage/issue for Palo Alto support page - `https://docs.paloaltonetworks.com/content/dam/techdocs/en_US/zip/snmp-mib/pan-10-0-snmp-mib-modules.zip` was failing to respond. Force pushed a trivial change to retrigger the pipeline which seems to work now. I think it's fine to review / merge if maintainers happy.

SuperQ (approved):
Thanks. Should we also add the validation check to the exporter config load?

generator/tree.go
```diff
@@ -396,6 +396,12 @@ func generateConfigModule(cfg *ModuleConfig, node *Node, nameToNode map[string]*
 					index.Type = "EnumAsInfo"
 				}
 
+				// The collector can only render certain types as index labels.
+				if !config.RenderableIndexTypes[index.Type] {
+					logger.Warn("Can't render index type as a label", "node", n.Label, "index", i, "type", index.Type)
```
SuperQ: I wonder if we should do stronger than warn here?

lukeod: Yeah i think you're right. Pushed a followup commit

SuperQ (approved):
Nice, thanks!

PR #1660, opened 2026-07-22
feat: convert exporter into active SNMP collector

kazuyuki114:


SuperQ:
No, this violates Prometheus best practices in so many ways.

PR #1663, opened 2026-07-22
Add Eaton ePDU module

YichanKim:
Adds an eaton_epdu module for Eaton ePDU units.

Same style as the existing raritan pdu modules - walks the input, group, outlet and environmental tables and exposes voltage, current, power, temp and humidity. MIB is pulled from librenms in the generator Makefile and snmp.yml is regenerated.

One thing to note: the regen also added a single field to the existing apcups module (upsHighPrecBatteryStateOfHealth) since the APC MIB was updated upstream. Left it in so the generator CI diff stays clean.

@SuperQ

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -78,6 +78,9 @@ YAMAHA_URL        := https://www.rtpro.yamaha.co.jp/RT/docs/mib/
 CYBERPOWER_VERSION := 2.11
 CYBERPOWER_URL     := https://dl4jz3rbrsfum.cloudfront.net/software/CyberPower_MIB_v$(CYBERPOWER_VERSION).MIB.zip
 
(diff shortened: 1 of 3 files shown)


bastischubert:
Hey @YichanKim ,
thanks for the PR. 
We try to use first class downloads from the mib vendors instead of unofficial mirrors from github - or our own mirror at https://github.com/prometheus-community/snmp.
That beeing said, https://www.eaton.com/content/dam/eaton/products/backup-power-ups-surge-it-power-distribution/firmware-emea/rack-pdu-g4-firmware-emea/eaton-rackpdu-g4-mib.zip would be the current place the mibs files are stored from the vendor.

YichanKim:
Thanks @bastischubert!
I tried that vendor zip first, but eaton.com resets the connection for curl (error 56). So I have mirrored the two MIBs (from that exact G4 zip) in https://github.com/prometheus-community/snmp/pull/35 and will point this PR's Makefile there. Thanks for the heads up

PR #1665, opened 2026-07-29
Make generator retry to download MIB-s on all errors (4XX included)

TafkaMax:
When using the generator, some URL-s have defences on them and return 403 or whatever else errors. For example the PICO/NEC url-s. Therefore us `retry-all-errors` on curl to use the retry mechhanism on 4XX errors aswell. Default is only for 5XX.

### DIFF
--- a/generator/Makefile
+++ b/generator/Makefile
@@ -15,7 +15,7 @@ MIBDIR   := mibs
 MIB_PATH := 'mibs'
 SNMP_YML ?= '../snmp.yml'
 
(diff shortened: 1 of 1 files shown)


bastischubert:
@TafkaMax do you mind signing the commit in order to pass the DCO check?

TafkaMax:
> @TafkaMax do you mind signing the commit in order to pass the DCO check?

Yeah, i need to find how to do that

TafkaMax:
OK added the signed-off-by

PR #1674, opened 2026-08-09
generator: map String64 textual convention to DisplayString

### LINKED ISSUE #1387: Exporter is treating "String64" syntax as "octetstring" instead of "displayString"

locker95:
Vendor MIBs (Dell iDRAC in #1387) use a `String64` textual convention on OCTET STRING without a DisplayString DISPLAY-HINT, so the generator left them as octet strings and the exporter printed hex.

Map `String64` to `DisplayString` the same way we already handle the DisplayString TC.

Fixes #1387

### DIFF
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -122,7 +122,7 @@ func prepareTree(nodes *Node, logger *slog.Logger) map[string]*Node {
 		}
 		// Vendor MIBs (e.g. Dell iDRAC) often define String64 as an OCTET STRING
 		// textual convention for printable text without a DisplayString DISPLAY-HINT.
-		if n.TextualConvention == "String64" {
+		if n.TextualConvention == "String64" || n.TextualConvention == "StringType" {
 			n.Type = "DisplayString"
 		}
 		if n.TextualConvention == "PhysAddress" {
--- a/generator/tree.go
+++ b/generator/tree.go
@@ -120,6 +120,11 @@ func prepareTree(nodes *Node, logger *slog.Logger) map[string]*Node {
 		if n.TextualConvention == "DisplayString" {
 			n.Type = "DisplayString"
 		}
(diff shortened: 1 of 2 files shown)


SuperQ:
I looked over the iDRAC MIB. There's a whole bunch of these similar syntaxes, but they don't contain a DISPLAY-HINT. 

I wonder if we should provide a more generic syntax to allow matching custom types to textual conventions.

locker95:
yeah, String64 was just the iDRAC one I hit. a generic 'map this TC to DisplayString' knob would be nicer if more vendor MIBs keep showing up without a DISPLAY-HINT.

happy to recast this PR that way if you'd rather not special-case String64/StringType.

PR #1678, opened 2026-08-19
Add type override ParseFloatString

SuperQ:
Add a new custom type override, `ParseFloatString` that allows simplifying the trivial "OCTET STRING is a number in ascci text" cases that typically reqiure a more complicated `regex_extracts` expression.
* Apply `ParseFloatString` to a number of generator.yml examples.


bastischubert (approved):
looks good, tested with my tplink ddm metrics as well

PR #1679, opened 2026-08-23
config: expand environment variables in the community field

randomizedcoder:
## Problem

`--config.expand-environment-variables` substitutes environment variables in the auth `username`, `password`, and `priv_password` fields, but **not** in `community`. So a config like:

```yaml
auths:
  my_v2c:
    version: 2
    community: ${SNMP_COMMUNITY}
```

passes the literal string `${SNMP_COMMUNITY}` to the device as the community. The SNMPv2c agent silently drops the request and the scrape fails with:

```
error getting target <ip>: request timeout (after 3 retries)
```

This is easy to misdiagnose as a network/timeout problem, especially since the same flag correctly sources SNMPv3 secrets from the environment — so v3 targets work while v2c targets time out, using the identical mechanism.

## Fix

Expand `community` alongside the other secret-bearing auth fields in `config.LoadFile` (it is already a `config.Secret`, so this mirrors the existing `password` handling exactly).

## Testing

Extended the existing `TestEnvSecrets` / `TestEnvSecretsMissing` / `TestEnvSecretsEmpty` cases (and the `snmp-auth-envvars.yml` fixture) to source `community` from `${ENV_COMMUNITY}` and assert it resolves. The fixture's community value doubles as the secret-masking canary, so this also verifies `community` is still redacted by the `String()` method.

Verified against live hardware (two Juniper EX switches via SNMPv2c): before the patch, `${VAR}` communities time out after 3 retries; after, both switches scrape successfully.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -63,6 +63,13 @@ func LoadFile(logger *slog.Logger, paths []string, expandEnvVars bool) (*Config,
 	if expandEnvVars {
 		var err error
 		for i, auth := range cfg.Auths {
+			if auth.Community != "" {
+				community, err := substituteEnvVariables(string(auth.Community))
+				if err != nil {
+					return nil, err
+				}
+				cfg.Auths[i].Community.Set(community)
+			}
 			if auth.Username != "" {
 				cfg.Auths[i].Username, err = substituteEnvVariables(auth.Username)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


SuperQ:
This is intentionally not supported.

PR #1690, opened 2026-09-06
Cleanup linting issues

SuperQ:
* Cleanup various linting issues reported by golangci-lint.
* Use `net/netip` IP pasrsing to produce address string output.


bastischubert (approved):
LGTM

